PR #336, opened 2026-06-16
fix(timetable): block destructive replace import when parsed set is empty

### LINKED ISSUE #309: bug: timetable "replace" import wipes school data (timetable + relief assignments) with no confirmation or undo
## Description
Importing a timetable in **replace** mode deletes **all** of the school's teachers, periods, timetable entries, **and relief assignments** before importing — with no confirmation step and no undo. An admin importing an empty or wrong file (or testing) silently destroys the school's data.

## Steps to reproduce
1. School has a populated timetable + relief assignments.
2. POST a `replace` import with empty arrays (or upload a wrong/empty file).
3. All teachers / periods / timetable entries / relief assignments are deleted; response is 200 `created: 0/0/0`.

(Observed live during P1 verification — a single empty replace wiped the seeded demo data.)

## Expected behaviour
A destructive replace should be guarded — e.g. block replace when the parsed set is empty, require an explicit confirmation, and/or preserve relief-assignment history (soft-delete / archive) rather than hard-deleting it.

## Area
Timetable import (`src/lib/timetableImportExecutor.ts`, `replace` mode)

## Severity
pilot-risky — data loss with no undo during the exact window (onboarding / imports) when admins are most likely to make a mistake. Closely related to the history-preservation need in the #288 follow-up.

## Notes
Found during testing this session.

guangshinhaha:
## Summary

Adds a server-side safety guard to the timetable import executor so a destructive **replace** import can no longer wipe a school's data when the parsed import set is empty.

`executeImport` now refuses an empty full-school replace **before the transaction opens**: if `mode === "replace"`, there is no `overrideId`, and there are zero teachers to create **and** zero entries to insert, it throws a new `EmptyReplaceImportError` ("Refusing to replace: the import contains no timetable data."). No `deleteMany` ever runs. The `/api/import-timetable` route (ASC-XML, CSV confirm, PDF confirm paths) maps that error to a `422` with the clear message.

## Why

In `replace` mode, `executeImport` deletes **every** teacher, period, timetable entry, relief assignment, and absence report for the school before re-inserting. When the upload was empty or the wrong file (parsing to zero teachers and zero entries), the delete still ran and nothing was inserted in its place — silently destroying all of the school's data while the API returned `200` with `created: 0/0/0`. This was observed wiping seeded demo data during testing (#309).

The guard closes the empty-file / wrong-file data-loss hole. Behaviour for a **non-empty** replace, for **merge** (which only clears the teachers it re-imports), and for **override** imports (scoped to a single override, never touch school-wide tables) is unchanged.

## Test plan

Added to `src/__tests__/timetableImportExecutor.test.ts` (pure-logic suite — the guard runs before any DB work, so the transaction is mocked):

- empty replace rejects with `EmptyReplaceImportError` and the transaction is **never** entered (asserts no deletion happens)
- error message is clear and actionable
- replace with teachers but no entries is **not** blocked (transaction runs)
- replace with entries but no new teachers is **not** blocked
- empty **merge** is **not** blocked
- empty **override** import is **not** blocked

Note: local `npm install` is blocked in this environment (private `@flow` registry returns 403, no `node_modules`), so `npm test` / `tsc` / `next build` could not be run locally — verification relies on CI.

## Follow-ups (not in this PR)

- A confirmation step in the import UI before a destructive replace.
- Soft-delete / history preservation so an accidental replace is recoverable rather than only blocked at the empty-set boundary.

Closes #309

https://claude.ai/code/session_01WbXeV2QEvCqxG158LZt9w1

---
_Generated by [Claude Code](https://claude.ai/code/session_01WbXeV2QEvCqxG158LZt9w1)_


guangshinhaha:
## Code review — APPROVE WITH NITS

**Overview:** Adds a pre-transaction `EmptyReplaceImportError` guard in `executeImport` and maps it to HTTP 422 in all three confirm handlers (ASC-XML, CSV, PDF). Tests in the pure-logic suite.

**Correctness:** The guard sits **before** `$transaction`, so when it fires no `deleteMany` runs — and the test proves it via `expect(transactionMock).not.toHaveBeenCalled()` (the strongest assertion here). merge/override exemptions are correct. 422 mapping is complete and consistent across all three handlers.

**Decision needed (edge case):** the guard checks `teachersToCreate.length === 0 && entries.length === 0` but ignores `periodsToCreate`. An ASC-XML replace that parses to **periods only** (zero teachers/entries) would still pass the guard and wipe teachers, relief assignments, absence reports, and entries — re-inserting only periods. Since the issue is specifically "wrong file wipes school data," consider also requiring `periodsToCreate.length === 0`, or keying the replace guard on `entries.length === 0` alone. Add a test for whichever behavior you choose.

**Nits:**
1. No `tests/api/` integration test for the 422 route mapping (duplicated x3) — an end-to-end assertion would close the loop.
2. Confirm CI green before merge (local build/test couldn't run — `@flow` registry 403).

Core fix is correct and verifiably closes the reported empty/wrong-file data-loss hole.

---

guangshinhaha (commented):
## Independent code review — Verdict: ✅ APPROVE-WITH-NITS

_Automated independent review (Claude). Read-only._

**Summary:** Adds a server-side `EmptyReplaceImportError` guard in `executeImport` that refuses a destructive full-school `replace` import when the parsed set is empty (zero teachers AND zero entries), throwing **before** the transaction/`deleteMany` runs; the route maps it to HTTP 422 across the ASC-XML, CSV-confirm, and PDF-confirm paths.

**Edge-case decision (the flagged one): correct and safe.**
- "Empty" = no teachers to create AND no resolvable entries, measured on planner output (post-FK-resolution). So a wrong-school/malformed file that parses to all-skipped rows is also caught — not just a zero-byte file.
- Guard fires before `$transaction` opens, so no `deleteMany` ever runs (the test asserts the transaction mock was never called — the right invariant).
- Non-empty path stays transactional (delete+insert in one interactive tx, 60s timeout → rolls back on partial failure).
- Override imports correctly exempted (`!plan.overrideId`); verified the only override path uses `mode: "merge"`, never `replace`.
- Multi-tenancy sound: `schoolId` from `requireSchool()` (session), never the request body.

**Findings**
1. **NIT** — the guard can't distinguish a genuinely-empty intended upload from a parse failure; both are blocked. That's the safe choice for a destructive op; flagging only so the deliberate conservatism is on record.
2. **MINOR** — no integration test in `tests/api/import-timetable.test.ts` asserting the route returns **422** and that the school's data **survives**. The unit guard test proves the throw; an API-tier test would prove the incident symptom is gone. Recommend as a follow-up.

**Test coverage:** ✅ Solid unit coverage (6 cases). Gap is the API-tier 422/data-survival test noted above.

**Before merge:** confirm CI green (author couldn't build locally — registry 403).

Issue #349, opened 2026-06-17
bug(auto-assign): confirm (preview=false) 403s at edge for large assignment sets — can preview but not commit


guangshinhaha:
## Description
Committing an auto-assign (`POST /api/auto-assign?preview=false`) for an absence with many periods returns **403** at the edge — same ~8 KB POST-body limit as #343. You can **preview** a large day's auto-assign but cannot **commit** it on the deployed env, because the confirm payload (`{date, weekType, assignments:[...]}`) exceeds the limit once there are enough proposals.

## Steps to reproduce
1. On the test env, create a full-day absence for a teacher with many periods (≥~8 timetabled).
2. `POST /api/auto-assign?preview=true` → 200 with N proposals.
3. `POST /api/auto-assign?preview=false` with those proposals → **403** (nginx edge HTML), nothing persisted.

## Expected
Confirm succeeds and persists the assignments.

## Actual
403 at the edge; the assignment set is never written.

## Area / Severity
auto-assign / infrastructure. **High** — blocks the core "review → commit" flow for any sizeable day. (Same root as #343; filing separately because the user-facing impact — can't commit relief — is distinct from "can't import a timetable".)

## Notes
Fix is the same edge `client_max_body_size` raise as #343, OR chunk the confirm payload client-side. Discovered while testing the workload rule (the confirm 403 silently produced a false "no workload persisted" state).

guangshinhaha:
Triage: **same root cause as #343 (edge body-size limit), and that was an infra fix — no repo change.** Per #343, the blocker was the AWS WAF managed rule `SizeRestrictions_Body`, which @chadinwork has since raised ("Firewall updated. Pl try again."). That rule is global to the edge, so the auto-assign confirm (`POST /api/auto-assign?preview=false`) should be unblocked by the same change.

**Action:** re-verify the review → commit flow for a large day on the test env (preview, then `preview=false`). If the confirm still 403s past ~8&nbsp;KB after the WAF change, reopen with the failing payload size and we'll chunk the confirm payload client-side as the fallback.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #343, opened 2026-06-17
bug: timetable import blocked by ~8KB edge POST body limit (403) — all formats & entry points unusable on test env


guangshinhaha:
## Description

Timetable import is broken end-to-end on the test environment (`reliefcher.string.sg`). An edge/reverse-proxy (nginx) in front of the app rejects any POST request with a body larger than ~8 KB with a bare `403 Forbidden`, before the request reaches Next.js. Every import flow serialises the parsed school into a single JSON body that is far larger than 8 KB, so **no real school can import a timetable** in any format (XML / PDF / CSV) via any entry point (onboarding, settings tab, override).

## Steps to reproduce

1. Sign in to a school admin account (or onboarding) on `https://reliefcher.string.sg`.
2. Go to onboarding (`/onboarding`) or **Settings → Import Timetable**.
3. Choose **aSc Timetables XML** and select a real export (e.g. `docs/data-xml/hci.xml` — 271 teachers).
4. Client-side parse succeeds and shows the Import Summary (271 teachers / 18 periods / 9045 entries).
5. Click **Import Timetable**.

## Expected behaviour

The parsed timetable is POSTed to `/api/import-timetable`, the app creates teachers/periods/entries, and the summary shows the created counts.

## Actual behaviour

`POST /api/import-timetable` returns **403 Forbidden** with an nginx HTML error page (not an app JSON response). Nothing is created (teacher count stays 0). The same happens for every real file and every format/entry point.

Controlled body-size sweep against `POST /api/import-timetable?source=asc-xml` (clean JSON, valid same-origin request, session cookie present):

| Request body size | Result |
|---|---|
| ~5 KB | 200 (app responds, `created: 0/0/0`) |
| ~8 KB | 200 (reaches app) |
| ~12 KB | **403 (nginx edge HTML)** |
| 16 KB – 1 MB | **403 (nginx edge HTML)** |
| `asctt2012.xml` payload (464 KB) | **403** |
| `hci.xml` payload (1.5 MB) | **403** |

The cutoff is ~8 KB. The payload is well-formed JSON (`{mode, teachers, periods, entries}`) with **zero** XML/HTML tags, so this is a request **body-size limit**, not a content/WAF-signature rule, and not an auth/CSRF issue (GET reads return 200 with the same session; a 5–8 KB POST reaches the app and returns 200).

## Area

timetable import (edge / infrastructure)

## Severity

[...]

PR #389, opened 2026-06-25
fix(dashboard): trim auto-assign confirm payload under the ~8 KB edge limit

guangshinhaha:
## Summary
Auto-assign **Confirm & Assign All** failed on busy days with a generic **"Failed to confirm auto-assign"** alert. Root cause: the confirm request body exceeded the **~8 KB nginx edge limit** (#349 — same edge as the timetable-import #343) and was rejected with a non-JSON `403 Forbidden` HTML page *before reaching Next.js*. `res.json()` then threw, so the client fell back to its generic message.

The body was bloated because the handler POSTed the **full enriched preview proposals** — each carrying `daySchedule[]`, `alternatives[]`, teacher/subject names and workload counts — when the route only persists three fields per assignment.

## Why
Preview enriches every proposal for display; confirm only needs `{ absenceReportId, timetableEntryId, reliefTeacherId }`. Sending the rest pushed ~21 proposals to ~21 KB, over the edge limit. Stripping to the persisted fields drops that to **~2.8 KB** — well under 8 KB — while keeping the single atomic transaction server-side. (Helper covers schools up to ~60 absences/day before approaching the limit again.)

## Changes
- `src/lib/autoAssignConfirm.ts` — new `toConfirmAssignments()` helper (documented with the #349 rationale)
- `src/__tests__/autoAssignConfirm.test.ts` — asserts only the 3 fields survive, and that a realistic 21-proposal payload goes from >8 KB to <8 KB
- `src/app/[slug]/dashboard/ReliefOverview.tsx` — confirm handler sends the minimal payload

## Test plan
- [x] `npm test` (new unit test passes)
- [x] `tsc --noEmit` clean
- [x] `npm run build` green (90/90 static pages, no `⨯`)
- [x] **Reproduced live** on `reliefcher.string.sg`: confirm bodies ≥ 8 KB → `403 text/html`; ≤ 6 KB → `200 application/json`. Probed via authenticated Playwright (dev OTP bypass).

## Follow-up (infra — see PR comment)
The code fix resolves the reported case, but the underlying ~8 KB edge limit is path-scoped and remains a latent trap for other large-payload endpoints. Tagging @chadinwork below.


guangshinhaha:
@chadinwork — infra ask: please raise the nginx edge `client_max_body_size` for the auto-assign confirm endpoint (or globally).

**Why:** the ~8 KB POST-body limit you raised earlier is **path-scoped to `/api/import-timetable` only** — `/api/auto-assign` still hits the default. Confirmed live today, both authenticated, identical padding:

| Endpoint | 6 KB body | 10 KB body |
|---|---|---|
| `/api/import-timetable` | `400` JSON (reaches app) | **`400` JSON — reaches app** ✅ |
| `/api/auto-assign` | `200` JSON | **`403 text/html` — blocked at edge** ❌ |

This PR trims the auto-assign confirm payload so it stays under 8 KB and unblocks the reported failure **without** needing this change. But the per-`location` limit is a latent trap for any future large-payload endpoint.

**Preferred:** bump `client_max_body_size` at the `server`/`http` level (e.g. `2m`) so we stop whitelisting per endpoint. **Minimum:** add `/api/auto-assign` to the same `location` block you used for `/api/import-timetable`.

Refs #349 (auto-assign confirm), #343 (timetable import — the original edge-limit report).

---

### 🔧 UPDATE — full list of endpoints still blocked (QA sweep, 2026-06-25)

The per-`location` whitelist approach has bitten us in **more places than auto-assign**. A full QA pass confirmed the ~8 KB cap is still rejecting these bulk endpoints at the edge (`server: awselb/2.0`, ALB `403 Forbidden` HTML page, *before* Next.js). **This strongly favours the global `client_max_body_size` bump** over whitelisting — there are at least 6 paths, and onboarding is among them.

Threshold pinned precisely on `/api/teachers/import`: **110 rows / 8.0 KB → `200`; 113 rows / 8.2 KB → `403`**. A realistic **120-teacher staff CSV (8.7 KB) → `403`**.

**🔴 Must-fix — block real onboarding/setup:**

| Endpoint | Carries | Note |
|---|---|---|
| `POST /api/teachers/import` | Staff CSV (Import modal **+ the Onboarding wizard**) | Confirmed `403` at 8.2 KB; UI sends the whole CSV unchunked → a normal-sized school **cannot import staff or onboard** |
| `POST /api/import-timetable` | Timetable (asc-xml / csv / xlsx) | Already exempted — listed for completeness |
| `POST /api/pool-teachers/import` | REMS bulk import (superadmin) | App enforces its own `MAX_IMPORT_BYTES = 5 MB` / 20k rows — the 8 KB edge cap makes it **~625× too small** |
| `POST /api/[slug]/timetable-overrides/[id]/import` | Exam / special-day timetable import | Same family (413-aware in code) |

**🟠 Also affected (large but lower-frequency):**

| Endpoint | Carries |
|---|---|
| `POST /api/auto-assign?preview=false` | Edited `assignments[]` on busy days (the #349 case this PR works around) |
| `POST /api/admin/leads/bulk-import` | Leads bulk CSV (superadmin CRM) |

**Note for AWS WAF (not just nginx):** if the edge is the AWS managed `SizeRestrictions_Body` rule, it inspects only the first 8 KB by default — so raising the limit also needs **oversize handling** configured for these paths, and `/api/pool-teachers/import` should be allowed up to a few MB to match the app's own 5 MB gate.

(Belt-and-suspenders, app-side: the staff/timetable import UIs currently POST the whole file in one request — chunking them client-side would remove the hard dependency on this config, but the infra bump is the real fix for the REMS + timetable paths.)

Issue #425, opened 2026-07-02
bug(assignments): manual assign + auto-assign confirm skip 'no self-cover' / 'no assign-while-absent' hard constraints (server trusts client reliefTeacherId)


guangshinhaha:
## Summary (correctness / integrity — hard constraints not enforced server-side)

PRD §9 lists two **hard constraints that must never be violated**: **no self-cover** (the absent teacher can't cover their own class) and **no assign-while-absent** (a teacher who is themselves absent that day can't be a reliever). These are enforced **only in the candidate picker** (`available-teachers`), not in the write paths. Both write endpoints trust the client-supplied `reliefTeacherId`, so a stale dashboard, a crafted request, or the "edit before publish" auto-assign flow can persist a nonsensical assignment.

This violates PRD §1 ("server-verified authorisation — never trust the client") and is a "one rule, missing at some entry points" gap: the picker has the logic, the write paths don't.

## Reproduction (confirmed on live test env `reliefcher.string.sg`, deployed `main`, school `maia-pri-1`)

Setup: reported teachers **Gerald** and **Leon** absent (FULL_DAY) on 2026-07-06; both own a period-1 class that day.

**Case 1 — self-cover accepted:**
```
POST /api/relief-assignments
{ absenceReportId:<Gerald's>, timetableEntryId:<Gerald's own P1 class>, reliefTeacherId:<Gerald>, date:"2026-07-06" }
→ 201 Created   (Gerald assigned to cover his own absent class)
```
The dashboard then shows Gerald's P1 slot "covered" by **Gerald himself**.

**Case 2 — absent teacher assigned as reliever:**
```
POST /api/relief-assignments
{ absenceReportId:<Gerald's>, timetableEntryId:<Gerald's P1 class>, reliefTeacherId:<Leon>, date:"2026-07-06" }
→ 201 Created   (Leon, absent the same day, assigned as reliever)
```

**Expected:** both → `400`/`409` rejected. The picker (`GET /api/dashboard/available-teachers`) correctly excludes both Gerald (self) and Leon (absent) — so the UI hides them — but the write path does not re-check.

(Cleanup done: all test assignments cancelled.)

## Root cause

`src/app/api/relief-assignments/route.ts` (POST) validates school-scope, concurrent double-book (`teacherBusy`), and slot-taken (`slotTaken`), but never checks:
1. `reliefTeacherId !== timetableEntry.teacherId` (self-cover), and
[...]

PR #426, opened 2026-07-02
fix(security): gate notification-log and school-config read APIs against TEACHER access (#424)

guangshinhaha:
Closes #424

**Stack position: 1/12** — base `main`. This is the first PR of the ticket-batch stack; merge this one first.

## Summary

- `GET /api/notifications` (the admin Notification Log data API) was authorized on school membership only — any TEACHER could read colleague names, emails, and free-text decline reasons. It is now gated on `ASSIGN_INTERNAL_RELIEF`, the same capability as its sibling cancel route, so SCHOOL_ADMIN and RELIEF_COORDINATOR keep the log and TEACHER gets 403.
- Swept the related membership-only reads flagged in the issue: `GET /api/school-settings` (summaryEmails PII + algoRules), `GET /api/matching-config`, and `GET /api/school-roles` are now `requireSchoolAdmin()`, mirroring each route's own admin-gated mutations.
- Verified callers before tightening: the only UI consumers of the swept GETs are the `MANAGE_SCHOOL_CONFIG`-gated settings tabs and the SCHOOL_ADMIN-gated onboarding wizard. `/api/periods` and `/api/positions` were left open (needed by the teacher reporting flow, per the issue).

## Changes

- `src/app/api/notifications/route.ts` — capability gate + shared `authErrorResponse` mapping (was a hand-rolled 401-only catch).
- `src/app/api/school-settings/route.ts`, `src/app/api/school-roles/route.ts`, `src/app/api/matching-config/route.ts` — GET → `requireSchoolAdmin()`; matching-config's duplicate local error helper now delegates to `authErrorResponse`.
- `tests/api/notifications-routes.test.ts` — TEACHER→403 (with a seeded PII row), RELIEF_COORDINATOR→200.
- `tests/api/school-settings.test.ts` — GET TEACHER→403, RELIEF_COORDINATOR→403.
- `tests/api/school-roles.test.ts` — GET TEACHER→403.
- `tests/api/matching-config.test.ts` — new file: 401 unauth / 403 teacher / 200 admin.

## Test plan

- `npx vitest run --config vitest.api.config.ts` on the 4 touched files: **64/64 pass**.
- `npm test`: 723 passed (baseline unchanged).
- `npm run build`: clean, 88/88 static pages.
- Full `test:api` baseline on `main` has 32 pre-existing failures in 9 unrelated files (stale-expectation tests); none touched by this PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #421, opened 2026-07-02
fix(`security`): validate teacherId/periodId belong to session school on POST /api/timetable (cross-tenant write IDOR)


guangshinhaha:
## User story

As a security reviewer, I want `POST /api/timetable` to reject a `teacherId` or `periodId` that belongs to another school, so that an admin of school A cannot attach school B's teachers/periods to school A's timetable.

## Background

MAIA pentest **#11 (Medium, CWE-639)** — cross-tenant write IDOR. `POST /api/timetable` writes the entry with the **session's** `schoolId` but accepts arbitrary `teacherId` and `periodId` from the request body without validating they belong to that school. An admin of school A can inject school B's IDs; the entry persists, and school B's period metadata (`period.schoolId`) then leaks in subsequent `GET` responses — corrupting multi-tenant referential integrity feeding relief-assignment, attendance and reporting logic.

**Verified still present on `fix/teachers-api-pii-leak` (2026-07-02):**
- `src/app/api/timetable/route.ts` — POST create/upsert (~lines 50, 72–79) uses `requireSchoolAdmin()` for `schoolId` but performs **no** `teacher.schoolId === schoolId` / `period.schoolId === schoolId` validation before insert.
- Contrast (correct pattern): `src/app/api/timetable/[id]/route.ts` DELETE scopes with `where: { id, schoolId }`.

### Cross-tenant IDs are rejected
- **Given** a `SCHOOL_ADMIN` session for school X
- **When** `POST /api/timetable` supplies a `teacherId` or `periodId` belonging to school Y
- **Then** the response is `403` (or `400`) and nothing is persisted

### Same-school writes still succeed
- **Given** both IDs belong to school X → `201` as before

### No foreign metadata in reads
- After the fix, no timetable entry references a cross-school teacher/period, and `GET` no longer embeds foreign `period.schoolId`

## Out of scope

- Broader centralized tenant scoping (tracked in #415) — this ticket is the targeted write-side foreign-key check.

## Technical context

- Before create/upsert, fetch the teacher and period and assert `.schoolId === schoolId`; reject on mismatch. Apply the same check to any PUT/PATCH path in `timetable/[id]/route.ts`.
- Add `tests/api/` coverage: cross-tenant `teacherId`/`periodId` → rejected; same-school → `201`.

## References

- MAIA pentest **#11 (Medium)**. Verified present on `fix/teachers-api-pii-leak` on 2026-07-02.
[...]

PR #427, opened 2026-07-02
fix(assignments): enforce no-self-cover / no-assign-while-absent on both write paths (#425)

guangshinhaha:
Closes #425

**Stack position: 2/12** — base `fix/notifications-read-api-role-gate` (PR #426). Merge after #426.

## Summary

- PRD §9's two hard constraints (no self-cover, no assign-while-absent) were enforced only in the candidate picker; both write paths trusted the client's `reliefTeacherId`. Reproduced on live: the absent teacher was 201'd covering **his own class**, and another absent teacher was 201'd as a reliever.
- New shared predicate `assignmentConstraintViolation()` + `getAbsentTeacherIds()` in `src/lib/scheduling/assignmentConstraints.ts`. The picker, the manual-assign route, and the auto-assign confirm path all use it — one source of truth.
- Status semantics: `SELF_COVER` → **400** (nonsensical request, can't arise from staleness), `RELIEVER_ABSENT` → **409** with a refresh hint (stale-dashboard shape, consistent with the route's other conflicts).
- Bonus hardening: the auto-assign confirm path persisted client-supplied `timetableEntryId`/`reliefTeacherId`/`absenceReportId` with **no validation at all**. It now tenant-scopes all three (400 on foreign/unknown IDs) — closing a cross-tenant write hole of the same class as #421.

## Changes

- `src/lib/scheduling/assignmentConstraints.ts` — new shared lib.
- `src/app/api/relief-assignments/route.ts` — constraint check after the existing tenancy 404s, before the concurrency pre-checks.
- `src/app/api/auto-assign/route.ts` — validates every element of the persisted batch (client-edited or algorithm-generated) before the transaction; one bad element rejects the whole batch.
- `src/app/api/dashboard/available-teachers/route.ts` — picker filter now delegates to the shared predicate.
- `src/__tests__/assignmentConstraints.test.ts` — 4 pure unit tests.
- `tests/api/relief-assignments.test.ts` — self-cover 400, absent-reliever 409, valid 201, absence-range boundary (reliever absent yesterday → allowed today).
- `tests/api/auto-assign.test.ts` — confirm-path: self-cover 400, absent-reliever 409, cross-tenant 400, mixed batch persists nothing, valid batch saves.

## Test plan

- Touched integration files: relief-assignments + auto-assign **28/28**, dashboard-available-teachers **4/4**.
- `npm test`: 727 passed (4 new).
- `npm run build`: clean, 88/88.
- `sick-report-to-assignment.test.ts` still shows its 2 **pre-existing baseline failures** (they fail at the absence-reports POST step, upstream of this change — identical on `main`).


src/app/api/auto-assign/route.ts
```diff
@@ -127,6 +131,59 @@ export async function POST(request: NextRequest) {
...
+            slotTeacherId: entry.teacherId,
+            absentTeacherIds,
+          });
+          if (violation) {
+            return NextResponse.json(
+              { error: violation.message, code: violation.code },
+              { status: violation.code === "SELF_COVER" ? 400 : 409 }
+            );
+          }
+        }
```
guangshinhaha, line 176: **Medium** -- TOCTOU between constraint check and transaction

The \`getAbsentTeacherIds\` query and the \`$transaction\` insert are separate operations. A concurrent absence-report creation between the check and the insert could let a newly-absent teacher slip through as a reliever. For a batch, the gap is wider because the validation loop adds latency before the transaction begins.

This is the same read-then-write pattern already documented in \`relief-assignments/route.ts\` (the TOCTOU header comment), so it's consistent — but worth noting that the batch path amplifies the window.

If you want to close it later, the validation + inserts could run inside a single interactive \`$transaction\` (rather than the current batched array form), so the absence query and the creates share the same snapshot:

```ts
const created = await prisma.$transaction(async (tx) => {
  const absentTeacherIds = await getAbsentTeacherIds(tx, schoolId, dateUTC);
  // ...validation loop using tx for entry/teacher/report lookups...
  return Promise.all(assignments.map((p) =>
    tx.reliefAssignment.create({ data: { ... } })
  ));
});
```

Not blocking — the existing pattern is acknowledged and tracked. Flagging for awareness.

---

src/app/api/dashboard/available-teachers/route.ts
```diff
@@ -22,6 +22,7 @@ import {
   buildTeacherDayPeriodNumbers,
 } from "@/lib/scheduling/availability";
 import { getSchoolTeacherSubjects, getSchoolTeacherClasses } from "@/lib/teacherSubjects";
+import { assignmentConstraintViolation } from "@/lib/scheduling/assignmentConstraints";
```
guangshinhaha: **Low** -- Consider also using \`getAbsentTeacherIds()\` here

This route still builds \`absentTeacherIds\` from its own inline query (the \`activeAbsences\` leg in \`Promise.all\`, unchanged in this diff). The write paths now use the shared \`getAbsentTeacherIds()\` helper, which runs the same query.

Today they are semantically identical, but if the absence query ever needs to change (e.g., filtering by a future \`status\` column on \`AbsenceReport\`, or respecting \`coverageType\`), updating only \`getAbsentTeacherIds\` would silently leave the picker with a stale definition of "absent" -- exactly the read/write drift this PR was built to prevent.

Consider replacing the inline \`activeAbsences\` query + \`new Set(...)\` with a call to \`getAbsentTeacherIds(prisma, schoolId, dateUTC)\` in the \`Promise.all\`.

Not blocking -- the current code is correct and the risk is future-only.

---

src/app/api/auto-assign/route.ts
```diff
@@ -127,6 +131,59 @@ export async function POST(request: NextRequest) {
...
+        const knownTeacherIds = new Set(reliefTeachers.map((t) => t.id));
+        const knownReportIds = new Set(absenceReports.map((r) => r.id));
+
+        for (const a of assignments) {
+          const entry = entryById.get(a.timetableEntryId);
+          if (!entry || !knownTeacherIds.has(a.reliefTeacherId) || !knownReportIds.has(a.absenceReportId)) {
+            return NextResponse.json(
+              { error: "One or more assignments reference records that don't exist in this school." },
+              { status: 400 }
+            );
```
guangshinhaha, line 174: **Medium** -- Batch validation reports only the first violation

The \`for (const a of assignments)\` loop returns on the first constraint violation. If a batch of 10 assignments has 3 violations (say, elements 2, 5, and 8), the admin sees only element 2's error, fixes it, resubmits, then sees element 5, and so on.

Consider collecting all violations and returning them in one response:

```ts
const violations: Array<{ index: number; code: string; message: string }> = [];
for (let i = 0; i < assignments.length; i++) {
  const a = assignments[i];
  const entry = entryById.get(a.timetableEntryId);
  if (!entry || !knownTeacherIds.has(a.reliefTeacherId) || !knownReportIds.has(a.absenceReportId)) {
    violations.push({ index: i, code: "INVALID_REF", message: "References records outside this school." });
    continue;
  }
  const v = assignmentConstraintViolation({ ... });
  if (v) violations.push({ index: i, ...v });
}
if (violations.length > 0) {
  return NextResponse.json(
    { error: "One or more assignments violate constraints.", violations },
    { status: 400 }
  );
}
```

This lets the client highlight all bad rows at once. Not blocking -- the current behavior is safe (it rejects the whole batch), just chatty on retry.

---

guangshinhaha:
## Code Review Summary

**PR #427**: fix(assignments): enforce no-self-cover / no-assign-while-absent on both write paths (#425)

### What Looks Good

- **Shared predicate design** -- `assignmentConstraintViolation()` is clean, well-typed (`ReadonlySet<string>` for the absent set), and returns a structured `{ code, message }` discriminant that both the picker and the write paths consume identically. This is the right abstraction level.
- **Status semantics** -- 400 for `SELF_COVER` (nonsensical, can't arise from staleness) vs 409 for `RELIEVER_ABSENT` (stale-dashboard shape, consistent with the route's existing `STALE_DASHBOARD` conflict) are well-reasoned and documented inline.
- **Cross-tenant validation** on the auto-assign confirm path closes a real security hole (client-supplied IDs were persisted without school-scoping). The `Promise.all` batch lookup pattern is efficient.
- **Test coverage is thorough** -- self-cover 400, absent-reliever 409, cross-tenant 400, mixed batch (one bad element rejects all), valid batch, and the boundary date case (reliever absent yesterday, allowed today). The tests verify both the HTTP status and that zero rows were persisted.

### Findings

| # | Severity | File | Finding |
| 1 | Medium | `auto-assign/route.ts` | TOCTOU between constraint check and transaction -- `getAbsentTeacherIds` runs outside the `$transaction`, creating a window where concurrent absence changes could slip through. Consistent with existing documented pattern; wider window for batches. |
| 2 | Low | `available-teachers/route.ts` | Inline absence query still exists alongside the new shared `getAbsentTeacherIds()` helper -- future drift risk if the query definition changes. |
| 3 | Medium | `auto-assign/route.ts` | Batch validation returns on the first violation only -- admin must fix-and-retry iteratively for multi-violation batches. |

All findings have corrected code suggestions in the inline comments.

### Verdict

**Approve with minor suggestions.** The security fix is solid, the shared predicate prevents the read/write drift that caused the original bug, and the test suite covers the critical paths. The findings are all "should fix eventually" not "must fix before merge."

---

Issue #414, opened 2026-06-29
fix(`assignments`): restore double-booking protection on relief assignments


guangshinhaha:
## User story

As a relief coordinator, I want the system to prevent two simultaneous assignments from booking the same timetable slot, so that a class is never double-staffed and the assignment records stay trustworthy.

## Background

Conflicting assignments are currently caught only by read-then-write pre-checks (`teacherBusy`, `slotTaken`) in `src/app/api/relief-assignments/route.ts`, which is a TOCTOU check, not an atomic guard. The DB-level backstop, a unique index on `(timetableEntryId, date)`, was dropped on 2026-06-12 in migration `20260612085608_add_timetable_overrides` and replaced with a plain index. The route's catch block only maps auth errors, so it does not handle Prisma's P2002 even if a unique constraint existed. Two truly-simultaneous requests whose pre-checks both pass can each create an ACTIVE row for the same slot. `auto-assign` already catches P2002 and returns 409; the manual assignment route does not. Surfaced by the Threat & Risk Assessment (integrity of relief activation).

Note: the unique index was dropped deliberately for the timetable-overrides feature. The first grooming step must establish why (what override scenario needs multiple rows per slot/date) so the restored guard does not regress overrides. A partial unique index scoped to ACTIVE status, or an application-level transactional guard, are likely candidates.

### Concurrent assignments to the same slot are rejected

- **Given** two relief coordinators with stale dashboards, both targeting the same timetable entry on the same date
- **When** both submit an assignment at the same time and both pre-checks pass before either commits
- **Then** exactly one assignment is created and the other request is rejected with HTTP 409, with no double-booked ACTIVE rows persisted

### Timetable overrides are not regressed

- **Given** the timetable-overrides scenario that motivated dropping the original unique index
- **When** the restored double-booking guard is in place
- **Then** the override scenario continues to function (the guard does not block legitimate multiple rows it was designed to allow)

## Out of scope

- Rate limiting the assignment route (separate ticket).
- Audit logging of assignments (separate ticket).
- Redesigning the timetable-overrides data model.

## Design assets

N/A

---

## Technical context

_Pending grooming._

## Data model

_Pending grooming._

## API contract

_Pending grooming._

## Error contract

_Pending grooming._

## Additional test scenarios

_Pending grooming._

## Hard constraints

_Pending grooming._

---

*🤖 Generated with aif-create-issue*

guangshinhaha:
Closed by PR chain #426–#437.

PR #428, opened 2026-07-02
fix(security): enforce role checks on slug-scoped timetable-overrides and rems-candidates (#419)

guangshinhaha:
Closes #419

**Stack position: 3/12** — base `fix/assignment-hard-constraints` (PR #427). Merge after #427.

## Summary

- MAIA pentest #2 (High) / #18 (Medium): `/api/[slug]/timetable-overrides` (GET/POST + `[id]` GET/PUT/DELETE + `[id]/import`) and `/api/[slug]/rems-candidates` (GET) checked tenant but not role — any TEACHER could read/inject relief-planning overrides or browse the REMS pool.
- New shared guard `requireCapabilityForSlug(slug, cap)` in `src/lib/auth.ts`: tenant match + capability + **SUPERADMIN passthrough** (fixes the inverted ACL that 403'd school-less superadmin sessions).
- Capability choices: timetable-overrides → `MANAGE_SCHOOL_CONFIG` (the exceptions UI lives in the settings page, which requires exactly this — RELIEF_COORDINATOR is also excluded); rems-candidates → `ASSIGN_EXTERNAL_RELIEF` (mirrors the sibling preferred-teachers route; RemsPool renders in the same coordinator-visible relief-teachers tab).

## Changes

- `src/lib/auth.ts` — `requireCapabilityForSlug()`, throwing the same strings the shared `authErrorResponse` maps.
- `src/app/api/[slug]/timetable-overrides/route.ts`, `[id]/route.ts`, `[id]/import/route.ts` — all five handlers gated.
- `src/app/api/[slug]/rems-candidates/route.ts` — GET gated.
- `tests/api/slug-route-role-gates.test.ts` — new: teacher 403 (GET+POST, nothing persisted), coordinator 403 on overrides / 200 on rems, admin 200/201, school-less SUPERADMIN 200 on both, cross-tenant 403 on both, `[id]` sub-route gated.

## Test plan

- New authz file + existing `rems-candidates-filters.test.ts`: **17/17 pass** (existing filter tests use the default SCHOOL_ADMIN helper role, unaffected).
- `npm test`: 727 passed. `npm run build`: clean 88/88.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #274, opened 2026-06-05
bug: re-assignment of declined relief covers not working


superb-sushi:
### Description
When a user declines a relief duty, the admin is alerted, and that period does display that it has been rejected (and turns back to red), suggesting that it still requires a new assignment. However, attempting to re-assign a new relief teacher to that duty throws a HTTP 409 Conflict error.

### Steps to Reproduce
1. Assign `Demo Admin` to an uncovered period on the `Dashboard` page
2. Navigate over to `/demo/teacher`. Taking on the persona of `Demo Admin`, decline the relief assignment.
3. Head back to the `Dashboard` page and look for that same period. Try to assign another staff member to that period.
4. Observe that nothing changes. A 409 conflict error appears in the console.

<img width="1864" height="946" alt="Image" src="https://github.com/user-attachments/assets/8746ec59-a120-42be-b8d9-eed21b6b3ff9" />

### Expected Behaviour
The admin should be able to re-assign that period to a new teacher successfully.

PR #429, opened 2026-07-02
fix(assignments): restore atomic double-booking protection on relief assignments (#414)

guangshinhaha:
Closes #414

**Stack position: 4/12** — base `fix/slug-route-role-gates` (PR #428). Merge after #428.

## Summary

- **Why the index was dropped (grooming step from the issue):** the 2026-06-12 overrides migration added `status` (cancelled rows kept as history) and the `replacedById` chain, and #274 keeps declined rows ACTIVE with `declinedAt` set — so multiple rows per `(timetableEntryId, date)` are legitimate. The true invariant is *one ACTIVE, non-declined row per slot*, which is a partial-index shape Prisma can't declare in `schema.prisma` (and hand-written migration SQL is forbidden). Hence the **app-level transactional guard** (as agreed).
- Mechanism: inside the create transaction, take `pg_advisory_xact_lock` on the slot key (manual route also locks the reliever+date key; all lockers sort keys, so no deadlock), then re-check for a live holder. Race loser → 409 `STALE_DASHBOARD`, matching the existing pre-check semantics. The pre-checks stay for their fast, attributed error messages.
- Also fixed: the auto-assign confirm path had **no slot check at all** since the index drop — its P2002 catch was dead code, so simply re-submitting the same confirm batch (double-click, stale tab) silently double-booked. It now locks every slot in the batch and rejects if any has a live holder.

## Changes

- `src/lib/scheduling/assignmentConstraints.ts` — `lockAssignmentKeys()` (sorted advisory xact locks; `::text` cast because Prisma can't deserialize `void`), lock-key builders, `BookingConflictError`.
- `src/app/api/relief-assignments/route.ts` — in-transaction lock + re-check of both `slotTaken` and `teacherBusy`; header comment rewritten to describe the new guarantee.
- `src/app/api/auto-assign/route.ts` — array-form transaction → interactive transaction with slot locks + live-holder check; `BookingConflictError` → 409.

## Test plan

- New tests: two concurrent same-slot POSTs → exactly `[201, 409]`, one ACTIVE row; two concurrent same-reliever/same-period POSTs → `[201, 409]`; cancelled history row does **not** block re-booking (the override scenario that motivated the index drop); duplicate auto-assign confirm → 409, one row.
- relief-assignments + auto-assign + relief-assignments-delete: **36/36 pass** (all #274 declined-slot and Phase 15.2 attribution tests unaffected).
- `npm test`: 727 passed. `npm run build`: clean 88/88.


guangshinhaha:
Review finding: the batch confirm path in `/api/auto-assign?preview=false` locks/rechecks only slot keys, and it does that before creating rows. A crafted `body.assignments` can include the same `timetableEntryId` twice in one batch, or assign the same reliever to two different entries in the same period, and both rows will be created because no in-batch duplicate/teacher-period check runs. I am patching this on the stack.

src/app/api/auto-assign/route.ts
```diff
@@ -185,20 +188,51 @@ export async function POST(request: NextRequest) {
...
-          );
+          // #414 — atomic double-booking guard, mirroring the manual route.
+          // The old array-form transaction relied on a P2002 from the unique
+          // (timetableEntryId, date) index, which the timetable-overrides
+          // migration dropped — so a re-submitted or concurrent confirm
+          // silently double-booked slots. Lock every slot key, re-check that
+          // none has a live holder, then create.
+          const created = await prisma.$transaction(async (tx) => {
+            await lockAssignmentKeys(
+              tx,
```
guangshinhaha: **High** — Auto-assign confirm path does not lock teacher keys, leaving a teacher-busy double-booking gap

The manual relief-assignments route correctly locks both slot and teacher keys:
```ts
await lockAssignmentKeys(tx, [
  assignmentSlotLockKey(timetableEntryId, date),
  assignmentTeacherLockKey(reliefTeacherId, date),  // ← teacher key locked
]);
```

But the auto-assign confirm path only locks slot keys:
```ts
await lockAssignmentKeys(
  tx,
  entryIds.map((id) => assignmentSlotLockKey(id, dateUTC)),  // ← no teacher keys
);
```

And the in-transaction re-check only verifies `SLOT_TAKEN` — there is no `TEACHER_BUSY` re-check.

**Failure scenario:** KP-A does a manual assign of Teacher X to Period 3 (slot A). Concurrently, KP-B confirms an auto-assign batch that also assigns Teacher X to Period 3 (slot B, different class). The manual route locks `relief-slot:A` + `relief-teacher:X`. The auto-assign route locks only `relief-slot:B`. No serialization occurs on the teacher key. Both transactions commit. Teacher X is now double-booked in Period 3.

This can also happen with two concurrent auto-assign confirms if their batches share a reliever in overlapping periods (client-edited batches via `body.assignments` are especially susceptible).

**Suggested fix — also lock teacher keys and re-check teacher-busy in the auto-assign transaction:**
```ts
const teacherIds = [...new Set(assignments.map((a) => a.reliefTeacherId))];
await lockAssignmentKeys(tx, [
  ...entryIds.map((id) => assignmentSlotLockKey(id, dateUTC)),
  ...teacherIds.map((id) => assignmentTeacherLockKey(id, dateUTC)),
]);

// After the SLOT_TAKEN check, also re-check teacher-busy:
// (Need to look up periodIds for each entry to check same-period conflicts)
```

Note: the teacher-busy lock key is per teacher+date (not per teacher+period), so it serializes broadly enough — the re-check query narrows to the specific period.

---

src/lib/scheduling/assignmentConstraints.ts
```diff
@@ -46,6 +46,58 @@ export function assignmentConstraintViolation(input: {
...
+ */
+export async function lockAssignmentKeys(
+  tx: Prisma.TransactionClient,
+  keys: string[],
+): Promise<void> {
+  const sorted = [...new Set(keys)].sort();
+  for (const key of sorted) {
+    // ::text — pg_advisory_xact_lock returns `void`, which Prisma's raw
+    // deserializer rejects; the cast makes the result a readable row.
+    await tx.$queryRaw`SELECT pg_advisory_xact_lock(hashtextextended(${key}, 0))::text`;
```
guangshinhaha, line 98: **Low** — `hashtextextended` requires PostgreSQL 13+

```ts
await tx.$queryRaw`SELECT pg_advisory_xact_lock(hashtextextended(${key}, 0))::text`;
//                                               ^^^^^^^^^^^^^^^^^ PG13+ only
```

This function was introduced in PostgreSQL 13. If the project ever targets PG 12 or earlier (unlikely given Docker Compose defaults, but worth documenting), this will fail with `function hashtextextended(text, integer) does not exist`.

The project Docker Compose and RDS targets almost certainly run PG 14+, so this is informational — but a one-line comment noting the PG 13 floor would save a future reader from a puzzling failure if the Docker image version is ever downgraded.

No action needed if PG 13+ is already the documented minimum.

---

guangshinhaha (commented):
## Code Review Summary — PR #429

### What Looks Good

- **Correct advisory-lock pattern for the manual route**: slot + teacher keys locked in sorted order, re-checked inside the transaction, `BookingConflictError` cleanly propagated to 409. This closes the TOCTOU gap documented in the header comment.
- **Well-factored shared primitives**: `lockAssignmentKeys`, `assignmentSlotLockKey`, `assignmentTeacherLockKey`, and `BookingConflictError` are cleanly extracted into `assignmentConstraints.ts` with thorough doc comments explaining why the advisory-lock approach was chosen over a partial unique index.
- **Thorough test coverage**: concurrent same-slot, concurrent same-reliever, cancelled-history-row re-booking, and duplicate-confirm tests all exercise the race conditions meaningfully (via `Promise.all`). The cancelled-row test is especially important — it validates the exact scenario that motivated dropping the strict unique index.
- **Pre-existing pre-checks preserved**: the fast, attributed 409 responses for the common stale-dashboard case still fire before entering the transaction, keeping the two-layer defense (fast pre-check + atomic guard) well separated with clear comments.

### Findings

| 1 | **High** | `src/app/api/auto-assign/route.ts` | Auto-assign confirm locks only slot keys — no teacher-busy lock or re-check. A concurrent manual + auto-assign (or two auto-assign confirms with client-edited batches) can double-book a teacher in the same period. |
| 2 | **Low** | `src/lib/scheduling/assignmentConstraints.ts` | `hashtextextended` is PG 13+. Informational — likely fine for this project, but a comment noting the floor helps future readers. |

Finding #1 should be addressed before merge to maintain parity with the manual route's guarantees.

---

Issue #417, opened 2026-06-29
bug: clicking school in school overrides (under admin page) opens demo dashboard


superb-sushi:
## Steps to reproduce

<img width="990" height="332" alt="Image" src="https://github.com/user-attachments/assets/1baa348a-3285-482b-9651-0b3f4c0338af" />

1. Sign in as a SUPER_ADMIN
2. On the `admin` page, navigate to `Platform Config` > `Algorithm & Matching`
3. Scroll to the `School Overrides` section, and click on the `view` button of any of the schools listed there

## Expected behaviour
Clicking on a school should display the school's current auto-assign rules configuration, displaying that school's customised rules and configuration instead of the demo dashboard.

## Actual behaviour
On the Admin → School Overrides page, clicking on a school to view its customised rules redirects the admin to the demo dashboard populated with sample data.

guangshinhaha:
Closed by PR chain #426–#437.

Issue #419, opened 2026-07-02
fix(`security`): enforce role check on `/api/[slug]/timetable-overrides` and `/api/[slug]/rems-candidates` (teacher can access admin-only data)


guangshinhaha:
## User story

As a school admin, I want admin-only scheduling endpoints to reject `TEACHER`-role sessions, so that a teacher in my own school cannot read or modify relief-planning configuration reserved for administrators.

## Background

The MAIA penetration test (run against `dev-reliefcher.edutech.works`) found that two school-scoped API routes enforce **tenant** isolation (school-slug / `schoolId` match) but do **not** enforce a **role** check — so any `TEACHER`-role session in the same school can reach admin-only data (vertical privilege escalation, CWE-639):

- **`/api/[slug]/timetable-overrides`** — *pentest #2 (High)*. `GET` lists all override configs and `POST` creates overrides for arbitrary date ranges. Overrides drive the school's relief-planning logic, so a malicious/compromised teacher can read or inject them. Secondary bug: a `SUPERADMIN` session is wrongly **rejected** (403) because the check requires a school-scoped session.
- **`/api/[slug]/rems-candidates`** — *pentest #18 (Medium)*. `GET` returns the full REMS candidate pool (potential PII) to any teacher.

**Verified still present on branch `fix/teachers-api-pii-leak` (2026-07-02):**
- `src/app/api/[slug]/timetable-overrides/route.ts` — GET (~line 13) and POST (~line 41) only check `session.user.schoolId !== school.id`; no capability/role gate.
- `src/app/api/[slug]/rems-candidates/route.ts` — GET (~line 31) only checks the `schoolId` match.

The **correct pattern already exists in sibling routes** (these were fixed and are not part of this ticket): `src/app/api/[slug]/preferred-teachers/route.ts` gates with `can(session.user.role, CAPABILITIES.ASSIGN_EXTERNAL_RELIEF)`, and `/api/dashboard` uses `requireCapability(VIEW_DASHBOARD)`. `TEACHER` has an empty capability set. Helpers: `src/lib/auth.ts` (`requireCapability`, `requireSchoolAdmin`) and `src/lib/capabilities.ts` (`can`).

### Teacher cannot read or write timetable-overrides
- **Given** a `TEACHER`-role session in school X
- **When** it calls `GET` or `POST` `/api/X/timetable-overrides`
- **Then** the response is `403` and no override is read or created

### Teacher cannot read rems-candidates
[...]

PR #430, opened 2026-07-02
fix(admin): School Overrides View link no longer bounces superadmin to the demo dashboard (#417)

guangshinhaha:
Closes #417

**Stack position: 5/12** — base `fix/relief-double-booking-guard` (PR #429). Merge after #429.

## Summary

- Root cause: the View link (`/{slug}/dashboard/settings#auto-assign-rules`, [AlgorithmPage.tsx:171](src/app/admin/algorithm/AlgorithmPage.tsx)) is fine — but `resolveSchoolSlugAccess` had no SUPERADMIN case. A superadmin session never matches the school slug, so `requireSchoolForSlugOrRedirect` hit the `wrong-slug` branch and bounced to the superadmin's "own" dashboard — the **demo** school for the seeded superadmin, exactly the reported symptom (school-less superadmins were bounced to /login instead).
- Fix: a new `superadmin` decision in the pure `resolveSchoolSlugAccess`; `requireSchoolForSlugOrRedirect` resolves the school **from the URL slug** for SUPERADMIN and renders the requested school's page (unknown slug → `/admin`). This is the page-side mirror of the SUPERADMIN passthrough #419 added to the API-side guard, and it fixes the same bounce on every other slug-scoped dashboard page a superadmin drills into, not just settings.

## Changes

- `src/lib/auth.ts` — `SlugAccessDecision` + `resolveSchoolSlugAccess` + `requireSchoolForSlugOrRedirect`.
- `src/__tests__/slugAccess.test.ts` — school-less superadmin → `superadmin`; demo-bound superadmin on a different slug → `superadmin` (the reported repro).

## Test plan

- `npm test`: 729 passed (2 new). `npm run test:api` touched file `relief-ops-access.test.ts`: pass. `npm run build`: clean 88/88.
- Manual: as SUPERADMIN, Admin → Platform Config → Algorithm & Matching → School Overrides → View now opens that school's settings with its rules section (worth a click-through on the test env after deploy).


guangshinhaha (commented):
## Code Review — PR #430

### What Looks Good

- Clean use of a discriminated union (`SlugAccessDecision`) to extend behavior — keeps the pure function unit-testable and the effectful function handling the new case at the right layer.
- The SUPERADMIN check is placed before the school-bound guards, so school-less superadmins are never misrouted through the `login` or `wrong-slug` branches.
- The two new test cases cover both the school-less superadmin and the demo-school-bound superadmin (the exact repro from #417).
- The superadmin branch returns `{ session, schoolId, school }` matching the existing contract, so all 4 page-level callers continue to work without changes.

### Findings

No critical or high-severity issues found. Two minor observations below (posted as inline comments).

---

src/lib/auth.ts
```diff
@@ -132,19 +132,27 @@ export const requireSchoolForSlug = cache(async (slug: string) => {
...
   | { kind: "login" }
-  | { kind: "wrong-slug"; ownSlug: string };
+  | { kind: "wrong-slug"; ownSlug: string }
+  | { kind: "superadmin" };
 
 export function resolveSchoolSlugAccess(
-  session: { user: { schoolId: string | null; school: { slug: string } | null } } | null,
+  session: {
+    user: { role: string; schoolId: string | null; school: { slug: string } | null };
+  } | null,
```
guangshinhaha, line 151: **Low · Dead-code guard**

The `!session` check here can never be true. When `decision.kind === "superadmin"`, `resolveSchoolSlugAccess` has already confirmed `session` is non-null (the first guard returns `{ kind: "expire" }` for null sessions). The `!school` check is the only one doing real work.

Not a bug — just dead code that could confuse a future reader into thinking a null session is reachable at this point.

```ts
    // Before (current)
    if (!school || !session) redirect("/admin");

    // After — drop the unreachable branch
    if (!school) redirect("/admin");
```

---

guangshinhaha, line 151: **Medium · Missing integration-test coverage for the superadmin redirect paths**

The pure `resolveSchoolSlugAccess` is well-covered by the two new unit tests. But the async `requireSchoolForSlugOrRedirect` now has a meaningful new branch with its own DB lookup and redirect logic that has no automated test:

1. **Valid slug** — superadmin resolves the school from the URL and gets `{ session, schoolId, school }`.
2. **Invalid slug** — superadmin is redirected to `/admin`.

Since this is a page-level guard (not an API route), a full integration test is admittedly heavy. But even a lightweight test that mocks `getSession` and `prisma.school.findUnique` would catch regressions if the guard logic is refactored later. Consider adding to `tests/api/` or a new `src/__tests__/slugAccessIntegration.test.ts` at some point.

Not a merge blocker — the unit tests + manual verification in the test plan are sufficient for now.

---

Issue #422, opened 2026-07-02
fix(`security`): bind session school to URL slug on `/[slug]/teacher` and `dashboard/relief-teachers`


guangshinhaha:
## User story

As a school user, I want slug-scoped pages to reject a URL whose school slug isn't my own, so that I cannot render another school's teacher portal or relief-teachers page.

## Background

MAIA pentest **#5 (Medium)** with residual **#4**. Most `/[slug]/dashboard/*` pages were fixed to call `requireSchoolForSlugOrRedirect(slug)`, but two routes were missed — so a user of school A can load school B's slug and get a fully-rendered foreign page. Because `SchoolProvider` is initialized from the **URL slug**, the page's client-side fetches run in the foreign school's context (the realistic cross-tenant entry point).

**Verified on `fix/teachers-api-pii-leak` (2026-07-02):**
- `src/app/[slug]/teacher/page.tsx` and `src/app/[slug]/teacher/layout.tsx` — **no** slug/session validation. `src/middleware.ts` (teacher branch, ~lines 56–64) only checks session existence, not slug match. **Still vulnerable.**
- `src/app/[slug]/dashboard/relief-teachers/page.tsx` — only `getSession()`; `await params` is discarded, no `requireSchoolForSlugOrRedirect`. **Still vulnerable.**
- `src/app/[slug]/layout.tsx` — `SchoolProvider` falls back to `getSchoolBySlug(slug)` when the session school ≠ slug, so it is **slug-derived**, not session-derived.
- Fixed reference pattern: the other dashboard pages (`page`/`settings`/`staff`/`timetable`) all call `requireSchoolForSlugOrRedirect(slug)`.
- Partial mitigation: teacher data APIs (`/api/teacher-dashboard`) resolve the teacher by `session.user.id`, so direct API data is session-scoped — but the foreign render + slug-derived provider is still a leak/entry point.

### Teacher route enforces slug binding
- **Given** a session for school X → `GET /Y/teacher` redirects to the user's own slug (or `403`); `GET /X/teacher` renders normally

### relief-teachers page enforces slug binding
- **Given** a session for school X → `GET /Y/dashboard/relief-teachers` redirects/`403`; `/X/...` renders

### SchoolProvider is session-authoritative
- When session school ≠ URL slug, the request is redirected **before** `SchoolProvider` is initialized with foreign data

## Out of scope

- The already-fixed dashboard pages.
[...]

PR #431, opened 2026-07-02
fix(security): validate teacherId/periodId belong to the session school on POST /api/timetable (#421)

guangshinhaha:
Closes #421

**Stack position: 6/12** — base `fix/school-overrides-view-link` (PR #430). Merge after #430.

## Summary

- MAIA pentest #11: `POST /api/timetable` wrote entries with the session's `schoolId` but trusted the body's `teacherId`/`periodId` — cross-tenant write IDOR. Both IDs are now asserted to resolve within the session school before create/upsert.
- One deliberate deviation from the ticket: rejection is **404**, not 403 — the codebase's anti-enumeration convention for foreign IDs (`relief-assignments` POST, `timetable/[id]` DELETE all 404). A 403 would confirm to school A that the guessed ID exists in some other school.
- Scope check per the ticket's "same check on PUT/PATCH": `timetable/[id]` has only DELETE, which already scopes with `where: { id, schoolId }` — no change needed.

## Changes

- `src/app/api/timetable/route.ts` — tenancy assertions on both FKs before the upsert.
- `tests/api/timetable.test.ts` — foreign `teacherId` → 404 + nothing persisted; foreign `periodId` → 404 + nothing persisted; same-school → 201.

## Test plan

- `timetable.test.ts` + `import-timetable.test.ts`: **33/33 pass**.
- `npm test`: 729 passed. `npm run build`: clean 88/88.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #377, opened 2026-06-24
Reporting page shows misleading MOE iCON sign-in prompt to logged-in teachers from another school


guangshinhaha:
## Summary

On `/{slug}/reporting`, a user who **is** authenticated but is not a teacher/admin of `{slug}`'s school is shown the same "Sign in with MOE iCON" prompt as a fully logged-out visitor. This is misleading: the user is already signed in, and the real reason for the block is that they don't belong to this school's roster (cross-tenant access denial).

## How it surfaced (pilot)

A maia pilot tester logged in via OTP as `maia-teacher-1a@gmail.com` (a valid **TEACHER** of `maia-pri-1`) and opened `/maia-sec-2/reporting`. The page showed "maia-sec-2 requires teachers to sign in with their MOE iCON account…", which read as a login/cookie bug. In fact login succeeded; the account simply isn't on maia-sec-2's roster, so the multi-tenancy gate (correctly) denied access — but with the wrong message.

## Root cause

[`src/app/[slug]/reporting/page.tsx`](src/app/[slug]/reporting/page.tsx) computes `isAuthorized` from two conditions:

- a `Teacher` linked to the session user **whose `schoolId === this school`**, or
- a `SCHOOL_ADMIN` of this school.

When neither holds, it renders the **sign-in prompt** — without distinguishing:
1. **no session at all** (genuinely needs to sign in), from
2. **valid session, wrong/other school** (signed in, but not authorized here).

Login is global (not per-school), so blocking login is not the right fix — the authorization *messaging* is.

## Expected behaviour

- **No session** → keep the current "Sign in with MOE iCON" prompt.
- **Authenticated but not a member of this school** → show an authorization message instead, e.g. "You're signed in as `<email>`, but this account isn't a teacher or admin of `<school name>`." Offer **Sign out / switch account**, and (where relevant) a link to the user's own school reporting page.

## Scope / acceptance

- Branch the unauthorized render on `session?.user` presence.
- New "wrong school / not on roster" state for authenticated-but-unauthorized users; no behaviour change for logged-out visitors.
- No data leakage (still must not render roster/periods to unauthorized users).
- Unit/integration coverage for: logged-out, teacher-of-other-school, admin-of-other-school, teacher-of-this-school (happy path).

## Notes

- Behaviour of the gate itself (deny cross-tenant access) is **correct** — this is a UX/messaging fix only.
[...]

PR #432, opened 2026-07-02
fix(security): bind session school to URL slug on /[slug]/teacher and dashboard/relief-teachers (#422)

guangshinhaha:
Closes #422

**Stack position: 7/12** — base `fix/timetable-cross-tenant-write` (PR #431). Merge after #431.

## Summary

- The two routes the pentest-#5 remediation missed now call `requireSchoolForSlugOrRedirect(slug)`: the **teacher subtree** (enforced in `[slug]/teacher/layout.tsx`, so every current and future teacher page is covered) and **`dashboard/relief-teachers/page.tsx`** (which had `getSession()` only and discarded `params`).
- A school-X user loading school Y's URL is redirected to their own dashboard before any foreign render; teachers get routed onward to their own `/teacher` surface by the dashboard layout's existing role bounce. SUPERADMIN passes through via the #417 decision.
- **AC deviation, deliberate:** the issue suggested also enforcing in `[slug]/layout.tsx` so `SchoolProvider` can never initialize from a foreign slug. That layout also hosts the **public reporting surface**, whose "signed in, wrong school" behavior (#377/#409) intentionally renders rather than redirects — a blanket layout redirect would regress it. With every *protected* page now gated before render, no foreign-school render (and no provider-context fetch) can reach the client, which satisfies the AC's intent.

## Changes

- `src/app/[slug]/teacher/layout.tsx` — async + slug guard.
- `src/app/[slug]/dashboard/relief-teachers/page.tsx` — slug guard replaces the bare session check.
- `tests/api/slug-page-binding.test.ts` — new: invokes the server components directly with DB-backed sessions; cross-slug → `NEXT_REDIRECT` to own dashboard, own slug → renders, no session → login redirect.

## Test plan

- New binding tests: **5/5 pass**. `npm test`: 729. `npm run build`: clean 88/88.


guangshinhaha (commented):
## Code Review Summary

### What Looks Good

- Correctly applies the established `requireSchoolForSlugOrRedirect` pattern to the two missed routes, closing the pentest gap with minimal, focused changes.
- Guarding at the teacher layout level is the right architectural choice -- it covers all current and future child pages under `/[slug]/teacher/*` without requiring per-page guards.
- The `relief-teachers/page.tsx` change is a net simplification: fewer imports, no manual redirect logic, and the slug is now actually used instead of discarded.
- The test file is well-structured with realistic DB-backed sessions, covering cross-slug redirect, own-slug render, and unauthenticated redirect.

### Findings

See inline comments.

---

tests/api/slug-page-binding.test.ts
```diff
@@ -0,0 +1,115 @@
...
+
+  it("renders for the user's own slug", async () => {
+    const own = await createSchool({ slug: "own-school" });
+    const user = await createUser({ schoolId: own.id });
+    await mockSession(await createSession(user.id));
+
+    const rendered = await ReliefTeachersPage(params("own-school"));
+    expect(rendered).toBeTruthy();
+  });
+});
```
guangshinhaha, line 115: **Medium** — Missing SUPERADMIN test coverage for a security-scoped PR

The PR body states "SUPERADMIN passes through via the #417 decision," but there is no test case documenting the expected SUPERADMIN behavior. Looking at `resolveSchoolSlugAccess`:

```ts
export function resolveSchoolSlugAccess(
  session: { user: { schoolId: string | null; school: { slug: string } | null } } | null,
  slug: string,
): SlugAccessDecision {
  if (!session) return { kind: "expire" };
  if (!session.user.schoolId || !session.user.school) return { kind: "login" }; // ←
  if (session.user.school.slug !== slug) {
    return { kind: "wrong-slug", ownSlug: session.user.school.slug };
  }
  return { kind: "ok" };
}
```

A SUPERADMIN without a `schoolId` (which is how they are seeded — see `prisma/seed.ts:240`) hits `{ kind: "login" }` and gets redirected to `/login`. This is consistent with the other dashboard pages (they all use the same guard), so it is not a regression — but a test case would document the expected behavior and catch any future change to SUPERADMIN school assignment.

Suggested addition to the test file:

```ts
it("redirects a school-less SUPERADMIN to /login (no schoolId)", async () => {
  await createSchool({ slug: "any-school" });
  const admin = await createUser({ role: UserRole.SUPERADMIN }); // no schoolId
  await mockSession(await createSession(admin.id));

  await expectRedirect(
    TeacherLayout({ children: null, ...params("any-school") }),
    "/login",
  );
});
```

This clarifies whether SUPERADMIN passthrough depends on impersonation (which grants a `schoolId` via the impersonated user) or on some other mechanism. Can be deferred if the SUPERADMIN-without-school flow is intentionally handled elsewhere.

---

Issue #423, opened 2026-07-02
fix(`security`): add semantic date validation to /api/absence-reports and /api/relief-assignments


guangshinhaha:
## User story

As an admin relying on relief reports, I want absence and relief-assignment dates validated against basic business rules, so that phantom records (backdated, far-future, weekend, or out-of-range dates, and duplicates) can't pollute attendance stats and the matching algorithm.

## Background

MAIA pentest **#14 & #15 (Medium, CWE-840)** — business-logic validation gaps.

- `POST /api/absence-reports` (`src/app/api/absence-reports/route.ts`): enforces only `endDate >= startDate` (~line 201) and a max span (180 days). It accepts far-past dates (2020), far-future dates (2099), weekend dates, and unlimited exact duplicates (no uniqueness / idempotency).
- `POST /api/relief-assignments` (`src/app/api/relief-assignments/route.ts`): accepts past (1990), weekend, and far-future (2030) dates, and dates that don't fall within the linked absence report's range (the report query selects only `id`, ~lines 50–53). Malformed→400, missing report→404 and duplicate-tuple→409 are enforced, but the 409 is racy (unique index removed 2026-06-12, ~lines 14–22).

**Verified still present on `fix/teachers-api-pii-leak` (2026-07-02).**

### Absence reports reject out-of-range / weekend / duplicate
- **Given** a teacher session
- **When** `POST /api/absence-reports` with a past-bound (before a reasonable academic-year window), far-future, or weekend date → `400`
- **And** when an identical `(teacherId, startDate, endDate)` report already exists → rejected/idempotent (no duplicate row)

### Relief assignment dates are validated
- **When** `POST /api/relief-assignments` with a weekend, far-past/future date, or a date outside the linked absence report's `[startDate, endDate]` → `400`
- Valid weekday within the absence range → `201` as before

## Out of scope

- Concurrency-race double-booking / restoring the unique index — tracked in **#414**. This ticket is date/dedup **semantic** validation only; coordinate with #414 to avoid conflicting fixes.
- Rate-limiting — tracked in **#413**.

## Technical context

- Add shared date validators: weekend check via `getUTCDay()`, bounds against a reasonable window / academic year, and absence-range containment for relief dates.
- For absence duplicates, add an application-level check or a unique constraint on `(teacherId, startDate, endDate)` via `npx prisma migrate dev` (schema change only — no hand-written migration SQL, per CLAUDE.md).
[...]

PR #433, opened 2026-07-02
fix(security): add semantic date validation to absence-reports and relief-assignments (#423)

guangshinhaha:
Closes #423

**Stack position: 8/12** — base `fix/slug-session-binding-pages` (PR #432). Merge after #432.

## Summary

- Shared validators in `src/lib/scheduling/dateValidation.ts`: **weekday-only** + a window of **[Jan 1 of the current year, today + 365d]**. I chose the school-year start over a rolling −30d past bound: backfilling earlier terms of the running year is a real admin workflow (and a rolling bound would also have broken every existing hardcoded test fixture mid-year).
- `POST /api/absence-reports`: weekend/out-of-window start or end date → 400; identical `(teacher, startDate, endDate)` resubmission → **409 `DUPLICATE_REPORT`** (application-level check — the racy-409/unique-index concern stays with #414 as scoped).
- `POST /api/relief-assignments`: weekend/out-of-window/malformed date → 400; date **outside the linked absence report's `[startDate, endDate]`** → 400 (the report query now selects the range instead of only `id`).
- New tests compute dates relative to *now*, so they can't age out of the window (the suite's hardcoded `2026-04-07` fixtures stay valid under the year-start bound, but will need the same dynamic treatment when the year rolls over — noted below).

## Found while testing (pre-existing, not fixed here)

The absence-reports route's fallback `reason = "MEDICAL_LEAVE"` is **not a valid `AbsenceReason` enum value**, so any request omitting `reason` 400s with "Invalid reason." — this is the actual root cause of most of the 17 pre-existing `absence-reports.test.ts` baseline failures (they predate this stack). Flagging separately rather than widening this PR.

## Test plan

- `tests/api/semantic-date-validation.test.ts`: **10/10** (5 absence + 5 relief cases incl. happy paths).
- `src/__tests__/dateValidation.test.ts`: 9 unit tests with injected "now".
- `absence-reports.test.ts`: still exactly its **17 pre-existing** failures — no new ones. relief-assignments + auto-assign: all pass.
- `npm test`: 738 passed. `npm run build`: clean 88/88.


guangshinhaha:
Review finding: the semantic date validation added for manual `/api/relief-assignments` does not cover the auto-assign confirm path. A crafted `/api/auto-assign?preview=false` request with client-supplied `assignments` can still create weekend, far-future, or out-of-absence-range relief assignments. I am patching this on the stack.

src/lib/scheduling/dateValidation.ts
```diff
@@ -0,0 +1,55 @@
...
+ * All checks operate on UTC dates — the app stores dates as UTC midnight.
+ */
+
+const DAY_MS = 24 * 60 * 60 * 1000;
+export const FUTURE_WINDOW_DAYS = 365;
+
+export function isWeekendUTC(date: Date): boolean {
+  const day = date.getUTCDay();
+  return day === 0 || day === 6;
+}
```
guangshinhaha: **Medium · Reuse** — `isWeekendUTC` duplicates `isWeekend` in `src/lib/scheduling/date.ts`

`date.ts` already exports an `isWeekend(dateStr: string)` that does the same weekday check via `getDay()` on a local-parsed date. Since all dates in this codebase are stored as UTC midnight, both functions produce the same result — but having two weekend checkers in sibling files (`date.ts` and `dateValidation.ts`) under `src/lib/scheduling/` invites future divergence (e.g., if someone adds a public-holiday overlay to one but not the other).

**Scenario:** A future contributor adds a holiday calendar to `isWeekend` in `date.ts` but misses `isWeekendUTC` — the dashboard skips holidays but the validation layer still allows them.

**Suggestion:** Reuse the existing function by accepting a `Date` overload in `date.ts`, or have `dateValidation.ts` import from `date.ts` and convert:

```ts
import { isWeekend } from "./date";

// Inside semanticDateViolation:
const isoStr = date.toISOString().slice(0, 10);
if (isWeekend(isoStr)) return "Date falls on a weekend.";
```

Alternatively, move the `Date`-based variant into `date.ts` as the canonical implementation and make the string-based one delegate to it.

---

src/app/api/absence-reports/route.ts
```diff
@@ -226,6 +227,19 @@ export async function POST(request: NextRequest) {
...
+    // ── Semantic date validation (#423) ──
+    // A multi-day range may span a weekend, but its endpoints must be school
+    // days and both must fall inside the accepted window.
+    for (const [label, d] of [["startDate", start], ["endDate", end]] as const) {
+      const violation = semanticDateViolation(d);
+      if (violation) {
+        return NextResponse.json(
+          { error: `Invalid ${label}: ${violation}` },
+          { status: 400 }
+        );
```
guangshinhaha, line 246: **Medium · Cross-file / Defence-in-depth** — `startDate` NaN is not guarded before the semantic check

The existing code on line 194 does `const start = new Date(startDate + "T00:00:00.000Z")` — if `startDate` is a truthy non-date string like `"abc"`, `start` becomes `Invalid Date`. The existing guards that follow do *not* catch this:

- `end < start` → `false` (NaN comparisons always false)
- `spanDays > MAX_DAYS` → `false` (NaN arithmetic)

So a NaN `start` would have silently proceeded to the DB write before this PR. The new `semanticDateViolation` call now catches it — which is good — but the ordering means the NaN protection is an *implicit side-effect* of the weekend/window check rather than an *explicit* structural guard.

This is not a bug introduced by this PR (it actually *fixes* a pre-existing gap), but it would be more robust to add an explicit NaN check on `start` right after line 194, next to the existing NaN check on `end` at line 201, so the defence doesn't depend on the semantic validation existing:

```ts
const start = new Date(startDate + "T00:00:00.000Z");
if (Number.isNaN(start.getTime())) {
  return NextResponse.json(
    { error: "Invalid startDate." },
    { status: 400 }
  );
}
```

This is pre-existing tech debt surfaced by the review — fine to address in a follow-up if you prefer not to widen this PR.

---

guangshinhaha (commented):
## Code Review — PR #433

### What Looks Good

- **Shared validation module** — centralising weekend + window checks in `src/lib/scheduling/dateValidation.ts` with injectable `now` is clean, testable design. Other routes can adopt it trivially.
- **Duplicate-report guard** — the application-level `409 DUPLICATE_REPORT` on `(teacher, startDate, endDate)` prevents phantom attendance-stat pollution without requiring a schema migration for a unique index (which is rightly scoped to #414).
- **Relief-date-inside-absence-range check** — verifying `date` falls within the linked absence report's `[startDate, endDate]` closes a real logic gap where a caller could create coverage for a day the teacher isn't absent.
- **Dynamic test dates** — `weekdayISO()`, `saturdayISO()`, `lastYearISO()`, `farFutureISO()` ensure the integration tests cannot age out of the validation window, avoiding the class of test-rot that hit the hardcoded `2026-04-07` fixtures.

### Findings

See inline comments below.

---

Issue #420, opened 2026-07-02
fix(`security`): neutralize CSV formula injection in teacher and lead exports


guangshinhaha:
## User story

As an admin who exports data to CSV, I want cell values that begin with formula characters to be neutralized, so that opening an export in Excel/LibreOffice cannot execute an attacker-supplied formula on my machine.

## Background

MAIA pentest **#3 (High, CWE-1236)**. `GET /api/teachers/export` returns CSV where cell values starting with `=`, `+`, `-`, `@` (or tab / carriage return) are written verbatim. A teacher/lead record whose `Name` is e.g. `=cmd|'/C calc'!A0` triggers DDE / `WEBSERVICE` formula execution when the export is opened in a spreadsheet application (command execution / data exfiltration on the victim's machine).

**Verified still present on `fix/teachers-api-pii-leak` (2026-07-02):**
- `src/lib/teacherCsv.ts` — `escapeCsvField()` (~lines 257–262) only quotes cells containing `,` / `"` / `\n` (RFC 4180). It does **not** neutralize formula-trigger characters. Consumed by `src/app/api/teachers/export/route.ts`.
- `src/app/api/admin/leads/export/route.ts` — `escapeCsv()` (~lines 14–19) shares the identical flawed pattern; the `status` field isn't escaped at all. **Same root cause, second endpoint.**
- Not vulnerable: `src/app/api/export/route.ts` wraps **all** cells in quotes, so formulas render as literal text — but standardizing on one helper is cleaner.

Formula strings currently also persist in the teacher `name` field and render in SSR, so output-side neutralization is the reliable fix.

### Formula-trigger cells are neutralized on export
- **Given** a teacher/lead whose exported cell begins with `=`, `+`, `-`, `@`, tab, or CR
- **When** the CSV is generated by `/api/teachers/export` or `/api/admin/leads/export`
- **Then** the cell is prefixed with a single apostrophe (or otherwise neutralized) and escaped per RFC 4180, so a spreadsheet renders it as text — not a formula

### No regression for benign values
- **Given** ordinary values → exported unchanged (aside from standard quoting)

## Out of scope

- Rejecting formula-like input at teacher/lead **creation** (defense-in-depth; can be a follow-up).

## Technical context

[...]

PR #434, opened 2026-07-02
fix(security): neutralize CSV formula injection across all export routes (#420)

guangshinhaha:
Closes #420

**Stack position: 9/12** — base `fix/semantic-date-validation` (PR #433). Merge after #433.

## Summary

- Shared `escapeCsvCell()` in `src/lib/csv.ts` (the repo's single-owner CSV module): apostrophe-prefix when the first char is one of `= + - @ \t \r`, then RFC-4180 quoting. All three export paths now use it:
  - `src/lib/teacherCsv.ts` (→ `/api/teachers/export`) — local formula-blind `escapeCsvField` deleted.
  - `/api/admin/leads/export` — duplicate local escaper deleted; the `status` field, previously written raw, is now escaped too.
  - `/api/export` — kept safe-ish by quote-everything before, but quoting alone doesn't stop Excel evaluating a leading `=`; standardized on the shared helper per the ticket.
- Known trade-off (standard OWASP behavior): values legitimately starting with `+` (e.g. `+65` phone numbers) gain a leading apostrophe in the raw CSV; Excel hides it on display. If a round-trip re-import of exported phones matters, the import side can strip a leading apostrophe as a follow-up.

## Test plan

- `src/__tests__/escapeCsvCell.test.ts` — all trigger chars, benign values unchanged, RFC-4180 cases, combined quote+neutralise, null/number inputs, round-trip through the sibling `parseCsv`.
- `export.test.ts` + `teachers-export.test.ts` + `import-timetable.test.ts`: **27/27 pass**.
- `npm test`: 744 passed. `npm run build`: clean 88/88.


guangshinhaha (commented):
## Code Review — PR #434: CSV Formula Injection Fix

### What Looks Good

- **Single-owner pattern is the right call.** Consolidating three independent CSV escape functions into one shared `escapeCsvCell` in `src/lib/csv.ts` eliminates the drift risk that caused the `lead.status` field to be unescaped in the first place.
- **OWASP-standard mitigation.** The apostrophe-prefix approach for formula-trigger characters (`= + - @ \t \r`) is the industry-standard defence against CWE-1236; the implementation matches the OWASP CSV Injection cheat sheet.
- **Good test coverage.** The test file covers all trigger characters, benign passthrough, RFC-4180 quoting, combined quote+neutralise, null/undefined/number coercion, and round-trip through the sibling parser.
- **Incremental, well-scoped change.** The PR touches exactly the files it needs to, deletes the superseded local functions, and does not introduce unrelated changes.

### Findings

See inline comments below, plus one file-level finding:

**Medium — Reuse / Consistency**: `src/app/onboarding/OnboardingWizard.tsx` (line 460) still defines its own `escapeCsv()` that lacks formula neutralisation. This function builds CSV that is POSTed to `/api/teachers/import` (not downloaded and opened in Excel), so it is not the same CWE-1236 attack surface. However, it creates a maintenance risk: if this CSV ever becomes downloadable, the old escaper silently re-introduces the vulnerability. Consider importing `escapeCsvCell` from `@/lib/csv` there too, or adding a comment (`// #420 — client-side only, not user-downloadable`) so the next person knows the omission is intentional. Not blocking — the current code path is safe.

---

src/app/api/admin/leads/export/route.ts
```diff
@@ -52,15 +46,17 @@ export async function GET(request: NextRequest) {
...
-    escapeCsv(lead.school),
-    escapeCsv(lead.workEmail),
-    escapeCsv(TOOL_LABELS[lead.currentTool] ?? lead.currentTool),
-    escapeCsv(lead.currentToolOther ?? ""),
-    lead.status,
-    escapeCsv(lead.internalNotes ?? ""),
+    escapeCsvCell(lead.name),
+    escapeCsvCell(lead.designation),
+    escapeCsvCell(lead.school),
+    escapeCsvCell(lead.workEmail),
```
guangshinhaha, line 55: **Low — Consistency**: `lead.createdAt.toISOString()` is the only cell value in this route that bypasses `escapeCsvCell`. ISO timestamps are machine-generated and can never contain formula triggers or structural CSV characters, so this is safe — but it breaks the "every cell goes through the shared escaper" invariant stated in the comment on line 48.

For consistency (and to avoid a future reviewer wondering "why is this one different?"):

```ts
    escapeCsvCell(lead.internalNotes ?? ""),
    escapeCsvCell(lead.createdAt.toISOString()),  // ← wrap for consistency
  ]);
```

Not blocking — purely cosmetic.

---

Issue #412, opened 2026-06-29
fix(`assignments`): write audit-log entries for relief assign/cancel/decline


guangshinhaha:
## User story

As a security reviewer, I want relief assignment, cancellation, and decline actions written to the append-only `AuditLog`, so that a false or malicious relief activation can be attributed to an actor and reconstructed after the fact.

## Background

Auth and admin actions (OTP request/verify, login, logout, impersonation, user create/role change, feature-flag and system-config changes) are consistently written to `AuditLog` via `writeAuditLog()` (`src/lib/auditLog.ts`). The relief activation surface is the gap: `src/app/api/relief-assignments/route.ts`, `src/app/api/relief-assignments/[id]/route.ts`, and `src/app/api/auto-assign/route.ts` write zero `AuditLog` entries. The `ReliefAssignment` row carries `assignedByUserId`, `cancelledAt`, and `declinedAt` columns, but there is no append-only security event log for create/cancel/decline. This was surfaced by the Threat & Risk Assessment (false-relief-activation risk): without an audit trail, a disputed or abusive assignment cannot be attributed.

### Relief assignment is recorded in the audit log

- **Given** an authenticated relief coordinator with `ASSIGN_INTERNAL_RELIEF`
- **When** they create a relief assignment
- **Then** an `AuditLog` row is written with the actor email, user id, school id, the action (e.g. `relief.assign`), outcome `SUCCESS`, and metadata identifying the assignment, the covered absence, and the relief teacher

### Cancel and decline are recorded in the audit log

- **Given** an existing relief assignment in the coordinator's school
- **When** the assignment is cancelled or declined
- **Then** an `AuditLog` row is written with the corresponding action (`relief.cancel` / `relief.decline`), outcome `SUCCESS`, and metadata identifying the assignment

### Audit-write failure does not break the activation flow

- **Given** the audit log write fails
- **When** a coordinator creates or cancels an assignment
- **Then** the assignment operation still succeeds and the failure is logged to stderr (consistent with the existing best-effort `writeAuditLog` contract)

## Out of scope

- Auto-assign per-row audit granularity (one summary audit entry for an auto-assign run is acceptable; per-assignment rows are not required here).
- Building a UI to view the audit log (admin audit-log viewer already exists separately).
- Alerting on `FAILURE` audit outcomes (tracked as a separate detection gap).

## Design assets

N/A

---

## Technical context

_Pending grooming._

## Data model

_Pending grooming._

## API contract

[...]

PR #435, opened 2026-07-02
fix(assignments): rate-limit relief assignment and auto-assign routes (#413)

guangshinhaha:
Closes #413

**Stack position: 10/12** — base `fix/csv-formula-injection` (PR #434). Merge after #434.

## Summary

- Wires the existing fixed-window limiter (`checkRateLimit`, Redis-backed when `REDIS_URL` is set) into the relief-activation routes, mirroring the `ABSENCE_REPORT_LIMITS` shape and the established 429 contract (`RATE_LIMITED` + `Retry-After`).
- Proposed defaults (per the ticket's "propose sane defaults during grooming"):
  - `RELIEF_ASSIGNMENT_LIMITS.perUser = 60/hour` — **one shared bucket** across `POST /api/relief-assignments` and `POST /api/[slug]/preferred-relief-assignments` (the `from-preferred` legacy wrapper delegates to the latter, so it's covered), sized so hand-filling a full bad day (~40–50 slots) never trips it.
  - `AUTO_ASSIGN_LIMITS.perUser = 30/hour` — one bucket for both preview and confirm, bounding fan-out while leaving room for preview-tweak-preview iteration.
- Checks run immediately after auth, before body parsing/DB work.
- New shared `rateLimitedResponse()` in `dto/respond.ts` so the 429 shape can't drift between routes.

## Test plan

- `tests/api/assignment-rate-limits.test.ts`: budget exhaustion → 429 with `Retry-After` on both routes; preferred path throttled by the shared manual bucket; separate users have separate budgets.
- Affected suites (relief-assignments, auto-assign, preferred-relief-assignments): **54/54 pass** — normal-usage tests unaffected since each test uses a fresh user/bucket.
- `npm test`: 744 passed. `npm run build`: clean 88/88.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #436, opened 2026-07-02
fix(assignments): write audit-log entries for relief assign/cancel/decline (#412)

guangshinhaha:
Closes #412

**Stack position: 11/12** — base `fix/assignment-rate-limits` (PR #435). Merge after #435.

## Summary

- `relief.assign` — written on `POST /api/relief-assignments` success with actor email/id, school, and `{assignmentId, absenceReportId, timetableEntryId, reliefTeacherId, date}` metadata.
- `relief.cancel` — written on `DELETE /api/relief-assignments/[id]`. The route hard-deletes the row, so the audit entry is deliberately written from the pre-delete snapshot — it's the only surviving record of who cancelled what.
- `relief.decline` — written in the teacher acknowledge route, **once per transition** (mirrors the route's existing once-per-response OutboundMessage discipline, so refresh/double-tap doesn't accumulate rows).
- `relief.auto_assign` — one summary row per confirmed run (`date`, `createdCount`, `assignmentIds`), per the ticket's "per-row granularity not required".
- Failure tolerance: all writes use the existing `writeAuditLog`, whose never-throws/stderr contract is already covered by `src/__tests__/auditLog.test.ts` ("swallows errors without throwing") — the activation flow cannot be broken by an audit failure.

## Test plan

- `tests/api/relief-audit.test.ts`: all four actions asserted with actor/school/metadata; decline idempotence (re-POST → still one row).
- relief-assignments + delete + auto-assign suites: **40/40 pass**.
- `npm test`: 744 passed. `npm run build`: clean 88/88.


src/app/api/relief-assignments/[id]/route.ts
```diff
@@ -19,6 +20,23 @@ export async function DELETE(
...
+      userId: session.user.id,
+      schoolId,
+      metadata: {
+        assignmentId: assignment.id,
+        absenceReportId: assignment.absenceReportId,
+        timetableEntryId: assignment.timetableEntryId,
+        reliefTeacherId: assignment.reliefTeacherId,
+        date: assignment.date.toISOString().slice(0, 10),
+      },
+    });
```
Comment (guangshinhaha) [inline src/app/api/relief-assignments/[id]/route.ts:38]: **[Medium] Audit log for cancel can be silently lost — the hard-delete and audit write are not atomic**

```typescript
    await prisma.reliefAssignment.delete({ where: { id } });

    // #412 — the row is hard-deleted, so this audit entry is the only
    // remaining record of who cancelled what.
    await writeAuditLog({  // ←
      action: "relief.cancel",
      ...
    });
```

**Problem:** The comment correctly notes that the audit entry is "the only remaining record" after the hard-delete. However, the delete and audit write are two independent DB operations. If the serverless function is interrupted between them (cold-start reclaim, deploy swap, or an unhandled exception in `writeAuditLog`'s Prisma call before the try/catch fires), the assignment row is gone with zero trace in the audit log. The other three audit sites (`relief.assign`, `relief.decline`, `relief.auto_assign`) do not have this problem because the source row survives independently.

**Suggestion:** Move the audit write *before* the delete (the assignment data is already captured), or wrap both in a `$transaction`. The simplest option:

```typescript
    // Write the audit record while the row still exists, so a crash
    // between audit-write and delete leaves us with an extra audit
    // entry (harmless) rather than a missing one (data loss).
    await writeAuditLog({
      action: "relief.cancel",
      outcome: "SUCCESS",
      actorEmail: session.user.email,
      userId: session.user.id,
      schoolId,
      metadata: { ... },
    });

    await prisma.reliefAssignment.delete({ where: { id } });
```

The crash-window is small and may never manifest in practice, so this is a "should fix at convenience" rather than a blocker.

---

guangshinhaha:
## Code Review Summary

| Severity | Count |
|----------|-------|
| Critical | 0 |
| High | 0 |
| Medium | 1 |
| Low | 0 |

### Findings

| 1 | Medium | `src/app/api/relief-assignments/[id]/route.ts` | Hard-delete and audit write are not atomic — a crash between the two loses the only record of the cancellation. Simple fix: swap the order (audit before delete). |

### What Looks Good

- **Consistent audit pattern across all four surfaces** — assign, cancel, decline, and auto-assign all use the same `writeAuditLog` shape with actor, school, and metadata. Easy to query and correlate.
- **Idempotency guard on decline** — the `!wasDeclined` check mirrors the existing `logTeacherResponse` pattern, so double-tap / refresh does not accumulate duplicate audit rows. The test explicitly verifies this with a second POST.
- **Test coverage is thorough** — all four audit actions have dedicated integration tests with realistic setup (school, admin, teachers, periods, assignments). The decline test specifically validates the double-tap scenario.
- **`writeAuditLog` contract is well-documented** — callers correctly treat it as best-effort (never throws), and comments reference the issue number for traceability.

---

PR #437, opened 2026-07-02
feat(security): centralize multi-tenant schoolId scoping with a tenant-scoped Prisma client (#415)

### LINKED ISSUE #415: feat(`security`): centralize multi-tenant schoolId scoping (RLS-lite)
## User story

As a security reviewer, I want multi-tenant `schoolId` scoping enforced in one central place rather than re-implemented in every route, so that a forgotten filter in a single query cannot leak one school's data to another.

## Background

Multi-tenancy is currently enforced purely in application code: most routes use the correct `findFirst({ where: { id, schoolId } })` pattern, and `requireSchoolForSlug` binds the URL slug to the session school. There is no database-level Row-Level Security (RLS) backstop (no `CREATE POLICY` / `ENABLE ROW LEVEL SECURITY` anywhere). The systemic risk flagged by the Threat & Risk Assessment is that isolation depends on every route author remembering to scope by `schoolId`; one missed filter is a cross-tenant leak with no safety net.

True Postgres RLS is a larger architecture change that collides with two constraints in this stack: (1) the interim Supabase connection role has `BYPASSRLS` and PgBouncer transaction-pooling requires transaction-scoped `SET LOCAL` GUCs, and (2) the project rule forbids hand-written migration SQL, which RLS policies require. This ticket therefore targets the near-term 80/20: a centralized application-layer scoping guard (a Prisma client extension that injects the tenant filter automatically). True DB-level RLS is captured as a follow-up to revisit on RDS (where roles and pooling are controlled directly).

### A query without an explicit school scope is automatically scoped

- **Given** a request executing in an authenticated school context
- **When** application code queries a tenant-scoped model without manually adding `schoolId`
- **Then** the query is automatically constrained to the current school and cannot return another school's rows

### Cross-tenant access is denied even when a route omits the filter

- **Given** a route that queries a tenant-scoped model by id but forgets to add `schoolId`
- **When** the id belongs to a different school
- **Then** no row is returned (the centralized guard prevents the cross-tenant read)

## Out of scope

- True database-level Postgres RLS policies (captured as a separate follow-up for the RDS migration; not implemented here).
- Raw SQL paths (`$queryRaw`): the application-layer guard does not cover these and they must continue to scope manually.
- Field-level or role-level over-exposure within a single school (a distinct concern).

## Design assets

```mermaid
sequenceDiagram
    participant Route as API route
    participant Ext as Prisma client extension
    participant DB as Postgres
    Route->>Ext: prisma.teacher.findMany() (no schoolId)
    Ext->>Ext: read current schoolId from request context
    Ext->>DB: SELECT ... WHERE schoolId = :current
    DB-->>Route: only current school's rows
```

---

## Technical context

_Pending grooming._

## Data model

_Pending grooming._

## API contract

_Pending grooming._

## Error contract

_Pending grooming._

## Additional test scenarios

_Pending grooming._

## Hard constraints

_Pending grooming._

---

guangshinhaha:
Closes #415

**Stack position: 12/12** — base `fix/relief-audit-logging` (PR #436). Merge last.

## Summary

- `tenantDb(schoolId)` (`src/lib/tenantDb.ts`): a Prisma client extension, cached per school, that
  - AND-merges `{ schoolId }` into `where` for `findMany` / `findFirst(-OrThrow)` / `count` / `aggregate` / `groupBy` / `updateMany` / `deleteMany` — an explicit foreign `schoolId` from the caller can never widen scope;
  - force-fills `schoolId` on `create` / `createMany` / `upsert` and **throws** on an explicit foreign value;
  - passes non-tenant models through untouched.
- All `requireSchool*` guards now return it as `db`, so route adoption is a mechanical `prisma` → `db` swap (follow-up sweeps can convert routes incrementally; new routes get it for free from the guard).
- **Coverage pin:** a test parses `schema.prisma` and fails if any model with a required `schoolId` is missing from `TENANT_MODELS` (User / MatchingConfig / AuditLog are documented nullable exclusions) — a new tenant model can't silently ship unscoped.

## Design deviations from the issue sketch (and why)

- **No implicit request context.** The sketch reads schoolId "from request context", but `AsyncLocalStorage.enterWith()` inside an awaited guard does not propagate back to the route handler's continuation — verified empirically with a Node repro (store is `undefined` after `await guard()`). `als.run()` would require wrapping every handler: same migration cost as `db`, with more magic. The explicit scoped client is the honest 80/20.
- **Unique-where ops (`findUnique`/`update`/`delete`) not rewritten** — Prisma's unique input can't take extra filters; routes keep the established `findFirst({ id, schoolId })` pre-check pattern for those.
- **Raw SQL** stays manually scoped (issue's own out-of-scope).

## Test plan

- `tests/api/tenant-db.test.ts`: **10/10** — filter-less `findMany` scoped; foreign-id `findFirst` → null (the forgotten-filter AC); foreign `schoolId` in where can't widen; scoped `count`/`updateMany`/`deleteMany`; `create` auto-fill + foreign-value throw; relation-bearing model scoped; non-tenant passthrough; client caching; schema coverage pin.
- **Full-stack verification for the whole 12-PR stack:** complete `npm run test:api` — the 32 failures are **byte-identical to the `main` baseline** (diff clean); the stack adds 68 tests, all passing. `npm test`: 744 passed. `npm run build`: clean 88/88.


guangshinhaha:
Review finding: `tenantDb` still allows write-scope bypasses. Relation-style `school: { connect: ... }` writes skip `schoolId` enforcement, and `upsert` only forces the create data while leaving `where`/`update` unscoped. I am patching this on the stack.

src/lib/tenantDb.ts
```diff
@@ -0,0 +1,150 @@
...
+      $allModels: {
+        async $allOperations({ model, operation, args, query }) {
+          if (!TENANT_MODELS.has(model)) return query(args);
+
+          if (WHERE_SCOPED_OPS.has(operation)) {
+            scopeWhere(args as WhereArgs, schoolId);
+            return query(args);
+          }
+
+          if (CREATE_OPS.has(operation)) {
```
guangshinhaha, line 124: **[High] `upsert` where-clause is unscoped and `update` sub-arg is unguarded -- cross-tenant update is possible**

```typescript
if (CREATE_OPS.has(operation)) {
  const a = args as DataArgs;
  if (Array.isArray(a.data)) {
    for (const row of a.data) forceSchoolId(row, schoolId, model);
  } else if (a.data) {
    forceSchoolId(a.data, schoolId, model);
  }
  if (a.create) forceSchoolId(a.create, schoolId, model); // ←  only `create` checked
  return query(args);                                       // ←  `update` sub-arg ignored
}
```

**Problem:** `upsert` is in `CREATE_OPS`, so the interceptor actively handles it -- but only partially. A Prisma `upsert` has three sub-args: `where` (unique constraint, correctly left alone), `create` (guarded by `forceSchoolId`), and `update` (unguarded). If a route does `db.teacher.upsert({ where: { id: someId }, update: { schoolId: foreignId }, create: {...} })`, the `update` path can (a) match a row belonging to another school via the unscoped unique `where`, and (b) reassign it via the unguarded `update` data. The `DataArgs` type omits `update`, so the field is never inspected.

Since the doc comment says unique-where ops are intentionally not rewritten, consider moving `upsert` out of `CREATE_OPS` entirely (into the pass-through bucket alongside `findUnique`/`update`/`delete`), or handle it explicitly:

**Suggestion:**
```typescript
type DataArgs = {
  data?: Record<string, unknown> | Record<string, unknown>[];
  create?: Record<string, unknown>;
  update?: Record<string, unknown>;  // add this
};

// In the CREATE_OPS branch, after the existing create/data checks:
if (operation === "upsert") {
  const a = args as DataArgs;
  if (a.create) forceSchoolId(a.create, schoolId, model);
  if (a.update && a.update.schoolId !== undefined && a.update.schoolId !== schoolId) {
    throw new Error(
      `tenantDb(${schoolId}): refusing to upsert-update ${model} row to foreign school ${String(a.update.schoolId)}`,
    );
  }
  return query(args);
}
```

Alternatively, if upserts on tenant models should always go through the `findFirst + scoped update` pattern (as the doc comment implies for unique-where ops), remove `upsert` from `CREATE_OPS` and add a runtime throw to force callers to the safe pattern:

```typescript
if (operation === "upsert" && TENANT_MODELS.has(model)) {
  throw new Error(
    `tenantDb: upsert on tenant model ${model} is disallowed — use findFirst({schoolId}) + create/update instead`,
  );
}
```

---

src/lib/tenantDb.ts
```diff
@@ -0,0 +1,150 @@
...
+  model: string,
+): void {
+  if (row.schoolId !== undefined && row.schoolId !== schoolId) {
+    throw new Error(
+      `tenantDb(${schoolId}): refusing to write ${model} row for foreign school ${String(row.schoolId)}`,
+    );
+  }
+  // Leave relation-style writes (`school: { connect: ... }`) alone — setting
+  // the scalar FK alongside them is a Prisma validation error.
+  if (row.school === undefined) {
```
guangshinhaha: **[Medium] `forceSchoolId` can be bypassed via relation-style `school: { connect: { id: foreignId } }`**

```typescript
function forceSchoolId(
  row: Record<string, unknown>,
  schoolId: string,
  model: string,
): void {
  if (row.schoolId !== undefined && row.schoolId !== schoolId) {
    throw new Error(/* ... */);
  }
  // Leave relation-style writes (`school: { connect: ... }`) alone — setting
  // the scalar FK alongside them is a Prisma validation error.
  if (row.school === undefined) {   // ←
    row.schoolId = schoolId;
  }
}
```

**Problem:** When a caller writes `db.teacher.create({ data: { school: { connect: { id: foreignSchoolId } } } })`, `row.school` is defined (the connect object), so `row.schoolId` is never set. And `row.schoolId` is `undefined`, so the first check passes silently. Prisma resolves the FK from the `connect`, creating a row belonging to the foreign school. No current app code uses this pattern, but the guard creates a false sense of safety — a future contributor using the relation style would silently bypass tenant scoping.

**Suggestion:**
```typescript
function forceSchoolId(
  row: Record<string, unknown>,
  schoolId: string,
  model: string,
): void {
  if (row.schoolId !== undefined && row.schoolId !== schoolId) {
    throw new Error(
      `tenantDb(${schoolId}): refusing to write ${model} row for foreign school ${String(row.schoolId)}`,
    );
  }
  if (row.school !== undefined) {
    // Relation-style write detected — we cannot safely inject the scalar FK
    // alongside it (Prisma validation error), but we also cannot verify the
    // connect target matches. Throw to force the caller to use the scalar FK.
    throw new Error(
      `tenantDb(${schoolId}): use schoolId (scalar FK) instead of school (relation) in ${model} creates — tenant scoping cannot verify relation connects`,
    );
  }
  row.schoolId = schoolId;
}
```

---

guangshinhaha:
## Review Summary

| Severity | Count |
|----------|-------|
| High | 1 |
| Medium | 1 |

### Findings

| 1 | High | `src/lib/tenantDb.ts:118` | `upsert` in `CREATE_OPS` only guards `create` sub-arg; `update` sub-arg and unscoped unique `where` allow cross-tenant updates |
| 2 | Medium | `src/lib/tenantDb.ts:100` | `forceSchoolId` skips enforcement when `school` relation-style write is used, bypassing tenant scoping |

### What Looks Good

- **TENANT_MODELS coverage pin test** is an excellent design: the schema-parsing test prevents new tenant models from silently shipping without scoping. This catches the most common class of regression at PR time.
- **Explicit scoped client over AsyncLocalStorage** is a pragmatic choice, well-documented with empirical findings. The `requireSchool*` guards returning `db` makes adoption a mechanical `prisma` -> `db` swap.
- **AND-merge strategy in `scopeWhere`** correctly prevents callers from widening scope with an explicit foreign `schoolId` -- the `AND` wrapping makes the tenant filter irremovable.
- **Test coverage** for the core read/write/create paths is thorough, including the cross-tenant `findFirst` by foreign ID and the explicit-foreign-schoolId-in-where case.

---

Issue #443, opened 2026-07-03
security: POST /api/auth/request-otp has no rate limiting — OTP flooding / email bombing / quota exhaustion


guangshinhaha:
## Summary
`POST /api/auth/request-otp` has **no rate limiting** (neither per-email nor per-IP), and there is no middleware throttle. Every request triggers a live OTPaaS send. This enables OTP flooding / email bombing of any known user and unbounded consumption of OTPaaS send quota.

## Location
`src/app/api/auth/request-otp/route.ts`

The handler never calls `checkRateLimit` (`src/lib/rateLimit.ts` is wired only into `absence-reports` and `interest`). `src/middleware.ts` has no rate-limit logic. `extractIp` / `extractUserAgent` are captured for audit but not used to throttle. Every request reaches OTPaaS `/otp` send (or the local email send).

## Failure scenario
An attacker scripts thousands of POSTs to `/api/auth/request-otp` with a victim's email → OTPaaS sends a flood of OTP emails (email bombing) and burns the OTPaaS send quota, which also breaks legitimate logins. A wordlist of emails can be swept similarly.

## Fix direction
Apply the existing `checkRateLimit` helper keyed on both email and IP (as `interest` already does — 5/hour), with a short per-email cooldown between OTP sends. Return a uniform response regardless of whether the email is known.

## Related
Not the same as #293 (OTPaaS allowlist). This is about missing throttling on the send endpoint.

---
*Filed from the 2026-07 full-app audit. Verified against code.*

guangshinhaha:
Fixed on `main` — verified against current `origin/main` (e5350ed). Landed in **#491** (`8c2df60`).

`src/app/api/auth/request-otp/route.ts` now rate-limits before any send: per-email + per-IP buckets (`OTP_REQUEST_LIMIT = { max: 5, windowSec: 3600 }`) via `checkRateLimit`, returning **429** with a `Retry-After` header and a uniform body (no account-existence oracle). Closing as completed.

PR #491, opened 2026-07-03
fix(security): close OAuth backslash open-redirect (#452) and rate-limit OTP sends (#443)

### LINKED ISSUE #452: security: Google (MOE iCON) OAuth post-login redirect allows backslash open-redirect (missing /\ guard the expire route has)
## Summary
The Google (MOE iCON) OAuth post-login redirect accepts a `next` path that starts with `/\`, which browsers normalize to `//` — a protocol-relative open redirect. The `/api/auth/expire` route already guards against exactly this; the OAuth flow does not.

## Locations
- `src/app/api/auth/google/callback/route.ts:145-146`
- `src/app/api/auth/google/route.ts:28` (set-side, same weaker check)

```ts
if (nextPath && nextPath.startsWith("/") && !nextPath.startsWith("//")) {
  return NextResponse.redirect(`${baseUrl}${nextPath}`);
}
```

Contrast `src/app/api/auth/expire/route.ts:16`, which additionally rejects `/\`:

```ts
next && next.startsWith("/") && !next.startsWith("//") && !next.startsWith("/\\")
```

`/\evil.com` passes `startsWith("/")` and `!startsWith("//")`, and `${baseUrl}${nextPath}` yields `https://app/\evil.com`, which several browsers treat as `https://app//evil.com` → navigation to `evil.com`.

## Failure scenario
An attacker sends a victim `/api/auth/google?next=/\evil.com`. After the victim authenticates, the callback 307-redirects to an attacker-controlled host — a convincing post-login phishing pivot.

## Fix direction
Add `&& !nextPath.startsWith("/\\")` to both the set-side and callback checks (or centralize the safe-relative-path check used by `expire`).

---

guangshinhaha:
## Summary

Two fast-follow security fixes from the 2026-07 audit.

**#452 — OAuth backslash open-redirect.** The Google (MOE iCON) OAuth flow persisted and honoured a post-login `next` path starting with `/\`, which browsers normalise to `//` — a protocol-relative open redirect (e.g. `/api/auth/google?next=/\evil.com` → post-auth 307 to `evil.com`, a convincing phishing pivot). The safe-relative-path guard on **both** the set-side (`src/app/api/auth/google/route.ts`) and the callback (`src/app/api/auth/google/callback/route.ts`) now also rejects `/\`, mirroring the stronger check already present in `src/app/api/auth/expire/route.ts`.

**#443 — unthrottled OTP send.** `POST /api/auth/request-otp` had no rate limiting, so a script could email-bomb a known victim and exhaust the OTPaaS send quota (which also breaks legitimate logins). Added `checkRateLimit` (from `src/lib/rateLimit.ts`, unchanged) keyed on **both email and IP**, 5/hour — the same policy `/api/interest` uses. The throttle sits before the user lookup and returns a **uniform 429** independent of whether the email is a known user, so it cannot be used as an account-existence oracle.

## Why

Both are security defects verified against code in the audit. #452 is a phishing pivot; #443 enables email-bombing and quota exhaustion.

## Test plan

- New pure-logic test `src/__tests__/googleOauthNextGuard.test.ts` codifies the safe-`next` invariant: `/\evil.com`, `//evil.com`, `https://evil.com`, non-slash and empty all rejected; `/dashboard` allowed.
- Extended `tests/api/google-oauth.test.ts`: set-side does **not** persist a `/\evil.com` next cookie (but does persist `/dashboard`); callback ignores a `/\evil.com` next cookie (falls back to role default) and honours a safe relative next.
- New `tests/api/request-otp-rate-limit.test.ts`: 429 after the limit for known and unknown emails; identical throttled response for known vs unknown (no oracle); per-IP bucket caps a single IP sweeping many emails.
- `npx tsc --noEmit` — passes. No `any`, no eslint-disable.
- `npm test` (unit) — 749 passed, 1 skipped.
- `npm run test:api` — **not run**: the Docker daemon is down in this environment, so the integration Postgres (port 5433) is unreachable. The new API tests should be run once the test DB is up / in CI.

Closes #452 #443


src/__tests__/googleOauthNextGuard.test.ts
```diff
@@ -0,0 +1,50 @@
...
+ * `src/app/api/auth/google/callback/route.ts` (redirect-side) and MUST match
+ * the stronger check in `src/app/api/auth/expire/route.ts`, which also rejects
+ * a backslash-prefixed path (`/\evil.com` normalises to `//evil.com` in
+ * browsers — a protocol-relative open redirect).
+ *
+ * This is a pure-logic mirror of that guard so the invariant is asserted
+ * without spinning up the route. If the guard in the routes ever diverges,
+ * these cases document exactly what "safe relative path" must mean.
+ */
+function isSafeRelativeNext(next: string | null | undefined): boolean {
```
guangshinhaha: **[Medium] Redirect guard is duplicated inline in 4 places — `login/page.tsx` was left on the weak version**

This PR hardens the `next` guard in `google/route.ts`, `google/callback/route.ts`, and (already) `expire/route.ts`, and mirrors it here in the test. But the same guard also lives in **`src/app/login/page.tsx:46`**, which still uses the pre-fix version:

```ts
// src/app/login/page.tsx:46 — NOT updated by this PR
if (next && next.startsWith("/") && !next.startsWith("//")) {   // ← missing !startsWith("/\\")
  setNextPath(next);
}
```

**Problem:** `/login?next=/\evil.com` passes this client guard and lands in `nextPath`. It is currently backstopped because `nextPath` is only forwarded to `/api/auth/google?next=...` (line 163), which this PR now re-validates server-side — so it is not directly exploitable today. But it is exactly the divergence this test's own docstring warns about ("If the guard in the routes ever diverges…"), and it already diverged.

**Suggestion:** Per CLAUDE.md ("shared logic must be declared once in `src/lib/` and imported — never redeclared"), extract the guard to one helper and import it in all four call sites + this test, instead of maintaining five copies:

```ts
// src/lib/safeRedirect.ts
export function isSafeRelativeNext(next: string | null | undefined): boolean {
  return Boolean(
    next && next.startsWith("/") && !next.startsWith("//") && !next.startsWith("/\\")
  );
}
```

Then `login/page.tsx` gets the backslash fix for free and can never drift again.

---

src/app/api/auth/request-otp/route.ts
```diff
@@ -25,6 +33,51 @@ export async function POST(request: Request) {
...
+    // user, so it cannot be used as an account-existence oracle.
+    const ip = extractIp(request);
+    const [emailDecision, ipDecision] = await Promise.all([
+      checkRateLimit({
+        key: `request-otp:email:${normalizedEmail}`,
+        max: OTP_REQUEST_LIMIT.max,
+        windowSec: OTP_REQUEST_LIMIT.windowSec,
+      }),
+      checkRateLimit({
+        key: `request-otp:ip:${ip ?? "unknown"}`,
```
guangshinhaha: **[Low] `ip ?? "unknown"` collapses all header-less requests into one shared throttle bucket**

```ts
checkRateLimit({
  key: `request-otp:ip:${ip ?? "unknown"}`,   // ← every null-IP request shares ONE bucket
  max: OTP_REQUEST_LIMIT.max,      // 5
  windowSec: OTP_REQUEST_LIMIT.windowSec,
}),
```

**Problem:** `extractIp()` returns `null` when neither `x-forwarded-for` nor `x-real-ip` is present. Every such request keys to `request-otp:ip:unknown`, so if a deployment ever serves this route without a proxy setting those headers, *all* users combined get just 5 OTP sends/hour and everyone else is locked out with a 429. On Vercel and the eventual GCC ALB `x-forwarded-for` is always populated, so this is not reachable in the current targets — flagging as latent/robustness only.

**Suggestion:** When IP is unknown, skip the IP bucket (rely on the per-email bucket) rather than funnelling everyone into a single shared counter:

```ts
const ip = extractIp(request);
const checks = [checkRateLimit({ key: `request-otp:email:${normalizedEmail}`, ...OTP_REQUEST_LIMIT })];
if (ip) checks.push(checkRateLimit({ key: `request-otp:ip:${ip}`, ...OTP_REQUEST_LIMIT }));
const decisions = await Promise.all(checks);
```

---

guangshinhaha:
## Code Review Summary

| Severity | Count |
|----------|-------|
| Critical | 0 |
| High     | 0 |
| Medium   | 1 |
| Low      | 1 |

The two security fixes themselves are correct: the `/\` backslash check closes the protocol-relative open redirect (browsers normalise `/\evil.com` → `//evil.com`), and it is applied consistently across `google/route.ts`, `google/callback/route.ts`, and `expire/route.ts`. The OTP throttle is enumeration-safe (identical 429 body for known vs unknown email, checked before the user-existence lookup) and has no off-by-one (`count <= max` → 5 allowed, 6th blocked). The two findings are a duplication/consistency gap (the sibling guard in `login/page.tsx` was left on the weak version — backstopped, not exploitable) and a latent shared-bucket edge case on null IP.

## What Looks Good
- Rate limit is placed **before** the dev-bypass and user-existence branches, so it cannot be turned into an account-existence oracle — and the tests assert this explicitly.
- Dual per-email **and** per-IP keying with `Math.max` on the two `retryAfterSec` values correctly surfaces the longer cooldown.
- Strong test coverage: pure-logic guard mirror, integration tests for both the cookie-set and callback-redirect sides, and IP-sweep + enumeration-safety cases.
- The new 429 shape (`error`/`code`/`retryAfterSec` + `Retry-After` header) is handled gracefully by the existing `login/page.tsx` client with no contract break.

---

Issue #473, opened 2026-07-03
bug: teachers CSV import is unbounded (no row/byte cap) and runs one interactive transaction over the whole file


guangshinhaha:
## Summary
`POST /api/teachers/import` accepts the CSV as an arbitrary JSON string with **no length/row cap**, and confirm mode wraps every create/update in a single interactive `prisma.$transaction` loop. A very large CSV holds a long transaction (Prisma interactive-transaction timeout → whole batch 500s and rolls back) and can auto-create unlimited `Position` rows.

## Location
`src/app/api/teachers/import/route.ts:14` (input), confirm-mode transaction loop ~:116-219

`parseTeacherCsv` has no row cap. Contrast `import-timetable` and `pool-teachers/import`, which enforce byte + row caps (5 MB / 20k rows).

## Failure scenario
A school admin pastes a 200k-row CSV (or a scripted client posts one): the request pins a DB connection for the transaction's full duration, then times out and rolls back with an opaque 500 — no partial progress, no row-count guidance.

## Fix direction
Add byte + row caps (mirroring `import-timetable`/`pool-teachers/import`), and batch or chunk the writes rather than one giant interactive transaction.

---
*Filed from the 2026-07 full-app audit.*

guangshinhaha:
Fixed on `main` — verified against `origin/main` (e5350ed). Landed in **#492** (`04e81db`).

`teachers/import/route.ts` now enforces `MAX_IMPORT_BYTES = 5MB` (checked against both `content-length` → 413 and actual `Buffer.byteLength` to defeat spoofing) and `MAX_IMPORT_ROWS = 20_000` (→ 400), matching the pool-teachers/import + import-timetable caps. The confirm-mode write remains a single interactive `$transaction` **by design**, but is now bounded up front — which is what this issue asked for. Closing as completed.

Issue #454, opened 2026-07-03
security: POST /api/timetable accepts cross-school teacherId/periodId (cross-tenant reference injection)


guangshinhaha:
## Summary
`POST /api/timetable` writes a `TimetableEntry` using `teacherId` and `periodId` from the request body **without verifying those rows belong to the caller's school**. `TimetableEntry` has independent `schoolId`, `teacherId`, `periodId` columns with no DB-level tenant-consistency constraint, so cross-school references persist.

## Location
`src/app/api/timetable/route.ts:46-81`

```ts
prisma.timetableEntry.create({
  data: { teacherId, /* from body */ ..., periodId, /* from body */ ..., schoolId },
})
```

The `findFirst` at ~line 61 includes `schoolId` but is only the upsert-detection path; when it misses, the create proceeds with unvalidated FKs.

## Failure scenario
A `SCHOOL_ADMIN` of school A posts a `teacherId`/`periodId` belonging to school B. An entry with `schoolId=A` but a school-B teacher/period is created. School A's dashboard/roster joins then render school B's teacher name and period times inside A's tenant, and the entry is invisible/uncleanable from B's side.

## Fix direction
Verify `teacherId` and `periodId` resolve to rows with `schoolId === session schoolId` before create; reject otherwise. (Same class as the teachers `schoolRoleId`/`positionId` and auto-assign cross-tenant issues.)

---
*Filed from the 2026-07 full-app audit. Verified against code.*

Issue #447, opened 2026-07-03
security: teachers create/update accept cross-school schoolRoleId/positionId without ownership check


guangshinhaha:
## Summary
`POST /api/teachers` and `PATCH /api/teachers/[id]` write `schoolRoleId` and `positionId` straight from the request body **without verifying those rows belong to the caller's school**. `SchoolRole` and `Position` are per-school models, but the FK on `Teacher` is bare, so cross-school IDs attach successfully.

## Locations
- `src/app/api/teachers/route.ts:69,72` — `schoolRoleId: schoolRoleId ?? null`, `positionId: positionId || null`
- `src/app/api/teachers/[id]/route.ts` (PATCH) — same pattern

Contrast the import route, which resolves roles/positions by name within `where: { schoolId }`.

## Failure scenario
School A's admin (or a compromised admin session) sets a teacher's `positionId` to school B's `Position`. Auto-assign's weekly-cap logic (`resolveWeeklyCap` reading `position.defaultMaxReliefPerWeek`) and role-priority ranking (`schoolRole.priorityOrder`) then silently use another tenant's config values, and school B's role/position names render inside school A's UI/exports.

## Fix direction
Before writing, verify `schoolRoleId` / `positionId` (when present) resolve to rows with `schoolId === session schoolId`; reject with 400/404 otherwise.

---
*Filed from the 2026-07 full-app audit. Verified against code.*

guangshinhaha:
Fixed on `main` — verified against `origin/main` (e5350ed). Landed in **#492** (`04e81db`).

Both handlers now do school-scoped ownership lookups before writing: POST `teachers/route.ts` and PATCH `teachers/[id]/route.ts` each run `prisma.schoolRole.findFirst({ where: { id, schoolId } })` and `prisma.position.findFirst({ where: { id, schoolId } })`, returning **400** on a cross-school / non-existent id (null-clear still allowed). Both tagged `// #447`. Closing as completed.

Issue #472, opened 2026-07-03
validation: several handlers return 500 instead of 400 for malformed date/JSON/enum input


guangshinhaha:
## Summary
Several handlers return **500** for malformed-but-parseable input where the project's testing rules require a **400** naming the bad field. Malformed dates/bodies reach Prisma (or throw) and surface as "Internal server error" with a logged stack.

## Locations
- `src/app/api/relief-assignments/route.ts:37` — `new Date(dateStr + "T00:00:00.000Z")` with no NaN check → `date: "garbage"` → Prisma 500.
- `src/app/api/export/route.ts:23-24` — same for `from`/`to` query params.
- `src/app/api/user/preferences/route.ts:14` — `await req.json()` outside try/catch → invalid JSON → unhandled 500.
- `src/app/api/teachers/[id]/route.ts:96` — `subjects.map(s => s.trim())` throws on non-string/non-array `subjects` → 500.
- `src/app/api/timetable/route.ts:11` (GET) — `weekType` cast straight into the Prisma where → `?weekType=BOGUS` → invalid enum → 500; POST — `dayOfWeek` not validated as int 1–5.
- `src/app/api/[slug]/timetable-overrides/route.ts:45`, `[id]/route.ts:69`, `[id]/import/route.ts:42` — `await req.json()` with no try/catch → malformed body → raw 500.

## Failure scenario
Any authorized client sending a malformed date/body/enum gets a 500 (and a logged stack) instead of a 400 identifying the bad field. Client-side validation-error handling can't distinguish bad input from a server fault.

## Fix direction
Validate/parse these inputs (Zod, or explicit NaN/enum/type guards) and return `validationErrorResponse` / 400. Several routes already use Zod — extend the same pattern to these.

---
*Filed from the 2026-07 full-app audit. One issue covering the input-validation-500 cluster (same defect class + fix).*

guangshinhaha:
Fixed on `main` — verified against `origin/main` (e5350ed). Landed across **#492** (`04e81db`, teachers/timetable), **#493** (`07d6468`, export dates) and **#502** (`8548e9b`, the rest).

[...]

PR #492, opened 2026-07-03
fix(api): reject cross-tenant FK writes and validate malformed input on timetable/teachers

guangshinhaha:
## Summary

Fixes a cluster of cross-tenant write IDORs and malformed-input 500s on the timetable and teachers API routes.

- **`Closes #454`** — `POST /api/timetable` verifies `teacherId` **and** `periodId` resolve to rows in the caller's school before create (already present from #421; now covered by dedicated cross-tenant tests that assert nothing persists).
- **`Closes #447`** — `POST /api/teachers` and `PATCH /api/teachers/[id]` now verify `schoolRoleId` and `positionId` (when present) belong to the caller's school; cross-school IDs are rejected with **400** and nothing is written. Clearing to `null` stays valid.
- **`Closes #473`** — `POST /api/teachers/import` gains a **5MB** byte cap (413) and **20,000** row cap (400), mirroring `import-timetable` / `pool-teachers/import`, so an oversized CSV is rejected before it pins a DB connection in the interactive transaction.
- Also fixes the **timetable + teachers parts of #472**:
  - `GET`/`POST /api/timetable` — validate `weekType` against the `WeekType` enum and `dayOfWeek` as an integer 1–5 → **400** naming the field (was Prisma 500).
  - `PATCH /api/teachers/[id]` — guard non-array / non-string `subjects` via `validationErrorResponse` → **400** (was `.map()` throw → 500).

## Why

`SchoolRole`, `Position`, `Teacher`, and `Period` carry independent per-school FKs with no DB-level tenant-consistency constraint. Without a server-side ownership check, an admin (or compromised session) of school A could attach school B's role/position/teacher/period — auto-assign's weekly-cap and role-priority logic would then silently read another tenant's config, and B's names/period metadata would render inside A's UI and exports. The malformed-input cases returned opaque 500s (with logged stacks) where the project's rules require a field-named 400, so clients cannot distinguish bad input from a server fault. The unbounded import could time out and roll back the whole batch with no partial progress.

## Test plan

- `tests/api/timetable.test.ts` — invalid `weekType` (GET + POST), out-of-range / non-integer `dayOfWeek`, valid enum happy paths.
- `tests/api/teachers.test.ts` — cross-tenant `schoolRoleId`/`positionId` rejection on POST and PATCH (asserting no write / teacher unchanged), same-school happy paths, and non-array / non-string `subjects` → 400.
- `tests/api/teachers-import.test.ts` — 413 on >5MB CSV, 400 on >20k rows (nothing written), normal CSV still imports.
- `npx tsc --noEmit` — clean.
- `npm test` (unit suite) — 744 passed / 1 skipped.
- API suite (`npm run test:api`) **not run** — no Postgres on :5433 and `docker` unavailable in this environment; the new integration tests should run in CI.

Edits are scoped to the four target routes plus their test files only.


src/app/api/teachers/route.ts
```diff
@@ -62,6 +62,37 @@ export async function POST(request: NextRequest) {
...
+        return NextResponse.json(
+          { error: "positionId does not belong to this school." },
+          { status: 400 },
+        );
+      }
+    }
+
     const result = await prisma.$transaction(async (tx) => {
       const teacher = await tx.teacher.create({
         data: {
```
guangshinhaha, line 117: **[Medium] `POST /api/teachers` still 500s on the same malformed input this PR fixes in `PATCH`**

```ts
const result = await prisma.$transaction(async (tx) => {
  const teacher = await tx.teacher.create({
    data: {
      name: name.trim(),
      schoolRoleId: schoolRoleId ?? null,   // ← non-string id → Prisma throws
      ...
      subjects: Array.isArray(subjects) ? subjects : [],  // ← ["Math", 42] → Prisma throws
```

**Problem:** This PR converts malformed-input 500s into named 400s in the sibling `PATCH` (subjects is now `z.array(z.string())`-validated), but the same two gaps remain live in `POST`:

- `subjects: ["Math", 42]` — `Array.isArray` passes, the non-string member reaches the `String[]` column and Prisma throws a `PrismaClientValidationError` → opaque 500 (the exact case `PATCH` now returns 400 for).
- A non-string truthy `schoolRoleId`/`positionId` (e.g. `42`) hits the new `#447` `findFirst({ where: { id } })` guard, where `id` is typed `String` → Prisma validation error → 500, instead of the intended 400.

Both are reachable by any direct API caller; the fix pattern already exists a few lines away in the PATCH handler.

**Suggestion:**
```ts
if (subjects !== undefined) {
  const parsed = z.array(z.string()).safeParse(subjects);
  if (!parsed.success) return validationErrorResponse(parsed.error);
}
if (schoolRoleId !== undefined && schoolRoleId !== null && typeof schoolRoleId !== "string") {
  return NextResponse.json({ error: "schoolRoleId must be a string." }, { status: 400 });
}
// same for positionId
```

---

guangshinhaha:
## Code Review Summary

| Severity | Count |
|----------|-------|
| Critical | 0 |
| High     | 0 |
| Medium   | 1 |
| Low      | 0 |

The core intent — tenant-scoped FK-ownership checks and malformed-input validation — is implemented correctly. The FK guards are all scoped by `schoolId` (not global), the `dayOfWeek`/`weekType` enum validation runs before Prisma, and the import byte/row caps land before the interactive transaction. Verified the empty-`weekType` GET/POST behavior change is not reachable from any live caller (the timetable UI always sends `"ODD"`/`"EVEN"`), so it is not a live regression.

## What Looks Good
- FK-existence checks are correctly tenant-scoped (`findFirst({ where: { id, schoolId } })`), closing the cross-tenant write IDOR without leaking existence via enumeration.
- Import size caps are enforced *before* the `$transaction`, with a defense-in-depth `Buffer.byteLength` check that doesn't trust a spoofable/absent `content-length`.
- Validation is placed after the 404/auth checks but before Prisma, so error precedence (401/403 → 404 → 400) stays consistent with the rest of the codebase.
- Thorough test coverage: cross-tenant rejection + "nothing persisted" assertions, non-array/non-string-member subjects, dayOfWeek range/type, and enum validation on both GET and POST.

---

PR #494, opened 2026-07-03
fix(auto-assign): remove dead P2002 catch on save path; add save-integrity tests (#438, #439)

### LINKED ISSUE #438: security: auto-assign save path writes client-supplied assignment IDs without school-ownership checks (cross-tenant IDOR write)
## Summary
`POST /api/auto-assign?preview=false` writes relief assignments from a client-supplied `body.assignments` array **without verifying that the referenced `absenceReportId`, `timetableEntryId`, and `reliefTeacherId` belong to the caller's school**. This is a cross-tenant IDOR write.

## Location
`src/app/api/auto-assign/route.ts:123-144`

```ts
const assignments = body.assignments || result.proposals;
...
prisma.reliefAssignment.create({
  data: {
    absenceReportId: p.absenceReportId,
    timetableEntryId: p.timetableEntryId,
    reliefTeacherId: p.reliefTeacherId,
    date: dateUTC,
    schoolId,                       // ← from session
    assignedByUserId: session.user.id,
  },
})
```

None of the three foreign keys is checked against `schoolId`. Contrast the sibling manual route `src/app/api/relief-assignments/route.ts:39-54`, which explicitly does three `findFirst({ where: { id, schoolId } })` ownership checks with the comment *"Without these checks a caller could reference another school's IDs and create a cross-tenant assignment."* The auto-assign save path was never given the same guard.

## Failure scenario
A `RELIEF_COORDINATOR` or `SCHOOL_ADMIN` of school A who possesses school B IDs (e.g. an admin who moved schools, or IDs leaked via an export) POSTs `/api/auto-assign?preview=false` with `assignments: [{ timetableEntryId: <B entry>, absenceReportId: <B report>, reliefTeacherId: <B teacher> }]`. A `ReliefAssignment` row is created referencing another school's teacher/report/entry, and the fire-and-forget `notifyReliefAssignment` then emails school B's teacher.

## Also note
No per-item shape validation: missing/garbage fields reach Prisma and 500 instead of returning 400.

## Fix direction
Apply the same three `findFirst({ where: { id, schoolId } })` ownership checks used in `relief-assignments/route.ts` to every item of the `assignments` array before creating rows; reject with 404/400 on any mismatch.

---

### LINKED ISSUE #439: bug: auto-assign save path has no conflict checks and its P2002 duplicate guard is dead code (unique index dropped) — silent double-booking
## Summary
The auto-assign confirm/save path has **no conflict checks**, and its only concurrency defense — a `catch` for Prisma `P2002` — is **dead code**, because the unique index it relies on was dropped. Result: the same slot can receive two `ACTIVE` relief assignments (or a teacher can be double-booked in one period) silently, with a `200 { saved: true }`.

## Location
`src/app/api/auto-assign/route.ts:155-173`

```ts
} catch (e) {
  if (e instanceof Prisma.PrismaClientKnownRequestError && e.code === "P2002") {
    return NextResponse.json({ error: "...another admin...", code: "STALE_DASHBOARD" }, { status: 409 });
  }
  throw e;
}
```

The comment says this catches a "partial collision with another KP's concurrent commit." But:

- `prisma/schema.prisma` `model ReliefAssignment` now has only `@@index([timetableEntryId, date])` — **no `@@unique`**.
- The unique index `ReliefAssignment_timetableEntryId_date_key` existed in the baseline migration but was **dropped** in `prisma/migrations/20260612085608_add_timetable_overrides/migration.sql:12` (`DROP INDEX "ReliefAssignment_timetableEntryId_date_key";`) and replaced with a plain index (line 45).

So `P2002` can never fire for this pair, and the save path also performs none of the `teacherBusy` / `slotTaken` pre-checks the manual `relief-assignments/route.ts` route does.

## Failure scenario
Two admins confirm auto-assign against a stale preview (or one admin replays with an `assignments` array covering already-covered slots): both writes succeed, the slot ends up with two `ACTIVE` assignments, two teachers are emailed for the same class, and the dashboard's `assignmentByEntry` map silently shows only one of them.

## Fix direction
Either restore a partial-unique constraint on active assignments per `(timetableEntryId, date)` and keep the catch, or add explicit pre-write conflict checks (as `relief-assignments/route.ts` does) inside the transaction. The current catch is misleading dead code.

---
*Filed from the 2026-07 full-app audit. Verified against code + migration history.*

guangshinhaha:
## Summary

Hardens and finalizes the `POST /api/auto-assign?preview=false` confirm/save path against the two 2026-07 audit findings.

By the time this branch was cut, prior PRs (#414 / #425 / #429) had already added the substantive protections both issues asked for:

- **#438 (cross-tenant IDOR write):** the save path already re-validates every element of the client `assignments` array — each `timetableEntryId`, `reliefTeacherId`, and `absenceReportId` must belong to the caller's session `schoolId` (`findMany({ where: { id: { in: … }, schoolId } })`), rejecting any cross-tenant ID with `400` before any write.
- **#439 (double-booking):** the save transaction already takes advisory locks and runs explicit pre-write conflict checks — `SLOT_TAKEN` (slot already `ACTIVE`) and `TEACHER_BUSY` (teacher covering a concurrent period), both surfaced as `409 { code: "STALE_DASHBOARD" }`.

What remained was the **dead `P2002` catch** that #439 explicitly called out as misleading. This PR removes it.

## Why

`ReliefAssignment` has **no** `@@unique` on `(timetableEntryId, date)` — the unique index `ReliefAssignment_timetableEntryId_date_key` was dropped in migration `20260612085608_add_timetable_overrides` and replaced with a plain index. So a Prisma `P2002` duplicate-key error can never fire on this write. Keeping the catch:

1. implied a DB-level uniqueness guard that does not exist, and
2. would have masked any genuinely unexpected error as a spurious `409 STALE_DASHBOARD`.

Removed the catch and the now-unused `Prisma` import; unexpected errors now propagate to the outer handler as `500`. Added a comment documenting why no `P2002` catch belongs here (the real guard is the locked `SLOT_TAKEN` / `TEACHER_BUSY` pre-checks). No DB constraint or `schema.prisma` change — a partial-unique-on-active-status isn't expressible in Prisma schema and hand-written migration SQL is disallowed by CLAUDE.md.

## Test plan

Added `tests/api/auto-assign-save.test.ts` (integration, `npm run test:api`) covering the save-path integrity contract:

- **#438** — cross-tenant `timetableEntryId`, `reliefTeacherId`, and `absenceReportId` each rejected with `400`, nothing persisted; in-tenant control batch saves.
- **#439** — confirming a slot that already has an `ACTIVE` assignment → `409 STALE_DASHBOARD`, slot keeps exactly one row (the original); two confirms of the same slot don't both persist; a teacher double-booked across a concurrent period in one batch → `409`, whole batch rejected (no partial persist).

- `npx tsc --noEmit` — clean.
- `npm test` (unit suite, also runs on pre-push) — 744 passed / 1 skipped.
- `npm run test:api` — **not run in this environment** (Postgres :5433 / Docker unavailable here). Needs a local run / CI to execute the new integration file. The new tests reuse the same helpers, mocks, and request-builder patterns as the existing `tests/api/auto-assign.test.ts` suite.

Closes #438
Closes #439


tests/api/auto-assign-save.test.ts
```diff
@@ -0,0 +1,302 @@
...
+          {
+            timetableEntryId: entry2.id,
+            absenceReportId: report2.id,
+            reliefTeacherId: reliever.id,
+          },
+        ],
+      }),
+    );
+
+    expect(res.status).toBe(409);
```
guangshinhaha, line 296: **[High] Within-batch TEACHER_BUSY test asserts 409, but the confirm path returns 200 — the guard only checks already-persisted rows, so it cannot detect a same-teacher/same-period conflict inside one batch**

```ts
// route.ts — the teacher-busy loop runs BEFORE any row in this batch is created
for (const a of assignments) {
  const teacherNowBusy = await tx.reliefAssignment.findFirst({   // ← DB only; batch rows not yet written
    where: { schoolId, reliefTeacherId: a.reliefTeacherId, date: dateUTC,
             status: "ACTIVE", declinedAt: null,
             timetableEntry: { periodId: entry.periodId } },
  });
  if (teacherNowBusy) throw new BookingConflictError("TEACHER_BUSY", ...);
}
// ... all creates happen AFTER this loop
```

**Problem:** For this test both assignments target *fresh* entries (`entry`, `entry2`) in the same period with the same reliever. At check time the DB has no ACTIVE rows for that teacher/period, so `teacherNowBusy` is `null` for both; `slot-taken` also passes (distinct `entryIds`, empty table); no constraint violation fires. Both rows are then created and the route returns **200**. `assignmentConstraintViolation` only guards self-cover / reliever-absent, and there is no in-memory tracking of teacher/period pairs already seen in the batch (confirmed in `src/lib/scheduling/assignmentConstraints.ts`). So this test fails: line 296 expects `409` (gets `200`) and line 300 expects `0` rows (gets `2`).

This is not only a test bug — it exposes a real gap: on the client-edited confirm path (#425, TC-DASH-10) a crafted/edited batch can double-book one reliever across two concurrent classes in a single submission, and the server does not reject it. CI did not catch this because the integration suite (`test:api`) is not run by CI on this PR — only the (unrelated) Vercel check ran.

**Suggestion:** Track teacher/period pairs seen earlier in the same batch and reject in-batch collisions, e.g.

```ts
const seen = new Set<string>();
for (const a of assignments) {
  const entry = entryById.get(a.timetableEntryId)!;
  const key = `${a.reliefTeacherId}:${entry.periodId}`;
  if (seen.has(key)) {
    throw new BookingConflictError("TEACHER_BUSY", "A teacher in this batch is assigned to two classes at the same time.");
  }
  seen.add(key);
  // ...existing DB teacherNowBusy check...
}
```

Then re-run `npm run test:api` to confirm the batch case goes green. (Verdict: PLAUSIBLE — traced from source; not executed here because Docker/test-DB is unavailable in this environment.)

---

guangshinhaha:
## Code Review Summary

| Severity | Count |
|----------|-------|
| Critical | 0 |
| High     | 1 |
| Medium   | 0 |
| Low      | 0 |

The core change — removing the dead P2002 catch — is **correct**. Verified that migration `20260612085608_add_timetable_overrides` drops `ReliefAssignment_timetableEntryId_date_key` and replaces it with a plain (non-unique) index, and the `create` call sets no unique field (`replacedById` / `ackToken` stay null), so `P2002` genuinely cannot fire on this write. Removing the unused `Prisma` import is safe (no other usage). One High finding on the new test suite is posted inline.

## What Looks Good
- **Dead-code removal is sound and well-documented** — the NOTE comment accurately captures the advisory-lock + pre-write-check rationale, and the concurrency invariant is upheld by `SLOT_TAKEN` (advisory-locked re-check), not the removed catch.
- **Cross-tenant IDOR tests (#438) are strong** — each asserts both the 400 status and that *nothing* was persisted, and they exercise all three foreign-key surfaces (entry / teacher / report).
- **SLOT_TAKEN double-booking tests are correct** — the pre-existing-ACTIVE-row and sequential double-confirm cases both correctly land on 409 `STALE_DASHBOARD` with a single surviving row.

---

PR #495, opened 2026-07-03
fix(admin): enforce user-mutation integrity (offboarding, cross-school block, awaited audit logs)

### LINKED ISSUE #456: security: clearing a user's schoolId leaves their linked Teacher row, so they retain old-school access
## Summary
Setting a user's `schoolId` to `null` via the admin Users PATCH (intended to remove them from a school) does **not** clear their linked `Teacher` row. Because `requireTeacher` resolves school access from the `Teacher` row (not `user.schoolId`), the user keeps full teacher-surface access to the old school.

## Location
`src/lib/uam.ts:145-151` (called from `src/app/api/admin/users/[id]/route.ts`)

```ts
export async function reconcileTeacherForUser(tx, { userId, role, schoolId, name }) {
  const existing = await tx.teacher.findUnique({ where: { userId } });
  if (existing) {
    if (schoolId && existing.schoolId !== schoolId) {        // ← falsy when schoolId === null
      await tx.teacher.update({ where: { id: existing.id }, data: { schoolId } });
    }
    return;   // schoolId === null ⇒ Teacher row untouched
  }
  ...
}
```

The PATCH sets `data.schoolId = schoolId || null` (route ~line 66). When that is `null`, the guard is falsy, so the `Teacher` row keeps its old `schoolId` and `userId`. `requireTeacher` (`src/lib/auth.ts:216-231`) then still resolves access to the old school.

## Failure scenario
A superadmin "removes a user from their school" by patching `schoolId: null` to revoke access. The user still passes `requireTeacher` and retains access to the old school's teacher surfaces (timetable, absences, acknowledge, dashboard); existing sessions also remain valid.

## Fix direction
When a user's `schoolId` is cleared, unlink or delete the linked `Teacher` row (or explicitly detach `userId`), and revoke active sessions.

---

### LINKED ISSUE #457: bug: moving a user between schools strands their TimetableEntry/AbsenceReport/ReliefAssignment in the old school
## Summary
Moving a user between schools (admin Users PATCH) re-homes only the `Teacher` row's `schoolId`. Its school-scoped child records — `TimetableEntry`, `AbsenceReport`, `ReliefAssignment` — are left pointing at the **old** school, producing a self-inconsistent graph.

## Location
`src/lib/uam.ts:147-149` (with `prisma/schema.prisma` — `TimetableEntry.schoolId`, `AbsenceReport`, `ReliefAssignment` each carry their own `schoolId`)

On a `schoolId` change only `Teacher.schoolId` is updated. `TimetableEntry` rows keep `schoolId = old` and `periodId` referencing old-school `Period`s; `AbsenceReport`/`ReliefAssignment` likewise. Nothing migrates or removes them, and no constraint prevents `Teacher.schoolId != TimetableEntry.schoolId`.

## Failure scenario
User moved from school A to B: A's dashboard/coverage queries (filtered by `TimetableEntry.schoolId = A`) still surface the departed teacher's slots and name (cross-tenant residue), while in B the teacher appears with no timetable. Auto-assign and absence coverage compute against a self-inconsistent graph.

## Fix direction
Decide the intended semantics of a cross-school move (most likely: block it, or archive/detach the teacher's timetable + open assignments) and enforce it, rather than silently re-homing only the `Teacher` row.

## Confidence
Medium — behavior verified; the intended product semantics of a cross-school move aren't documented.

---

### LINKED ISSUE #458: security: impersonation and admin-mutation audit logs are fire-and-forget and can be dropped on serverless freeze
## Summary
Security-sensitive audit-log writes across the admin surfaces are **fire-and-forget** (`writeAuditLog(...)` is not awaited). On serverless (Vercel/GCC), the instance can freeze after the response is returned, dropping the write — exactly the events the audit trail exists to capture. The repo itself documents this hazard and awaits in one place, but not the rest.

## Locations (unawaited `writeAuditLog`)
- `src/app/api/admin/impersonate/route.ts:36,47,65` and `impersonate/stop/route.ts`
- `src/app/api/admin/users/route.ts` (create) and `users/[id]/route.ts` (update — incl. SUPERADMIN grants)
- `src/app/api/admin/schools/route.ts` and `schools/[id]/route.ts` (delete)
- `src/app/api/admin/system-config/route.ts`
- `src/app/api/admin/matching-config/route.ts`

The repo documents the hazard and does the right thing in `admin/feature-flags/route.ts:54-56`:
> "a fire-and-forget write can be dropped when the serverless instance freezes after the response (Vercel/GCC)" — and `await`s there.

Everywhere else the call is not awaited; for impersonation only a `console.log` (in `startImpersonation`) survives.

## Failure scenario
On Vercel/GCC, `admin.impersonation.start`/`stop`, `admin.user.update` (including role escalations to SUPERADMIN), and `admin.school.delete` audit rows are written after the response and lost when the instance freezes — leaving no durable trail of the most sensitive admin actions.

## Fix direction
`await writeAuditLog(...)` in all admin-mutation handlers (as feature-flags already does), or move audit writes into the same transaction as the mutation.

---

guangshinhaha:
## Summary

Fixes three admin-mutation correctness/security issues from the 2026-07 audit. Authoritative scope decision applied: **cross-school user moves are not supported** — offboard + re-add is the path.

- **Closes #456** — Clearing a user's `schoolId` now offboards them properly. `reconcileTeacherForUser` detaches the linked `Teacher` row (`userId = null`, keeping the record and its school-scoped history in the old school), and the Users `PATCH` handler revokes the user's live sessions (`session.deleteMany`). Access resolves via the `Teacher` row, so unlinking is what removes old-school access; session revocation stops an already-logged-in session from lingering. (`Teacher.userId` is nullable in the schema, so no schema change was needed.)
- **Closes #457** — A user with a linked `Teacher` being PATCHed to a **different non-null** school is rejected with **400** (clear message: cross-school moves unsupported; offboard + re-add). Re-homing only `Teacher.schoolId` would strand the teacher's `TimetableEntry`/`AbsenceReport`/`ReliefAssignment` rows in the old school. Same-school PATCH is unchanged.
- **Closes #458** — Every `writeAuditLog(...)` across the admin mutation handlers is now `await`ed, mirroring `feature-flags`: users create/update, schools create/delete, system-config, matching-config (×3), impersonate start (×3 incl. failure paths), impersonate stop. Fire-and-forget writes can be dropped when a serverless instance freezes after the response (Vercel/GCC).

## Why

Because `requireTeacher` resolves school access from the `Teacher` row (not `user.schoolId`), clearing `schoolId` alone left the user with full old-school access and live sessions. Silent cross-school re-homing produced a self-inconsistent tenant graph. And the audit trail — the record of the most sensitive admin actions — was being written after the response and could be lost on freeze.

## Files

Edited only: `src/lib/uam.ts`, the eight admin route handlers in scope, and `tests/api/admin-users-teacher-link.test.ts`. No schema change; no `src/lib/auth.ts` change.

## Test plan

New/updated integration tests in `tests/api/admin-users-teacher-link.test.ts`:
- Clearing `schoolId` unlinks the `Teacher` (record kept, `userId` null, school unchanged) **and** deletes the user's sessions.
- Cross-school move is blocked with 400 and the whole transaction is rolled back (Teacher + user both stay in school A).
- The `admin.user.update` audit row exists the instant the response resolves (proves the write is awaited).
- Updated the prior "keeps school in sync on move" test to assert the new block-with-400 behavior.

- `npx tsc --noEmit` — clean.
- `npm test` (unit suite) — 744 passed, enforced on push.
- `npm run test:api` — **not run locally**: Postgres :5433 / Docker daemon are down in this environment. CI runs the integration suite on push.


src/app/api/admin/users/[id]/route.ts
```diff
@@ -80,19 +85,29 @@ export async function PATCH(
...
         userId: u.id,
         role: u.role,
         schoolId: u.schoolId,
         name: u.name?.trim() || u.email,
       });
 
+      // Offboarding (school cleared): the Teacher row is now detached, so the
+      // user has lost their teacher-surface access. Revoke live sessions too so
+      // an already-logged-in session can't keep using the old school (#456).
+      if (teacherUnlinked) {
```
Comment (guangshinhaha) [inline src/app/api/admin/users/[id]/route.ts]: **[Low] Session revocation is gated on `teacherUnlinked`, so offboarding a user with no linked Teacher row leaves live sessions intact**

```ts
const { teacherUnlinked } = await reconcileTeacherForUser(tx, { ... });

// Offboarding (school cleared): ... Revoke live sessions too ...
if (teacherUnlinked) {                                    // ←
  await tx.session.deleteMany({ where: { userId: u.id } });
}
```

**Problem:** `reconcileTeacherForUser` only returns `teacherUnlinked: true` when an *existing Teacher row* is detached. A user offboarded by clearing `schoolId` who has no Teacher row — e.g. a pure `SCHOOL_ADMIN` — gets `teacherUnlinked: false`, so their live sessions are **not** revoked even though their school was just cleared. Access is largely neutralised because `user.schoolId` is now null (read fresh per request), but this diverges from the PR-stated "offboarding revokes sessions" intent and leaves an unnecessary live session — a defense-in-depth gap if any admin route gates on role without strictly scoping by `schoolId`.

**Suggestion:** revoke sessions whenever the school transitions to null, independent of the teacher-unlink outcome. The route already loaded `existing` (the pre-update user), so:
```ts
const schoolCleared = data.schoolId === null && existing.schoolId !== null;
if (teacherUnlinked || schoolCleared) {
  await tx.session.deleteMany({ where: { userId: u.id } });
}
```

---

guangshinhaha:
## Code Review Summary

| Severity | Count |
|----------|-------|
| Critical | 0 |
| High     | 0 |
| Medium   | 0 |
| Low      | 1 |

## What Looks Good
- **Cross-school block is airtight.** The reject lives inside `reconcileTeacherForUser`, which is called within the `$transaction`, so a blocked move rolls back the user’s `schoolId` update too — the test asserts both the Teacher and the User stay in school A. No path re-homes `Teacher.schoolId` silently.
- **Offboarding is atomic for teachers.** Teacher detach (`userId = null`) and `session.deleteMany` run in the same transaction, so a partial state (detached-but-sessions-live) is impossible. History is preserved in the old school as intended.
- **Awaited audit writes are safe.** `writeAuditLog` swallows its own errors (try/catch → `console.error`), so awaiting flushes the row before responding without risking turning a committed mutation into a 500.
- **Solid test coverage** for all three concerns (cross-school block, offboarding + session revocation, audit durability) plus a clear, actionable user-facing error message for the blocked move.

---

Issue #285, opened 2026-06-11
bug: odd/even week categorisation ignores the school's timetable cycle (ISO-parity only)


guangshinhaha:
## Description
The dashboard / auto-assign decide a date's ODD vs EVEN week purely from ISO week-number parity. They never consult the school's configured `timetableCycle`, so a school whose academic odd/even rotation is offset from ISO parity sees the **wrong week's timetable**, and 10-day-cycle schools aren't handled at all.

## Steps to reproduce
1. Import a TWO_WEEK timetable for a school whose "Week 1 (odd)" does **not** line up with ISO odd weeks (common — depends on term start).
2. Open the dashboard on a date the school considers an odd week but which falls in an even ISO week.
3. Observe the dashboard shows the EVEN-week timetable (and auto-assign fills the wrong periods).

## Expected behaviour
Week type should derive from the school's cycle — e.g. a per-school "week 1 anchor date" for TWO_WEEK, and a day-count-based rule for TEN_DAY — not from ISO parity.

## Actual behaviour
- `getWeekType(date)` returns `getISOWeekNumber(date) % 2` only — [src/lib/weekType.ts:12-14](src/lib/weekType.ts#L12-L14).
- `School.timetableCycle` (WEEKLY/TWO_WEEK/TEN_DAY) is **written** by the import executor but **never read** by any scheduling code (grep: only `timetableImportExecutor.ts` references it).
- Dashboard uses `weekOverride || getWeekType(...)` — [src/app/api/dashboard/route.ts:47](src/app/api/dashboard/route.ts#L47) — and auto-assign receives `weekType` from the client; neither validates it against the school's cycle.
- **TEN_DAY** (schema comment says days 1-5 = ODD / 6-10 = EVEN) is unimplemented.

## Area
Scheduling / dashboard / auto-assign (week categorisation)

## Severity
high — core scheduling correctness; affects which timetable + relief periods are shown for any school whose rotation is offset from ISO parity. A manual ODD/EVEN override exists as a workaround but the default is wrong for these schools.

## Notes
Fix needs: (1) a per-school anchor (which calendar week is the school's "Week 1/odd") for TWO_WEEK; (2) day-count logic for TEN_DAY; (3) compute/validate week type server-side from `school.timetableCycle` rather than trusting a client-supplied param. Found during P2 verification.

guangshinhaha:
**Pilot priority: P0** (`pri:blocking`)

[...]

PR #496, opened 2026-07-03
fix(scheduling): correct week-type, dayOfWeek, 2-opt consecutive, and disjoint-absence handling

### LINKED ISSUE #445: bug: dashboard/available-teachers uses ISO week parity instead of the school's cycle/anchor (#285 fix missed) — wrong-week availability
## Summary
`GET /api/dashboard/available-teachers` derives the week type from **ISO week parity** (`getWeekType`) instead of the school's timetable cycle + Week-1 anchor (`weekTypeForSchool`). The teacher picker therefore shows availability computed against the wrong week's timetable. This is the same bug class fixed for `/api/dashboard` and `/api/auto-assign` under #285 — this route was missed.

## Location
`src/app/api/dashboard/available-teachers/route.ts:66`

```ts
const currentWeekType = weekOverride || getWeekType(selectedDate);   // legacy ISO parity
```

Both `/api/dashboard` (`dashboard/route.ts:61-67`) and `/api/auto-assign` (`auto-assign/route.ts:53-57`) were converted to:

```ts
weekTypeForSchool(date, school.timetableCycle, school.weekOneAnchorDate)
```

The `busy` / `covering` maps and `daySchedule` in the picker are then built from the wrong week's `TimetableEntry` rows.

## Failure scenario
In a `TWO_WEEK` / `TEN_DAY` school whose Week-1 anchor is offset from ISO parity, an admin opens the teacher picker for an uncovered period. Teachers who are actually teaching that period (in the correct week) are listed as free, and the admin assigns one into a clash. The dashboard (correct week) and the picker (ISO week) also disagree, so periods flagged uncovered by the dashboard can resolve against a different timetable in the picker.

## Fix direction
Replace `getWeekType(selectedDate)` with `weekTypeForSchool(...)` using the school's `timetableCycle` and `weekOneAnchorDate`, matching `dashboard`/`auto-assign`.

---

### LINKED ISSUE #478: bug: latestAbsencePerTeacher collapses disjoint same-day partial absences, dropping uncovered periods
## Summary
`latestAbsencePerTeacher` collapses **all** absence reports for a teacher on a date down to the single most-recently-`createdAt` one, with no check that the reports actually cover the same span/periods. Two legitimately-distinct partial-day reports for the same teacher/day (disjoint periods) lose one of them — its periods are never surfaced as uncovered and get no relief.

## Location
`src/lib/scheduling/absenceCoverage.ts:65-76` (used by `autoAssign.ts:128` and daily-summary)

```ts
for (const r of reports) {
  const existing = latest.get(r.teacherId);
  if (!existing || r.createdAt > existing.createdAt) latest.set(r.teacherId, r);
}
```

The doc comment frames dedup as "admin revised an earlier report" but there's no same-span check; auto-assign and the summary build their uncovered-slot set solely from `Array.from(latestByTeacher.values())`.

## Failure scenario
A teacher has two separate `PARTIAL_DAY` reports for the same day covering disjoint periods (e.g. an AM medical appointment filed Monday, then a separate PM commitment filed later). Only the later-created report survives dedup; the earlier report's periods are never surfaced as uncovered — no relief is proposed and they're omitted from the SLT summary.

## Fix direction
Only collapse reports that genuinely supersede each other (same span/periods, or an explicit revision link). Distinct non-overlapping reports for the same teacher/day should both contribute their periods.

## Confidence
Medium — behavior verified; impact depends on whether multiple same-day partial reports are an expected workflow (they appear to be permitted).

---

### LINKED ISSUE #479: bug: auto-assign 2-opt swap can split a deliberately-consecutive relief pairing (prefer_consecutive excluded from Stage 2)
## Summary
During the auto-assign 2-opt improvement pass, `scoreCandidateForSlot` hard-codes `priorPeriodTeacherId: null`, so the `prefer_consecutive` rule scores a constant 1 for every candidate and cannot see the consecutive pairing that Stage 1 built. A 2-opt swap can therefore split a deliberately-consecutive relief pairing with no compensating term. The code comment acknowledges this as a "known limitation … acceptable for v1."

## Location
`src/lib/autoAssign.ts:659-663, 681` (`priorPeriodTeacherId: null` in the Stage-2 scoring)

## Failure scenario
Stage 1 keeps teacher T on both P3 and P4 of class 8A (via `prefer_consecutive`). A 2-opt swap that raises total weighted score by moving T off P4 (e.g. a marginally better subject match elsewhere) is applied, splitting the consecutive pair — undoing the rule's intent.

## Why filed
Self-documented, but it is a real correctness gap between Stage 1 and Stage 2 (the rule the algorithm advertises can be silently undone), not merely a style choice. Filing as low so it's tracked rather than lost in a code comment.

## Fix direction
Thread the actual prior-period teacher into `scoreCandidateForSlot` during 2-opt so `prefer_consecutive` participates, or add a penalty term for breaking a Stage-1 consecutive pair.

---
*Filed from the 2026-07 full-app audit. Verified against code (self-documented limitation).*

### LINKED ISSUE #486: bug: auto-assign derives dayOfWeek from a local-timezone date (off-by-one on non-UTC hosts)
## Summary
`autoAssign` derives the day-of-week from a date parsed in the **server's local timezone**, one line after parsing the same date as UTC. On any host whose timezone is not UTC, `dayOfWeek` can be off by one and the algorithm loads the wrong day's timetable.

## Location
`src/lib/autoAssign.ts:106-108`

```ts
const dateUTC = new Date(dateStr + "T00:00:00.000Z");  // UTC (correct)
const selectedDate = new Date(dateStr + "T00:00:00");   // local — no Z
const dayOfWeek = selectedDate.getDay();                // derived from local
```

`dayOfWeek` then feeds `timetableEntriesForDate({ ... dayOfWeek, weekType })` (line 235), which selects which teachers are scheduled that day.

## Impact
Latent. Singapore has no DST and Vercel functions default to UTC, so today the two parses coincide and the bug is dormant. But the inconsistency is real: if the app is ever run with a non-UTC `TZ` (a local dev machine west of UTC, a mis-set container), `getDay()` returns the previous day for `T00:00:00` local, and auto-assign proposes relief against the wrong day's timetable. Low probability, high confusion when it hits.

## Fix direction
Derive `dayOfWeek` from a UTC-explicit date: use `dateUTC.getUTCDay()`, or parse `selectedDate` with the `...T00:00:00.000Z` suffix like line 106. Add a unit test that pins `dayOfWeek` for a known date independent of `process.env.TZ`.

---
*Filed from the 2026-07 full-app audit (DX/tests/perf pass). Verified against code. Low severity / latent.*

guangshinhaha:
## Summary

Four scheduling/auto-assign correctness fixes from the 2026-07 audit.

| Issue | File | Fix |
|---|---|---|
| #445 | `api/dashboard/available-teachers/route.ts` | Use `weekTypeForSchool(date, school.timetableCycle, school.weekOneAnchorDate)` instead of ISO-parity `getWeekType`, matching `/api/dashboard` and `/api/auto-assign` (#285). Loads the school and honours the `weekType` override. |
| #486 | `lib/autoAssign.ts` | Derive `dayOfWeek` from the UTC-explicit `dateUTC.getUTCDay()`, not a local-timezone `getDay()` — off-by-one on non-UTC hosts. |
| #479 | `lib/autoAssign.ts` | Thread the ACTUAL prior-period teacher into the 2-opt `scoreCandidateForSlot` (was hard-coded `null`) so `prefer_consecutive` participates and a swap that would split a Stage-1 consecutive pairing is charged for it. The hypothetical swapped world is reflected via an override map so adjacent same-class pairs score correctly. |
| #478 | `lib/scheduling/absenceCoverage.ts` | `latestAbsencePerTeacher` now keys by `(teacherId, span signature)` — genuine revisions (same span) still collapse to the latest, but two disjoint same-day PARTIAL_DAY reports both survive and contribute their uncovered periods. Backward compatible: bare `{teacherId, createdAt}` reports keep the legacy per-teacher collapse. |

## Why

Each is a latent or active wrong-answer:
- #445/#486 load the wrong week's / wrong day's timetable, so availability is computed against the wrong schedule.
- #479 lets Stage 2 silently undo the `prefer_consecutive` rule Stage 1 advertised.
- #478 drops the periods of a second same-day partial absence — no relief proposed, omitted from the SLT summary.

## Test plan

New pure-logic tests (`src/__tests__/`, no DB):
- `autoAssignDayOfWeek.test.ts` — `dayOfWeek` pins to the correct weekday and is UTC-stable regardless of `process.env.TZ` (#486).
- `scheduling/absenceCoverage.test.ts` — disjoint same-day partial absences both surface (via `periodRequirements` and legacy `periodIds`); genuine revisions still collapse; spans don't collapse across teachers (#478).
- `autoAssignTwoOpt.test.ts` — `prefer_consecutive` now fires a swap that forms a consecutive pairing, and does NOT break an existing pairing when nothing compensates (#479).

Verification:
- `npm test` — 753 passed / 1 skipped, all green.
- `npx tsc --noEmit` — clean.
- `npm run test:api` — **not run locally** (Docker unavailable in this environment); CI runs it. The existing `dashboard-available-teachers.test.ts` passes an explicit `weekType=ODD` override, so the #445 change is transparent to it, and CI will exercise the school-cycle path.
- `npm run build` — the nested worktree environment has a stray home-dir `node_modules` that shadows resolution and fails the build identically on the clean base commit (unrelated to this diff, which touches no UI/import surface). Vercel builds in a clean tree.

Closes #445
Closes #478
Closes #486
Closes #479


src/lib/scheduling/absenceCoverage.ts
```diff
@@ -51,28 +51,81 @@ interface AbsenceWithCreatedAt {
...
-  const latest = new Map<string, T>();
+export function latestAbsencePerTeacher<
+  T extends AbsenceWithCreatedAt & AbsenceSpanFields,
+>(reports: ReadonlyArray<T>): Map<string, T> {
+  // Keyed by teacherId::spanSignature so disjoint same-day reports co-exist,
+  // while genuine revisions (same span) still collapse to the latest.
+  const latestByKey = new Map<string, T>();
   for (const r of reports) {
-    const existing = latest.get(r.teacherId);
+    const key = `${r.teacherId}::${absenceSpanSignature(r)}`;
```
guangshinhaha, line 122: **[Medium] Keying by exact span signature lets partially-overlapping same-teacher reports both survive, producing duplicate uncovered slots**

```ts
const key = `${r.teacherId}::${absenceSpanSignature(r)}`; // ←
const existing = latestByKey.get(key);
if (!existing || r.createdAt > existing.createdAt) {
  latestByKey.set(key, r);
}
```

**Problem:** The span signature only collapses reports whose period sets are *identical*. Two same-teacher, same-day reports with **overlapping but non-identical** spans get different signatures and both survive. Concretely: report A requires `{p1,p2}` (`req:p1,p2`), report B requires `{p2,p3}` (`req:p2,p3`). Both are kept. Downstream in `autoAssign` (`src/lib/autoAssign.ts` ~L156-170) each surviving report is intersected with the *same* `teacherDayEntries`, so the p2 timetable entry is pushed into `uncoveredSlots` twice. `slotsToFill` (L209) is only filtered against DB-persisted assignments, never deduped by `timetableEntryId`, so Stage 1 emits two proposals for the one p2 slot — a double relief assignment for a single class period (or a unique-constraint failure at persist time). Pre-PR the collapse-by-teacher key made this impossible; the fix trades it for this overlap case. The purely-disjoint case the PR targets is handled correctly.

**Suggestion:** Dedupe uncovered slots by `timetableEntryId` at the consumption site so overlap can never double-emit, regardless of how spans are keyed:

```ts
// in autoAssign, after building uncoveredSlots
const byEntry = new Map<string, UncoveredSlot>();
for (const s of uncoveredSlots) byEntry.set(s.timetableEntryId, s);
const dedupedSlots = [...byEntry.values()];
```

(or union overlapping spans per teacher before keying). Add a test with two overlapping-span reports asserting the shared period yields a single slot.

---

guangshinhaha:
## Code Review Summary

| Severity | Count |
|----------|-------|
| Critical | 0 |
| High     | 0 |
| Medium   | 1 |
| Low      | 0 |

## What Looks Good
- **#486 dayOfWeek fix is correct and well-tested** — switching to `dateUTC.getUTCDay()` makes the timetable-day lookup offset-proof, and the new `autoAssignDayOfWeek.test.ts` pins the TZ-stability invariant across UTC, west-of-UTC, and east-of-UTC zones.
- **2-opt prior-period threading (#479) is sound** — traced the FORMS-a-pairing and DOES-NOT-break-a-pairing cases: the `swapOverride` map correctly reflects the post-swap world for adjacent same-class slots, so `prefer_consecutive` now scores current vs. swapped symmetrically. Both new tests match the traced behaviour.
- **Week-type derivation is now consistent across all three routes** — `available-teachers`, `/api/dashboard`, and `/api/auto-assign` all feed a **local**-parsed date into `weekTypeForSchool` (which uses local date methods), so the picker resolves the same week as the dashboard (#445/#285).
- **Disjoint-span handling has good test coverage** — revision-collapse, cross-teacher isolation, and both `periodRequirements` and legacy `periodIds` paths are covered.

## Note (out-of-diff, not blocking)
`src/app/api/auto-assign/route.ts` (preview branch) still derives `dayOfWeek` via local `selectedDate.getDay()` — the same class of bug #486 fixes in `autoAssign()`. Not touched by this PR; worth a follow-up so preview stats match the assignment path on non-UTC hosts.

---

PR #499, opened 2026-07-03
feat(settings): hide unwired email surfaces behind email_notifications flag

### LINKED ISSUE #483: bug: relief-assignment notifications are fire-and-forget and the outbox is never retried or surfaced (silent notification drops)
## Summary
Relief-assignment email notifications are fire-and-forget and their durable outbox (`OutboundMessage`) is **never drained or retried**. A failed send transitions the outbox row to `FAILED` and is then forgotten — the relief teacher is never notified, there is no retry, and nothing surfaces the failure to a coordinator.

## Location
- `src/lib/notifyAssignment.ts:20` — comment: *"PR c will replace the direct send with a worker that picks up PENDING rows and dispatches them."* That worker was never built.
- `src/lib/notifyAssignment.ts:82-88` — on send failure the row is marked `FAILED` and the outer catch swallows the error (`console.error` only).
- `src/workers/handlers/notification.ts` — the notification queue handler is an explicit **no-op skeleton** (`Received unexpected job … — no-op`); it drains nothing.
- Callers fire-and-forget: `src/app/api/relief-assignments/route.ts:169-173` and `src/app/api/auto-assign/route.ts` both do `import(...).then(...).catch(() => {})`.

## Failure scenario
A coordinator assigns relief; the email provider is briefly down (or the teacher has no `User` row — `notifyAssignment.ts:46-49` returns early and writes no outbox row at all). The dashboard shows the assignment as made, the teacher never hears about it, and no one is alerted. The `FAILED`/absent outbox row is the only trace and nothing reads it.

## Fix direction
Either (a) build the promised notification worker to poll `PENDING`/`FAILED` `OutboundMessage` rows with bounded retry + backoff, or (b) if the direct-send model is intentional, surface undelivered assignments in the dashboard (a badge/list of assignments whose latest outbox row is `FAILED` or missing) so a human can act. Add integration coverage for the failure path (see the critical-path test-coverage issue).

---
*Filed from the 2026-07 full-app audit (DX/tests/perf pass). Verified against code; no notification-outbox drain issue existed previously.*

guangshinhaha:
## Summary

App email delivery is **out of scope this phase**: the notifier (`src/lib/notifier.ts`) no-ops in production — no email provider is wired and `resend` is not a dependency. The **only** real email sent in prod is the OTPaaS login code. So any UI or job that implies working app email delivery is misleading while it sits dormant.

This PR hides those surfaces behind a new **OFF-by-default** feature flag rather than deleting them, so the code is ready the moment a provider is wired. Complements the closed #483 cleanup.

### Changes
- **`src/lib/featureFlags.ts`** — new `email_notifications` flag (default OFF), registered in the same shape as the existing pilot-visibility flags (`rems_directory`, `timetable_exceptions`, …). The seed derives from `FEATURE_FLAGS`, so it is covered automatically.
- **`SettingsClient.tsx`** — gates the **summary-email recipients** section (`daily-summary`) and the **Notification Log** tab (`notification-log`) behind `useFeatureFlag("email_notifications")`, including the scroll-spy `sectionIds` filter (same pattern used for `report_access` / `timetable_exceptions`).
- **`NotificationLogTab.tsx`** — defensive in-component guard (`return null` when the flag is off) so the surface can never leak if mounted elsewhere.
- **`api/cron/daily-summary/route.ts`** — no-ops (returns `{ skipped: true, reason: "email_notifications flag disabled" }`) when the flag is off, via server-side `getFeatureFlag`, after the existing `CRON_SECRET` auth check.

### Why
The daily SLT summary cron isn't scheduled, the notification outbox worker is a no-op skeleton, and delivery is inline fire-and-forget only. Surfacing a Notification Log / summary-email UI implies a working delivery pipeline that doesn't exist in the pilot. Hiding (not deleting) keeps the wiring intact for go-live.

### Test plan
- `tsc --noEmit` — clean.
- `npm run build` — succeeds end-to-end.
- Unit: `featureFlags`/`demoFeatureFlags` suites pass (10/10).
- Integration (`tests/api/cron-daily-summary.test.ts`): existing delivery-path tests now enable the flag via a shared `beforeEach`; added a dedicated test asserting the cron no-ops (no emails, no `OutboundMessage` rows) when the flag is off. Integration suite needs Postgres on :5433 (unavailable locally) — **runs in CI**.


guangshinhaha:
LGTM 👍

---

Issue #481, opened 2026-07-03
bug: in-app-notifications unreadCount undercounts when a user has more than 50 notifications


guangshinhaha:
## Summary
`GET /api/in-app-notifications` computes the unread badge count from a **capped slice** of the 50 most-recent notifications, so the count silently undercounts once a user has more than 50 notifications.

## Location
`src/app/api/in-app-notifications/route.ts:11-17`

```ts
const notifications = await prisma.notification.findMany({
  where: { userId: session.user.id },
  orderBy: { createdAt: "desc" },
  take: 50,
});
const unreadCount = notifications.filter((n) => !n.read).length;   // bounded by the 50-row slice
```

The unread count is bounded by whatever unread rows fall in the newest 50; once ≥50 read notifications are newer than the unread ones, genuinely-unread items are excluded entirely.

## Failure scenario
A `SCHOOL_ADMIN`/`RELIEF_COORDINATOR` who accumulates >50 notifications (e.g. many teacher-decline notifications — created one-per-admin per decline) sees a wrong/stale unread badge. The committed test only seeds 5 rows, so it doesn't exercise the cap.

## Fix direction
Compute `unreadCount` with a separate `prisma.notification.count({ where: { userId, read: false } })` rather than from the paged slice.

---
*Filed from the 2026-07 full-app audit. Verified against code.*

guangshinhaha:
Fixed on `main` — verified against `origin/main` (e5350ed). Landed in **#502** (`8548e9b`).

`in-app-notifications/route.ts`: `unreadCount` is now a separate `prisma.notification.count({ where: { userId, read: false } })` run in `Promise.all` alongside the `take: 50` `findMany` — no longer `notifications.filter(n => !n.read).length` over the capped slice, so it's accurate past 50. Closing as completed.

Issue #467, opened 2026-07-03
bug: positions silently coerce invalid defaultMaxReliefPerWeek to null (uncapped) instead of 400


guangshinhaha:
## Summary
`positions` create/update silently coerce an invalid `defaultMaxReliefPerWeek` to `null` (which means "no cap") instead of returning 400. A form value sent as a string, `0`, or a negative all become "uncapped" with a 200/201 — the opposite of intent.

## Location
`src/app/api/positions/route.ts:46-50` and `src/app/api/positions/[id]/route.ts:45-49`

```ts
typeof defaultMaxReliefPerWeek === "number" && defaultMaxReliefPerWeek >= 1
  ? Math.floor(defaultMaxReliefPerWeek)
  : null   // "3" (string), 0, negative all silently → null (uncapped)
```

## Failure scenario
An admin sets a position's weekly relief cap to `0` (intending "never assign") or a client sends the number as a string. The value is silently stored as uncapped, and auto-assign's weekly-cap logic treats the position as having no limit.

## Fix direction
Validate the field (reject non-numeric / <1 with 400, or define explicit semantics for `0`), rather than silently coercing to null.

---
*Filed from the 2026-07 full-app audit. Verified against code.*

guangshinhaha:
Fixed on `main` — verified against `origin/main` (e5350ed). Landed in **#502** (`8548e9b`).

Both `positions/route.ts` (POST) and `positions/[id]/route.ts` (PATCH) now **400** when `defaultMaxReliefPerWeek` is a non-number, non-finite, or `< 1` (catches string, 0, negative, NaN). The old `... >= 1 ? Math.floor(...) : null` silent coercion is gone; `null` still explicitly clears the cap. Closing as completed.

Issue #466, opened 2026-07-03
security: /api/import-timetable leaks raw internal error messages in 500 responses


guangshinhaha:
## Summary
`/api/import-timetable` returns raw internal error messages (`err.message`) in its 500/failure responses. Prisma/driver error strings can include SQL fragments, constraint names, and column names, which are echoed to the client.

## Location
`src/app/api/import-timetable/route.ts:543-552` (also ~:108-113, ~:272-277, ~:460-466)

The outer catch returns `{ error: message }` where `message = err.message`; per-handler catches return `Import failed: ${message}`. Repo convention elsewhere is a generic "Internal server error" with detail only in `console.error`.

## Failure scenario
Any unexpected DB error during an admin import echoes internal engine details (schema/constraint names) to the browser.

## Fix direction
Return a generic message to the client; log the detail server-side only. Keep the actionable validation errors (parse/plan errors) but scrub raw engine strings.

---
*Filed from the 2026-07 full-app audit.*

guangshinhaha:
Fixed on `main` — verified against `origin/main` (e5350ed). Landed in **#502** (`8548e9b`).

`import-timetable/route.ts`: the outer catch and all per-handler 500 catches now `console.error` the detail and return a generic `{ error: "Internal server error" }` — no more `err.message` in 500 bodies. (The one remaining `${message}` interpolation is a deliberate **422** parse-failure path for user data, not an internal-error leak.) Closing as completed.

Issue #462, opened 2026-07-03
bug: onboarding/complete returns 500 instead of 401 for unauthenticated callers (#383 convention)


guangshinhaha:
## Summary
`POST /api/onboarding/complete` returns **500** for an unauthenticated caller instead of **401**, violating the repo-wide convention established in #383/#406 (commit 490a1b8).

## Location
`src/app/api/onboarding/complete/route.ts:19` (+ catch)

`requireAuth()` throws `"Unauthorized"`, but the bare `catch { return 500 }` swallows it. Other guarded routes map this via `authErrorResponse`.

## Failure scenario
A logged-out client POSTing `/api/onboarding/complete` gets 500 instead of 401, so client-side session-expiry handling can't distinguish "log in again" from a server fault.

## Fix direction
Wrap the handler and route the caught error through `authErrorResponse(err)` before falling back to 500.

---
*Filed from the 2026-07 full-app audit.*

guangshinhaha:
Fixed on `main` — verified against `origin/main` (e5350ed). Landed in **#502** (`8548e9b`).

`onboarding/complete/route.ts` catch now routes through `authErrorResponse(err)` before the generic 500, so an unauthenticated POST (`requireAuth()` throws "Unauthorized") maps to **401** per the #383/#406 convention. Closing as completed.

PR #502, opened 2026-07-03
fix(api): return correct 4xx status codes for input-validation error cluster

### LINKED ISSUE #461: bug: periods/school-roles POST return 500 on duplicate instead of 409 (positions handles it)
## Summary
Several config-creation routes create rows without a duplicate pre-check, and their catch blocks map only auth errors — so a duplicate that trips a `@@unique` constraint (Prisma `P2002`) surfaces as a generic **500** instead of a 409/400. The `positions` route already does this correctly (returns 409), so the pattern is inconsistent.

## Locations
- `src/app/api/periods/route.ts:35-44` — `Period @@unique([schoolId, number])`; creating a duplicate period number → P2002 → 500.
- `src/app/api/school-roles/route.ts:39-44` — `SchoolRole @@unique([schoolId, name])`; duplicate role name → P2002 → 500.

Correct precedent: `src/app/api/positions/route.ts:36-41` checks and returns 409.

## Failure scenario
An admin re-adds an existing period number or role name and gets "Internal server error" instead of a clear "already exists" message.

## Fix direction
Pre-check for the duplicate (or catch `P2002`) and return 409 with a field-specific message, matching `positions`.

---
*Filed from the 2026-07 full-app audit. Verified against code. One issue covering both routes (same defect + fix).*

guangshinhaha:
## Summary

Fixes a cluster of handlers that surfaced client errors as generic 500s (or silently mis-coerced input), aligning them with the repo-wide convention (auth throws -> 401/403 via `authErrorResponse`; bad input -> 400/409 via explicit guards; raw engine detail never reaches the client).

| Issue | Route(s) | Fix |
|---|---|---|
| #461 | `periods`, `school-roles` POST | Pre-check the `@@unique` constraint -> **409** (mirrors `positions`) instead of leaking Prisma `P2002` as a 500 |
| #462 | `onboarding/complete` | Route the caught auth throw through `authErrorResponse` -> **401** for logged-out callers, not 500 |
| #466 | `import-timetable` | Unexpected engine errors return a generic **"Internal server error"** (detail stays in `console.error`); actionable parse/plan errors (400/413/422) unchanged |
| #467 | `positions` POST/PATCH | Reject invalid `defaultMaxReliefPerWeek` (non-numeric, 0, negative, NaN) with **400**; `null` still means uncapped |
| #481 | `in-app-notifications` | Compute `unreadCount` via `notification.count({ read: false })`, fixing the undercount past the 50-row display slice |
| #472 (partial) | `user/preferences`, `relief-assignments` | Wrap `req.json()` -> **400** on invalid JSON; guard malformed (NaN) date -> **400** |

## Why

Client-side session-expiry and validation handling can't distinguish "bad input / log in again" from "server fault" when everything comes back 500. #466 additionally scrubs raw Prisma/driver strings (SQL fragments, constraint/column names) from client responses.

Scope note: this PR only touches the `user/preferences` + `relief-assignments` parts of #472; the remaining #472 locations (`export`, `teachers/[id]`, `timetable`, `timetable-overrides`) are owned by other PRs and are intentionally untouched.

## Test plan

- Extended `periods`, `school-roles`, `onboarding`, `positions`, `relief-assignments` integration tests with the new 4xx cases.
- New `in-app-notifications.test.ts` (unread count beyond the 50-row slice), `user-preferences.test.ts` (malformed JSON -> 400), and `import-timetable-error-scrub.test.ts` (mocks the executor to throw a raw engine error, asserts a generic 500 with no leaked detail).
- `tsc --noEmit` clean; `npm test` (unit suite) 744 passing; `npm run build` succeeds. The new `tests/api/**` suite requires Postgres on :5433 and runs in CI.

Closes #461 #462 #466 #467 #481


guangshinhaha:
LGTM 👍

---

PR #516, opened 2026-07-06
fix(import): destroy pdfjs loading task, not the document proxy

guangshinhaha:
## Summary

aSc PDF timetable import failed in prod with **`PDF parsing failed: i.destroy is not a function`** (seen on the ACS Barker upload). The parse actually *succeeds* — it's only the cleanup call that throws, so the whole import discards all extracted data.

**Root cause:** In pdfjs-dist v6 the resolved `PDFDocumentProxy` no longer exposes `destroy()` — teardown moved onto the *loading task*. `parseAscPdf` called `doc.destroy()`, which throws `doc.destroy is not a function` (minified to `i.destroy` on Vercel).

This is the next error in the chain after #510: externalizing pdfjs let the fake worker actually run, which exposed this lifecycle call.

## Why

- The bug fired *after* a fully successful parse (verified: 108 pages / ~3,400 entries extracted from both the Odd and Even Week ACS Barker PDFs), so every import failed with zero data.
- Not ACS-specific — it would fail for any PDF.

## Fix

Hold the loading task and destroy that instead of the document proxy:
```ts
const loadingTask = pdfjsLib.getDocument({ data });
const doc = await loadingTask.promise;
// ...parse...
await loadingTask.destroy();   // was: await doc.destroy()
```

## Test plan

- Added a regression test driving the real `getDocument() → loadingTask.destroy()` lifecycle with a tiny inline (base64) PDF. The ACS Barker staff fixtures are gitignored, so this path was previously only unit-tested at the helper level — which is why the lifecycle bug slipped through.
- Reproduced the original crash against the real ACS Barker PDF → applied fix → both PDFs parse clean.
- `npm test` — 764 passed / 1 skipped.
- `tsc --noEmit` — clean.
- `npm run build` — clean (`✓ Generating static pages (88/88)`).


src/lib/parseAscPdf.ts
```diff
@@ -379,7 +386,7 @@ export async function parseAscPdf(
     allEntries.push(...entries);
   }
 
-  await doc.destroy();
+  await loadingTask.destroy();
```
guangshinhaha, line 389: **[Medium] Teardown is skipped when a page parse throws, leaking the pdfjs worker**

```ts
  const loadingTask = pdfjsLib.getDocument({ data });
  const doc = await loadingTask.promise;
  // ...
  for (let p = 1; p <= doc.numPages; p++) {
    const page = await doc.getPage(p);          // ← can reject on a malformed PDF
    const content = await page.getTextContent(); // ← can reject too
    // ...
  }
  await loadingTask.destroy();  // ← never reached if the loop above throws
```

**Problem:** This PR correctly moves teardown to `loadingTask.destroy()`, but it still only runs on the success path. A corrupt or partially-malformed uploaded PDF that makes `getPage()` / `getTextContent()` reject will propagate out of `parseAscPdf` before `destroy()` is called, leaking the pdfjs worker and document. Reachable from `import-timetable/route.ts:312-315`, where two user-supplied PDFs are parsed via `Promise.all` — an attacker or a bad export can trigger it, and leaked workers accumulate across imports on a reused serverless instance.

**Suggestion:**
```ts
  const loadingTask = pdfjsLib.getDocument({ data });
  const doc = await loadingTask.promise;
  try {
    // ... existing page loop, building pages/allEntries/warnings ...
    return { pages, entries: allEntries, warnings };
  } finally {
    await loadingTask.destroy();
  }
```

---

guangshinhaha:
## Code Review Summary

| Severity | Count |
|----------|-------|
| Critical | 0 |
| High     | 0 |
| Medium   | 1 |
| Low      | 0 |

## What Looks Good
- Correct root-cause fix for the pdfjs-dist v6 API change: teardown moved from the (now `destroy()`-less) `PDFDocumentProxy` to the loading task, matching the installed `^6.0.227`.
- Type model updated honestly — a dedicated `PdfLoadingTask` interface with no `any` casts, and `destroy()` removed from `PdfDoc` so the compiler enforces the correct call site.
- Ships a regression test that drives the real `getDocument() → loadingTask.destroy()` lifecycle against an inline minimal PDF, so the prod "i.destroy is not a function" failure can never silently return.
- The `getDocument()` return-type widening stays backward-compatible — callers depend only on the unchanged `PdfParseResult`, so no call sites break.

---

PR #517, opened 2026-07-06
fix(import): recognize 1/0 as Admin values in staff CSV import

superb-sushi:
## Summary
- The "Admin" column in the teacher/staff CSV import only recognized "yes"/"true" as truthy. Files exported with 1/0 conventions (e.g. from Excel/Sheets boolean columns) silently parsed as `false` — an admin grant would be dropped with no error or warning. Added `parseCsvBoolean()` in `src/lib/teacherCsv.ts` to also accept "1".
- Added two notes to the import modal (`ImportModal.tsx`):
  - Existing staff are matched by exact **Name** — a mismatched name creates a new teacher instead of updating the existing one.
  - Admin access is only granted for institutional email domains (`moe.edu.sg`, `hci.edu.sg`, or the `reliefcher.com` exemption) the Admin column is silently skipped for anyone else, even if it says yes.

## Why
Reported while testing bulk staff imports: an Admin column using 1/0 values wasn't being reflected in teacher records after import, with no indication why. Root cause was the boolean parser's narrow accepted- value set. The modal notes address the two most common points of confusion admins hit during import (name-matching semantics and the domain restriction on admin grants), surfaced while investigating this.

## Test plan
- [x] `npm test` — 772 passing, incl. new case covering "1"/"0"/"no"
      Admin values
- [x] `npm run build` — clean, no type errors
- [x] `npm audit` — no new high/critical (7 pre-existing moderate,
      unrelated)
- [x] Manual click-through: re-import a CSV with `Admin` set to `1`
      for an existing teacher and confirm the admin grant applies


superb-sushi:
Fixes #518

PR #520, opened 2026-07-06
feat(timetable): allow period 0 and offer create-or-skip for missing periods

guangshinhaha:
## Summary

Two timetable-import UX fixes surfaced testing real school files:

**A — Period number 0 is now valid.** Schools run a "period 0" registration/assembly slot (e.g. HCI's aSc export has period 0 = 07:40–08:00). It was rejected everywhere; the floor drops from 1 to 0 while negatives stay rejected.

**B — Missing periods offer a choice instead of a hard block.** A CSV/PDF/XLSX file that references a period the school hasn't set up used to disable the Import button and tell the admin to go create it by hand. Assembly periods like 102/103 (from PDF `AS2`/`AS3`) always tripped this. It's now a user-selectable choice: **auto-create** the missing periods (placeholder times) or **skip** the lessons in them. (ASC-XML was already exempt — it auto-creates periods from the file's metadata.)

## Why

- Period 0 is a genuine slot in real exports, so importing HCI-style timetables was impossible.
- The hard block made PDF/CSV/XLSX imports unusable whenever assembly/registration periods appeared, forcing manual period setup that no admin thinks to do up front.

### A — Period 0
- `PeriodNumberSchema` (used by `POST /api/periods`) floor `1 → 0`, message updated.
- CSV parser (`timetableImport.ts`) accepts period 0; still warns on negatives; message updated.
- Settings "Period Number" input `min={1} → min={0}`.
- `parseAscXml.ts`: keeps a genuine `period="0"` **period element and card**, but still skips a truly *absent* period attribute. Previously `parseInt(el.getAttribute("period") || "0")` coerced a missing attribute to `0`, making "no period" and "period 0" indistinguishable — now the raw attribute is checked for null/empty before parsing.

### B — Missing-period choice
- `timetableImportExecutor.ts`: period creation is no longer ASC-XML-only — any source can pass period defs in `input.periods` and have them created. Removed the CSV-specific fail-fast on missing periods in favour of the existing generic per-entry skip path. `incomingPeriodNumbers` now includes supplied periods for all sources.
- `route.ts`: `handleCsvConfirm` / `handlePdfConfirm` accept a `missingPeriodMode` (`"create"` | `"skip"`; absent ⇒ skip). For `create`, the route computes the missing referenced numbers and builds placeholder-timed defs; the response surfaces `autoCreatedPeriods`. Both preview endpoints now return `missingPeriodEntryCount` so the UI can show how many lessons "skip" would drop.
- `ImportTimetableTab.tsx`: the missing-periods panel now presents the two radio options; the Import button is disabled only until a mode is picked (plus the existing `!importMode` / `isImporting` conditions), not merely because periods are missing. Post-import message notes auto-created placeholder periods and the skipped count.

## Design decision — placeholder times for auto-created periods

PDF/CSV/XLSX carry **no period times** (`parseAscPdf` doesn't extract them; CSV/XLSX have none), but `Period` requires `startTime`/`endTime` with `endTime > startTime`. Auto-created periods therefore get a **clearly-placeholder `00:00`–`00:01`** and the post-import result tells the admin to edit real times in Settings → Periods. Times are **not** derived from the period number — numbers like `102` aren't valid hours. This is the recommended default from the task; no cleaner established convention existed in the codebase.

## Period-0 sentinel usages found

Grepped for `0`-as-sentinel around period numbers. Nothing breaks by allowing real 0:
- `timetableImportExecutor.ts` (`newlySkipped` builder): `entry.period.kind === "existing" ? 0 : entry.period.number` — the `0` is a **display placeholder in a skip-reason payload** for a period resolved by existing ID (number unknown at that point). It's not a real period and never becomes one; allowing real period 0 doesn't affect it.
- XLSX header detection in `ImportTimetableTab.tsx` uses `Number(header) > 0` to pick period columns. Left as-is intentionally: empty cells stringify to `""` and `Number("") === 0`, so `>= 0` would misclassify blank columns as period 0. STA's TeachersMaster format has no period 0, so this is correct.

## No Prisma migration

Confirmed — neither task changes `prisma/schema.prisma`. All changes are validation floors, parser logic, executor/route behaviour, and UI. `Period.number` was already a plain `Int` with no DB-level `>= 1` constraint.

## Test plan

- **Unit (`npm test`)** — full suite **770 passed, 1 skipped**. New/updated: `PeriodNumberSchema` accepts 0 / rejects −1; CSV parser accepts period 0, warns on negatives; new `parseAscXml.test.ts` (jsdom env) pins real `period="0"` kept vs. absent-attribute skipped; executor tests updated from CSV fail-fast to skip-vs-auto-create planning.
- **Integration (`npm run test:api`)** — `periods` + `import-timetable` suites **40 passed**: `/api/periods` POST accepts `number: 0` and rejects `-1`; PDF/CSV confirm with a missing period + `create` creates a `00:00`-timed period and imports all; with `skip` imports the rest and reports the skipped lessons.
- **`tsc --noEmit`** — clean on all changed files.

> Environment note: this sandbox's egress policy blocks the private `@flow` registry, so a clean `npm ci` (and therefore the local `next build` / full pre-push `tsc`) couldn't run here — the pre-push hook failure was confined to that dependency-graph corruption in untouched files (`ioredis`/`@flow`/`date-fns`), not to any changed code. Please confirm the Vercel/CI build is green on this branch, since `next build` could not be exercised locally.

https://claude.ai/code/session_01WPWJqDgsxPPC8CqP2aq9WX

---
_Generated by [Claude Code](https://claude.ai/code/session_01WPWJqDgsxPPC8CqP2aq9WX)_


src/app/api/import-timetable/route.ts
```diff
@@ -150,6 +197,12 @@ async function handleCsvPreview(
   const missingPeriods = parsed.periodNumbers
     .filter((n) => !existingPeriodNums.has(n))
     .sort((a, b) => a - b);
+  // How many parsed rows fall in a missing period — the "skip" option drops
+  // exactly these, so the UI shows the count next to the choice.
+  const missingPeriodSet = new Set(missingPeriods);
+  const missingPeriodEntryCount = parsed.rows.filter((r) =>
```
guangshinhaha, line 222: **[Medium] CSV preview under-reports skipped rows once missing periods no longer block the import**

```ts
  const missingPeriodEntryCount = parsed.rows.filter((r) =>
    missingPeriodSet.has(r.periodNumber),
  ).length;

  let planSkippedWarnings: { lineNumber: number; message: string }[] = [];
  if (missingPeriods.length === 0) {   // ← planner only runs when nothing is missing
    // ... surfaces unresolvable-teacher rows as preview warnings
  }
```

**Problem:** The preview only runs `planImport` to surface *teacher*-unresolvable skips when `missingPeriods.length === 0`. Before this PR that gate was fine — a CSV with missing periods aborted at confirm, so preview warnings didn't matter. Now confirm proceeds (create/skip), so a CSV that has **both** missing periods **and** rows whose teacher can't be resolved will preview without any warning about those teacher rows, yet confirm silently drops them (they come back in `skipped`). The admin also sees "skip drops N lessons" counting only the missing-period rows, undercounting the true number dropped. The gating comment on line 190 ("The planner treats CSV missing-periods as a fatal error, so we don't call it here") is now stale — the planner no longer treats them as fatal.

**Suggestion:** Run the planner in preview regardless of `missingPeriods`, and drop only the period-missing skips from the warning list (those already have the dedicated panel), keeping teacher skips:
```ts
  const plan = await planImport({ schoolId, source: "csv", mode: "merge",
    teachers: teacherInputs, periods: [], entries: entryInputs });
  if (!isPlanError(plan)) {
    planSkippedWarnings = plan.skippedEntries
      .filter((s) => !missingPeriodSet.has(s.periodNumber)) // period-missing shown in the panel
      .map((s) => ({ lineNumber: s.lineNumber ?? 0,
        message: `${s.reason} (teacher "${s.teacherName}", period ${s.periodNumber})` }));
  }
```
And refresh the comment on line 188–190. (PDF preview has no equivalent teacher-warning pass, so it's unaffected.)

---
*🤖 aif-code-review · Claude Code*

---

src/app/api/import-timetable/route.ts
```diff
@@ -244,12 +303,18 @@ async function handleCsvConfirm(
     weekType: r.weekType,
   }));
 
+  const { periods, autoCreatedPeriods } = await resolveMissingPeriods(
+    schoolId,
+    entryInputs,
+    body.missingPeriodMode ?? "skip",
```
guangshinhaha: **[Low] `missingPeriodMode` is never validated — any unexpected value silently means "skip"**

```ts
  const { periods, autoCreatedPeriods } = await resolveMissingPeriods(
    schoolId,
    entryInputs,
    body.missingPeriodMode ?? "skip",   // ← "cerate" typo / garbage ⇒ silently drops lessons
  );
```

**Problem:** `body` is untyped JSON from the client; `missingPeriodMode` is only asserted as `MissingPeriodMode` at the type level, never checked at runtime. `resolveMissingPeriods` treats anything `!== "create"` as skip, so a malformed value (client bug, a `"cerate"` typo, a stale client) silently discards the missing-period lessons instead of failing loudly. The rest of this route validates its inputs with Zod / explicit `mode`/`timetableCycle` guards, so this is the odd one out. Same pattern at the PDF confirm (line 513).

**Suggestion:** Reject an out-of-set value rather than defaulting:
```ts
if (body.missingPeriodMode !== undefined &&
    body.missingPeriodMode !== "create" &&
    body.missingPeriodMode !== "skip") {
  return NextResponse.json(
    { error: "missingPeriodMode must be 'create' or 'skip'." }, { status: 400 });
}
```
(Absent is still a legitimate "skip" default; only a *present but invalid* value should 400.)

---
*🤖 aif-code-review · Claude Code*

---

guangshinhaha (commented):
## Code Review Summary

| Severity | Count |
|----------|-------|
| Critical | 0 |
| High     | 0 |
| Medium   | 1 |
| Low      | 1 |

Two well-scoped fixes, cleanly implemented with matching unit + integration coverage. No correctness or data-loss bugs found. Two inline comments left — both are observability/robustness polish, neither blocks merge.

## What Looks Good
- **`parseAscXml` absent-vs-zero distinction is exactly right** — guarding the raw attribute for `null`/empty *before* `parseInt` fixes the real root cause (`parseInt(x ?? "0")` conflating "no period" with "period 0"), rather than patching it downstream.
- **The executor refactor is behaviour-preserving for ASC-XML while extending cleanly** — collapsing the period-plan into one `replace && asc-xml` branch vs. a shared create/upsert split keeps the old paths identical and lets CSV/PDF reuse the existing deferred-resolve machinery instead of a parallel code path.
- **`resolveMissingPeriods` only ever emits defs for genuinely-missing numbers, and `planImport` re-checks existence** — the double-read means a concurrently-created period routes to upsert, so there's no `@@unique(schoolId, number)` violation.
- **Placeholder times (`00:00–00:01`) are honestly surfaced** — the post-import message tells the admin to edit them, and times are deliberately *not* derived from the period number. Good call flagging that `102` isn't a valid hour.

## One thing to confirm before merge
CI is not reporting on this PR (`check_runs` is empty, `mergeable_state: unstable`) and the description notes `next build` couldn't run locally due to the `@flow` egress block. Per the repo's deployment-verification rule, please confirm the Vercel build is green on `116c23e` before merging — the changes live inside an existing client component with no new top-level `useSearchParams`, so I'd expect it to pass, but it hasn't been exercised.

---
*🤖 aif-code-review · Claude Code*

---

src/app/api/import-timetable/route.ts
guangshinhaha: Retracting this finding — it was wrong on closer analysis. `parseTimetableCsv` builds `teacherNames` from exactly the rows it emits (every valid row's teacher is added to `teacherSet`), so a CSV *declares every teacher it references*. In `planImport` merge mode every such teacher is therefore either existing or in `newTeacherNames` and always resolves — a CSV row can never be skipped for a teacher reason in preview. The only skips the planner could produce here are the missing-period ones, which the dedicated panel already surfaces. So `planSkippedWarnings` is empty for CSV whether or not the planner runs, and there's no under-reporting. The `missingPeriods.length === 0` gate is correct as-is.

The one real thing in this comment was the stale gating comment ("planner treats CSV missing-periods as a fatal error"), which was left over after the create-or-skip change removed that fail-fast. Fixed in 83c50ea to state the actual reason. Resolving.

---
*🤖 aif-code-review · Claude Code*

---

src/app/api/import-timetable/route.ts
guangshinhaha: Fixed in 83c50ea. Added a shared `resolveMissingPeriodMode` helper used by both the CSV and PDF confirm handlers: absent still defaults to `"skip"`, but a present-but-invalid value now returns 400 instead of silently dropping the missing-period lessons. Covered by new integration tests (invalid mode → 400 + nothing created, on both CSV and PDF; absent → skip). Resolving.

---
*🤖 aif-code-review · Claude Code*

---

src/lib/timetableImportExecutor.ts
```diff
@@ -160,37 +165,28 @@ export async function planImport(input: PlanInput): Promise<ImportPlan | PlanErr
...
-    }
-  }
-
-  // ── Period plan (ASC-XML brings period metadata; CSV does not) ───────
+  // ── Period plan ──────────────────────────────────────────────────────
+  // Any source may now supply period definitions in `input.periods` and have
+  // them created (previously ASC-XML only). CSV/PDF/XLSX use this to
+  // auto-create the periods a file references but the school hasn't set up
+  // yet (e.g. assembly periods 102/103 from a PDF). Entries whose period is
+  // neither existing nor supplied here are skipped, not fatal — the caller
```
guangshinhaha, line 173: **[Low · KIV] Removing the CSV missing-period fatal guard silently changes the timetable-overrides import path**

\`\`\`ts
  // ── Period plan ──────────────────────────────────────────────────────
  // Any source may now supply period definitions in `input.periods` and have
  // them created (previously ASC-XML only). CSV/PDF/XLSX use this to
  // auto-create the periods a file references but the school hasn't set up
  // yet ... Entries whose period is
  // neither existing nor supplied here are skipped, not fatal — the caller  // ←
  // decides between "auto-create" (pass defs) and "skip" (pass none).
\`\`\`

**Problem:** `planImport` is called by two routes. The main `import-timetable` route was updated to pair this change with `resolveMissingPeriods` + a `missingPeriodMode` choice, so the admin picks create-vs-skip. But the second caller — `src/app/api/[slug]/timetable-overrides/[id]/import/route.ts` (source `csv`, `periods: []`, no `missingPeriodMode`) — was **not** updated. Before this PR, a CSV override referencing a period the school doesn't have returned a fatal `PlanError` → the route responded **422** with an actionable "Add them in Settings → Periods before importing." Now those rows are silently dropped into `skippedEntries` and the override import proceeds with the rest. The clear, blocking error the admin relied on is gone, and there is no create option on that path — it can only skip.

Since timetable exceptions/overrides are behind the `timetable_exceptions` flag (OFF this phase), this is **not a live bug** — filing as **KIV**, not a merge blocker. But it should be resolved before that feature ships.

**Suggestion:** When the override feature is enabled, give its route the same `resolveMissingPeriods`/`missingPeriodMode` handling as the main import route, or re-assert an explicit missing-period guard there so the override import keeps failing loudly instead of silently skipping:
\`\`\`ts
// timetable-overrides/[id]/import/route.ts — before planImport
const existing = await prisma.period.findMany({ where: { schoolId: school.id }, select: { number: true } });
const have = new Set(existing.map((p) => p.number));
const missing = [...new Set(parsed.rows.map((r) => r.periodNumber))].filter((n) => !have.has(n)).sort((a, b) => a - b);
if (missing.length > 0) {
  return NextResponse.json(
    { error: `Periods ${missing.join(", ")} are referenced but do not exist. Add them in Settings → Periods first.` },
    { status: 422 },
  );
}
\`\`\`

---

guangshinhaha:
## Code Review Summary

| Severity | Count |
|----------|-------|
| Critical | 0 |
| High     | 0 |
| Medium   | 0 |
| Low      | 1 |

One Low finding, flagged **KIV** (dormant `timetable_exceptions` path) — not a merge blocker. The two prior review threads (CSV preview skip-count under-reporting; `missingPeriodMode` runtime validation) are both addressed and resolved in the current diff.

## What Looks Good
- **Period-0 as a real slot, not a sentinel:** the `parseAscXml` fix replaces the lossy `parseInt(attr || "0")` with an explicit null/blank guard (`rawPeriod === null || rawPeriod.trim() === ""`) plus `Number.isFinite(num) && num >= 0`, so a genuine `period="0"` survives while a truly-absent attribute is still dropped — and the accompanying jsdom unit tests pin exactly that distinction.
- **Contained schema widening:** `PeriodNumberSchema.min(0)` has a single consumer (the periods route), and no `period.number` falsy/`|| 0`/`> 0` sentinel logic in the codebase breaks on a real 0 — verified the surviving `?? 0` / sort-fallback sites are nullish-safe.
- **Missing-period handling is safe by construction:** `missingPeriodMode` is runtime-validated (present-but-invalid ⇒ 400, absent ⇒ skip), auto-created placeholders use `00:00–00:01` which satisfies `endTime > startTime`, and the `period.createMany` stays `schoolId`-scoped so multi-tenancy holds. xlsx correctly rides the CSV confirm path and transmits the mode.
- **Strong test coverage:** create/skip/absent/invalid-mode cases are covered at both the API-integration and unit layers, including the DB assertion that the placeholder period is (not) created.

---

PR #521, opened 2026-07-07
feat: AI-assisted timetable seeding (superadmin)

guangshinhaha:
## Summary
Superadmin-only tool that uses an LLM (via **OpenRouter**) to normalise a messy, text-extractable timetable file into the app's existing flat-CSV format, then seeds it into a chosen school through the **existing deterministic import path** with a human review gate. The model is a *normaliser, not a committer* — it never touches the DB.

Flow: pick school → paste/upload → AI drafts CSV → **human reviews/edits** → confirm → `parseTimetableCsv → planImport → executeImport` (one transaction, chosen `schoolId`).

## Why
Onboarding timetables arrive in shapes the deterministic parsers can't handle (odd `.xlsx`, non-aSc PDFs, one-off CSVs). The 2026-06 agentic-ops spike concluded an LLM only pays off where a task is *both* unstructured *and* backed by real substrate data — relief assignment fails that test, but messy timetable ingestion is the one place it holds. The LLM's silent-error risk is contained by keeping a human in the loop and never letting model output reach the DB without an explicit confirm.

Design spec: `docs/superpowers/specs/2026-07-06-ai-timetable-seed-design.md`
Implementation plan: `docs/superpowers/plans/2026-07-06-ai-timetable-seed.md`

## Scope / gating
- `SUPERADMIN` role **+** OFF-by-default `ai_timetable_seed` feature flag (auth checked before flag; both fail closed).
- Text-only inputs (`.csv`/`.txt`/`.xlsx`/`.pdf`/paste). No vision.
- **No new npm dependency** — OpenRouter via plain `fetch`. New env: `OPENROUTER_API_KEY`, `OPENROUTER_MODEL` (model id swappable; can point at a Claude model).
- Reuses the existing import executor with `source: "csv"` — no parallel commit engine.
- Byte (5 MB) + row (20 000) caps and 422 extraction errors mirror the existing import route.

## Test plan
- **Unit:** OpenRouter client, normaliser (prompt + size cap + fence strip), extraction adapters (csv/txt/xlsx/pdf), flag registration. Model call mocked.
- **Integration (`tests/api`):** 403 unauth / non-superadmin / flag-off, 404 unknown school, draft happy path, **confirm writes to the chosen school only** (tenancy: other school stays 0), 413 row cap, 422 unsupported file. 8/8 passing.
- `npm test` 782 passing · `npm run test:api` green · `npm run build` clean (89/89 static pages).

## Not in this PR (KIV)
- Vision/scan input; oversized-file chunking; a fortnightly `timetableCycle` selector on the seed UI (seeded week-typed rows don't auto-flip a school's cycle).

## Human verification still needed
- Live OpenRouter smoke (real key + a real messy file) — deferred; requires a key not available in CI.


src/lib/extractTimetableText.ts
```diff
@@ -0,0 +1,74 @@
...
+      chunks.push(cells.join("\t"));
+    });
+  });
+  return chunks.join("\n");
+}
+
+async function extractPdfText(bytes: Buffer): Promise<string> {
+  const pdfjs = await loadPdfjs();
+  const ab = bytes.buffer.slice(bytes.byteOffset, bytes.byteOffset + bytes.byteLength);
+  const doc = await pdfjs.getDocument({ data: ab as ArrayBuffer }).promise;
```
guangshinhaha: **[Medium] `extractPdfText` never destroys the pdfjs loading task — resource leak on every PDF draft**

```ts
async function extractPdfText(bytes: Buffer): Promise<string> {
  const pdfjs = await loadPdfjs();
  const ab = bytes.buffer.slice(bytes.byteOffset, bytes.byteOffset + bytes.byteLength);
  const doc = await pdfjs.getDocument({ data: ab as ArrayBuffer }).promise; // ← loading task discarded
  // ... reads pages ...
  return pages.join("\n");   // ← returns without tearing the task down
}
```

**Problem:** The loading task returned by `getDocument()` is discarded — only its `.promise` is kept — so `.destroy()` is never called. Each PDF upload on the draft path leaks the pdfjs task/worker resources. This is exactly the leak the recent fix `6a3f40a` ("destroy pdfjs loading task, not the document proxy") addressed in `src/lib/parseAscPdf.ts`, which keeps the task and calls `await loadingTask.destroy()` after reading. The `PdfjsLib` interface here also omits `destroy()`, so the type currently forbids the cleanup.

**Suggestion:**
```ts
interface PdfLoadingTask {
  promise: Promise<{
    numPages: number;
    getPage: (n: number) => Promise<{
      getTextContent: () => Promise<{ items: { str?: string }[] }>;
    }>;
  }>;
  destroy(): Promise<void>;
}
interface PdfjsLib {
  getDocument: (opts: { data: ArrayBuffer }) => PdfLoadingTask;
}

async function extractPdfText(bytes: Buffer): Promise<string> {
  const pdfjs = await loadPdfjs();
  const ab = bytes.buffer.slice(bytes.byteOffset, bytes.byteOffset + bytes.byteLength);
  const loadingTask = pdfjs.getDocument({ data: ab as ArrayBuffer });
  const doc = await loadingTask.promise;
  try {
    // ... existing page-reading loop ...
    return pages.join("\n");
  } finally {
    await loadingTask.destroy();
  }
}
```

---

src/lib/openrouter.ts
```diff
@@ -0,0 +1,56 @@
...
+  };
+}
+
+export async function openRouterChat(messages: ChatMessage[]): Promise<ChatCompletion> {
+  const apiKey = process.env.OPENROUTER_API_KEY;
+  const model = process.env.OPENROUTER_MODEL;
+  if (!apiKey) throw new Error("OPENROUTER_API_KEY is not set");
+  if (!model) throw new Error("OPENROUTER_MODEL is not set");
+
+  const res = await fetch(OPENROUTER_URL, {
```
guangshinhaha: **[Low] OpenRouter `fetch` has no timeout — a stalled provider hangs the request**

```ts
const res = await fetch(OPENROUTER_URL, {   // ← no AbortSignal / timeout
  method: "POST",
  headers: { Authorization: `Bearer ${apiKey}`, "Content-Type": "application/json", /* ... */ },
  body: JSON.stringify({ model, temperature: 0, messages }),
});
```

**Problem:** OpenRouter fronts many upstream providers and LLM completions can legitimately take tens of seconds or stall entirely. With no `AbortSignal`, a hung upstream connection keeps the superadmin request open until the platform function timeout, holding the connection and the (up to 200k-char) request in memory with no clean error. A bounded timeout turns this into a fast, actionable 5xx.

**Suggestion:**
```ts
const res = await fetch(OPENROUTER_URL, {
  method: "POST",
  headers: { /* ... */ },
  body: JSON.stringify({ model, temperature: 0, messages }),
  signal: AbortSignal.timeout(60_000), // fail fast instead of hanging
});
```
(`AbortSignal.timeout` is available on the Node runtime this app targets; the resulting `TimeoutError` is already funnelled into the caller`s catch → 422/500.)

---

guangshinhaha:
## Code Review Summary

| Severity | Count |
|----------|-------|
| Critical | 0 |
| High     | 0 |
| Medium   | 1 |
| Low      | 1 |

## What Looks Good
- **The AI is a normaliser, not a committer.** The drafted CSV rides the existing `parseTimetableCsv → planImport → executeImport` rails unchanged, so tenancy scoping, period resolution, and the single-transaction commit are all reused rather than reimplemented — no parallel commit engine to keep in sync.
- **Defense-in-depth gating.** Both the server page (`getFeatureFlag` + `requireSuperAdmin` → redirect) and the API route (`requireEnabledSuperadmin` → 403) enforce role + OFF-by-default flag; the sidebar link is a UI convenience gated on the same flag, not the security boundary.
- **Fails safe on bad model output.** Draft returns the raw CSV with the parse error (never auto-commits), and `parseTimetableCsv` returning `fatal` on 0 valid rows means an empty/garbage draft can never reach `executeImport` in `replace` mode — closing the obvious wipe-the-timetable data-loss path.
- **Thorough test coverage** across the new surface: auth (401/403/flag-off), 404 bad school, tenancy isolation (rows land only in the chosen school), row-cap 413, and unsupported-file 422.

---

guangshinhaha:
Closing unmerged. The AI timetable-seeding step is moving OUT of the government codebase to avoid external-LLM review contention (IDSC). It is replaced by a Claude Code skill (`timetable-to-csv`) — Claude itself is the AI, no OpenRouter, no outbound call — that emits the exact CSV the existing manual Timetable Import already accepts. Spec: docs/superpowers/specs/2026-07-08-timetable-to-csv-standalone-design.md. Skill PR: feat/timetable-to-csv-skill. This branch is kept as the porting source for the phase-2 standalone web app — please do NOT delete it.

Issue #276, opened 2026-06-10
security concerns with loke


guangshinhaha:
Security concerns raised in the session with Loke (granola transcript in comments below). Tracking every item as a checklist — verified against the codebase 2026-06-18.

## ✅ Shipped / verified done

- [x] **Rate-limiting on absence reporting** — `checkRateLimit` in the absence-reports route (per-IP for public, per-user for authed), returns `429 + Retry-After`. *(Phase 23 PR d)*
- [x] **Gate `/reporting` behind SSO** — page + API are auth-required for everyone (Issue #330). Unauthed/wrong-school visitors see only the MOE iCON sign-in; no roster/subject/timetable leak. `publicReportRequiresAuth` (`n`) is now a no-op.
- [x] **Self-report-only constraint** — signed-in teachers are locked to their own teacher row; only `SCHOOL_ADMIN` can submit on behalf of others (Issue #330). Caps blast radius of a compromised identity.
- [x] **Non-`@moe` SSO for Hwa Chong** — `hci.edu.sg` is in `ALLOWED_EMAIL_DOMAINS` (`src/lib/authDomains.ts`), enforced in the Google callback after an `email_verified === true` check. *(resolves the #291 auth concern)*
- [x] **Timetable upload — size / row DoS caps** — 5 MB body cap + 20k row cap server-side (`413`), plus client-side file-size check. *(#284)*
- [x] **Timetable upload — format whitelist** — only CSV / ASC-XML / PDF handler paths exist (no generic file sink), matching the "csv, pdf, xml — the rest cannot" position.
- [x] **Timetable upload — no unsafe persistence** — uploads are parsed in-memory from the request body (CSV text / PDF base64 / pre-parsed XML); verified zero filesystem/blob writes in the import route and parsers. Nothing is stored, so nothing to delete.

## 🟡 Mostly covered — light verification outstanding

- [ ] **Upload content validation / sanitization** — shape validation runs (validate-shape → plan → execute), DB writes are parameterized via Prisma (no SQLi), and React auto-escapes on render (no raw-HTML sink). *Gap:* no dedicated test for malicious/escaped payloads surviving ingestion. Add a covering test, then close.

## 🛡️ Open — security-review deliverables (pre-pilot, none started)

[...]

PR #531, opened 2026-07-09
feat(security): malware-scan uploaded files via ClamAV REST

guangshinhaha:
## Summary

Scans every uploaded file through the [benzino77/clamav-rest-api](https://github.com/benzino77/clamav-rest-api) ClamAV service **before it is parsed**, gating the upload on the verdict. The presence of `MALWARE_SCAN_URL` is the on/off gate.

- **Server-side scan** at the two paths where raw bytes reach the server: timetable **PDF** import (`parseAscPdf` — the only server-side binary parser) and the **pool-teacher multipart CSV**. Infected → 422, scanner unreachable → 503.
- **Client pre-scan** for browser-parsed uploads (timetable CSV / aSc XML / STA XLSX / override CSV): the raw `File` is POSTed to a new `POST /api/scan-file` (SCHOOL_ADMIN) endpoint before the browser parses it; infected/scanner-down blocks the upload with a friendly message.
- **New lib:** `src/lib/malwareScan.ts` (server client + `assertFileClean` + `malwareScanErrorResponse`) and `src/lib/malwareScanClient.ts` (browser `scanFileBeforeUpload`).
- **Infra:** `clamav` + `clamav-rest` added to `docker-compose.local.yml`; `MALWARE_SCAN_URL` documented in both compose files and `.env.example`; runbook in `docs/security/malware-scanning.md`.

Design/plan: `docs/superpowers/specs/2026-07-08-malware-scanning-uploads-design.md`, `docs/superpowers/plans/2026-07-08-malware-scanning-uploads.md`.

## Why

For the GCC/government deployment, uploaded files must pass antivirus scanning. In DEV/PRD the antivirus service runs as a sibling ECS service reachable via **ECS Service Connect** — the app points at it through `MALWARE_SCAN_URL`.

## Behaviour / semantics

- `MALWARE_SCAN_URL` **unset** ⇒ scanning **disabled**, upload proceeds (this is how **Vercel/test** runs — it can't reach an internal ECS service).
- Configured + **infected** ⇒ block.
- Configured + **unreachable / timeout / non-200 / malformed response** ⇒ block (**fail-closed**).
- Threat model: browser-parsed CSV/XML/XLSX only ever reached the server as parsed *text*, so scanning them is defense-in-depth; the genuine binary-parse exposure is the PDF path, which is scanned server-side. No new npm dependencies (Node 18+ globals only). No schema changes.

## Test plan

- `npm test` — 813 pass (adds `malwareScan` + `malwareScanClient` unit suites).
- `npm run test:api` — new `scan-file`, `import-timetable-pdf-scan`, `pool-teachers-import-scan` all pass (12/12). Remaining suite failures are the pre-existing red baseline, unchanged by this branch.
- `npm run build` — ✓ 89/89 static pages.
- `npm audit` — no high/critical; no dependency changes.

## Human verification (needs the local ClamAV stack)

Bring up the stack (`docker compose -f docker-compose.local.yml up -d clamav clamav-rest`, wait ~1 min for signatures) and upload an EICAR test file (see `docs/security/malware-scanning.md`) through Timetable Import (CSV/PDF) and the override import → expect the upload blocked with the virus name; a clean file passes. With `MALWARE_SCAN_URL` unset, both pass (disabled).

## Notes / deferred

- **Vercel/test runs with scanning disabled by design** (env var unset). If real scanning is wanted there later, a hosted ClamAV endpoint would be needed.
- Pool-teacher import is on the dormant (flag-OFF) REMS path; scanned anyway since the multipart bytes are in hand.
- Not fixed (low value / documented): `/api/scan-file` returns 500 (not 400) for a non-multipart Content-Type; `app` service does not `depends_on` `clamav-rest` (avoids blocking local dev on the ~90s clamd signature load — runbook documents the window).


src/app/api/scan-file/route.ts
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * POST /api/scan-file — client pre-scan for browser-parsed uploads.
+ * Accepts multipart/form-data with a single `file` field. Auth: SCHOOL_ADMIN
+ * (all browser-parsed import surfaces are school-admin surfaces).
+ */
+export async function POST(request: Request) {
+  try {
+    await requireSchoolAdmin();
+
+    const form = await request.formData();
```
guangshinhaha, line 33: **[Medium] The 5 MB cap is enforced only after the entire request body has been buffered into memory**

```ts
export async function POST(request: Request) {
  try {
    await requireSchoolAdmin();

    const form = await request.formData();   // ← buffers the whole multipart body first
    const file = form.get("file");
    if (!(file instanceof File)) {
      return NextResponse.json({ error: "No file provided." }, { status: 400 });
    }
    const bytes = Buffer.from(await file.arrayBuffer());
    if (bytes.byteLength > MAX_SCAN_BYTES) {  // ← size check runs after the full read
```

**Problem:** `request.formData()` reads the full request body into memory before the `MAX_SCAN_BYTES` check runs, so the check can never bound memory use — an authenticated school admin can POST a multi-hundred-MB body and the process buffers all of it before rejecting. Vercel's ~4.5 MB function body cap masks this on the interim test env, but the GCC ALB/ECS target has no default body limit. `POST /api/import-timetable` already guards exactly this with a declared Content-Length short-circuit (#276 hardening, `route.ts` lines 633–639); this new route skips that guard.

**Suggestion:** reuse the import-timetable pattern before touching the body (keep the post-read `byteLength` check as the fallback for absent/under-reporting headers):
```ts
const declaredLength = Number(request.headers.get("content-length"));
// small margin on top of MAX_SCAN_BYTES to allow for multipart framing overhead
if (Number.isFinite(declaredLength) && declaredLength > MAX_SCAN_BYTES + 16 * 1024) {
  return NextResponse.json({ error: "File exceeds the 5 MB limit." }, { status: 400 });
}
const form = await request.formData();
```

---
*🤖 aif-code-review · claude-fable-5*

src/app/[slug]/dashboard/settings/ImportTimetableTab.tsx
```diff
@@ -479,6 +489,15 @@ export default function ImportTimetableTab({ onImportSuccess, hideImportMode }:
     if (!file) return;
     setParseError(null); setImportResult(null);
     if (file.size > MAX_IMPORT_FILE_BYTES) { setParseError(TOO_LARGE_MESSAGE); return; }
+    try {
+      await scanFileBeforeUpload(file);
```
Comment (guangshinhaha) [inline src/app/[slug]/dashboard/settings/ImportTimetableTab.tsx:494]: **[Medium] Awaiting the pre-scan before reading `oddFile`/`evenFile` introduces a stale-closure race in the odd/even rotation flow**

```tsx
    try {
      await scanFileBeforeUpload(file);      // ← new async gap (full file upload round-trip)
    } catch (err) { /* ... */ }
    const newOdd = week === "odd" ? file : oddFile;    // reads state captured at render time
    const newEven = week === "even" ? file : evenFile;
    if (week === "odd") setOddFile(file);
    if (week === "even") setEvenFile(file);
    if (newOdd && newEven) { /* merged parse */ }
```

**Problem:** before this PR, `newOdd`/`newEven` were computed synchronously in the change handler, so the second file selection always saw the first file in state. Now the handler suspends on a network round-trip (uploading the whole file — seconds on slow links). If the user selects the odd file and picks the even file while the odd scan is still in flight, both invocations hold closures from the same render where `oddFile === evenFile === null`: each computes only its own file as present, both `setOddFile`/`setEvenFile` land, but `if (newOdd && newEven)` is false in **both** — the merged parse never runs and no error is shown. The UI shows both files selected with no preview, until the user re-selects one.

**Suggestion:** keep the handler to validate + scan + store, and move the pairing/parse into an effect that always sees the latest pair:
```tsx
// in handleRotationFile, after the scan:
if (week === "odd") setOddFile(file); else setEvenFile(file);

// parse whenever the pair is complete
useEffect(() => {
  if (!oddFile || !evenFile) return;
  (async () => {
    try {
      const [oddText, evenText] = await Promise.all([readFileAsText(oddFile), readFileAsText(evenFile)]);
      setParsedAsc(mergeParseResults(parseAscXml(oddText, "ODD"), parseAscXml(evenText, "EVEN")));
    } catch {
      setParseError("Failed to parse XML files. Please check the file format.");
      setParsedAsc(null);
    }
  })();
}, [oddFile, evenFile]);
```
(Alternative with a smaller diff: mirror the two files in `useRef`s updated synchronously before the `await`, and read the refs for the pairing check — avoids restructuring but is easier to get subtly wrong.)

---
*🤖 aif-code-review · claude-fable-5*

src/app/api/import-timetable/route.ts
```diff
@@ -403,6 +404,15 @@ async function handlePdfPreview(schoolId: string, body: PdfPreviewBody) {
       const oddBuf = base64ToArrayBuffer(body.oddPdfBase64!);
       const evenBuf = base64ToArrayBuffer(body.evenPdfBase64!);
 
+      try {
+        await assertFileClean(Buffer.from(oddBuf), "timetable-odd.pdf");
```
guangshinhaha: **[Low] Odd and even PDFs are scanned sequentially while the parse right below runs the same pair concurrently**

```ts
      try {
        await assertFileClean(Buffer.from(oddBuf), "timetable-odd.pdf");   // ← serial round-trip 1
        await assertFileClean(Buffer.from(evenBuf), "timetable-even.pdf"); //   serial round-trip 2
      } catch (err) { /* ... */ }

      const [oddResult, evenResult] = await Promise.all([
        parseAscPdf(oddBuf),
        parseAscPdf(evenBuf),
      ]);
```

**Problem:** two independent scanner round-trips run in series on a user-facing request. Each carries the default 30 s `MALWARE_SCAN_TIMEOUT_MS`, so the worst-case wait before the user gets a fail-closed 503 doubles to ~60 s, and even the happy path pays two full sequential uploads of up to 5 MB each to `clamav-rest`.

**Suggestion:**
```ts
try {
  await Promise.all([
    assertFileClean(Buffer.from(oddBuf), "timetable-odd.pdf"),
    assertFileClean(Buffer.from(evenBuf), "timetable-even.pdf"),
  ]);
} catch (err) {
  const scanRes = malwareScanErrorResponse(err);
  if (scanRes) return scanRes;
  throw err;
}
```

---
*🤖 aif-code-review · claude-fable-5*

src/lib/malwareScanClient.ts
```diff
@@ -0,0 +1,52 @@
...
+      "Could not reach the malware scanner. Please try again in a moment.",
+    );
+  }
+
+  if (res.status === 503) {
+    throw new FileRejectedError(
+      "The malware scanner is unavailable. Please try again in a moment.",
+    );
+  }
+  if (!res.ok) {
```
guangshinhaha, line 36: **[Low] Permanent 4xx failures are reported to the user as retryable ("Please try again")**

```ts
  if (res.status === 503) {
    throw new FileRejectedError(
      "The malware scanner is unavailable. Please try again in a moment.",
    );
  }
  if (!res.ok) {                                                        // ← 400 / 401 / 403 all land here
    throw new FileRejectedError("Could not scan the file. Please try again.");
  }
```

**Problem:** the generic `!res.ok` branch covers responses that will never get better by retrying: `400` (file exceeds 5 MB / no file) and `401`/`403` (expired session). Concretely, `TimetableExceptionsSection.handleImport` calls `scanFileBeforeUpload` with **no** client-side size cap (unlike `ImportTimetableTab`, which checks `MAX_IMPORT_FILE_BYTES` first), so a >5 MB override CSV gets "Could not scan the file. Please try again." — and it does so even when scanning is disabled server-side, because `/api/scan-file` runs its 5 MB check *before* the `isMalwareScanConfigured()` check. The user is told to retry an error that is deterministic. (Timetable exceptions is behind the OFF `timetable_exceptions` flag, so this is a pre-enablement paper cut rather than a live bug — but the same message also masks expired-session 401s on the live timetable-import surfaces.)

**Suggestion:** surface the server's own error message for non-503 failures instead of the generic retry text:
```ts
if (!res.ok) {
  let detail: string | undefined;
  try {
    detail = ((await res.json()) as { error?: string }).error;
  } catch {
    // keep the generic message
  }
  throw new FileRejectedError(detail ?? "Could not scan the file. Please try again.");
}
```
(Optionally also add the `MAX_IMPORT_FILE_BYTES` pre-check in `TimetableExceptionsSection.handleImport` for parity with `ImportTimetableTab`.)

---
*🤖 aif-code-review · claude-fable-5*

guangshinhaha:
## Code Review Summary

| Severity | Count |
|----------|-------|
| Critical | 0 |
| High     | 0 |
| Medium   | 2 |
| Low      | 2 |

## What Looks Good
- **Fail-closed semantics are enforced end-to-end and exhaustively**: empty result arrays, malformed rows (missing/non-boolean `is_infected`), non-JSON bodies, non-200s, and timeouts all map to `MalwareScanUnavailableError` → 503 — an ambiguous verdict is never read as clean, and the unit suite pins each of these cases individually.
- **Single-env-var on/off gate with a clean disabled-vs-down distinction** (`isMalwareScanConfigured()` vs `MalwareScanUnavailableError`), read lazily inside functions per the repo's `otpaas.ts` idiom — no config drift, no feature flag to keep in sync.
- **Scanning is placed exactly where raw bytes exist**: client pre-scan for browser-parsed formats, inline server scan for the two genuine server-side byte intakes — including defense-in-depth on the PDF path, the one place untrusted binary actually hits a server parser (`parseAscPdf`).
- **Thoughtful test fixtures**: the minimal valid one-page PDF in `import-timetable-pdf-scan.test.ts` (with a comment explaining why an arbitrary byte string would make the disabled-path assertion fail for the wrong reason) shows the disabled path exercises a real parse, not an accidental early error.

---
*🤖 aif-code-review · claude-fable-5*

chadinwork:
I like the graceful failure in case of no env var.

Looks good to me, please merge and let's give it a try in DEV.

Issue #592, opened 2026-07-30
feat(mobile UI): Make Mobile Summary Metric Cards Interactive Shortcut Filters


J1a3n:
### **User Story**
As a school relief coordinator on a mobile device, I want the summary metric cards (Absent Teachers, Uncovered Periods, Covered Periods) to act as tappable buttons, So that I can instantly jump to or filter the exact list of periods I need to take action on without manually searching for them.

### **Context**
The mobile dashboard header currently features four rounded rectangular metric blocks: Absent Teachers, Uncovered Periods, Covered Periods, and % Coverage Rate.
While these cards visually resemble large buttons, they are currently static text containers. This creates a "broken affordance"—users naturally tap them expecting a reaction, but nothing happens. Making the first three cards interactive will turn key high-level stats into direct action shortcuts.

### **Acceptance Criteria**

**Visual Affordance & Feedback**
- Tappable Styling: Update the UI styling of the actionable metric cards (Absent Teachers, Uncovered Periods, Covered Periods) to visually indicate affordance (e.g., active tap/ripple states, subtle chevron indicators, or pressed-state animations).
- Non-Actionable Card: Keep % Coverage Rate as a static informational display card. Change the display to donut sector percentage chart to distinguish from other tappable buttons.

**Interactive Navigation Shortcuts**
- Tap 'Absent Teachers': Tapping takes the user directly to the list of absent teacher cards 
- Tap 'Uncovered Periods': Tapping takes user to the "period to fill" panel and turn on **uncovered** filter. #591 
- Tap 'Covered Periods': Tapping filters takes user to the "period to fill" panel and turn on **covered** filter. #591 

### **Out of Scope**
Tappable drill-down actions on the % Coverage Rate card (remains read-only for now).

### **UX / Design Notes**
Touch Target: Ensure the entire surface area of each rounded rectangle is a valid touch target (minimum 44x44 pt).
Feedback: Add instant feedback (<100ms response time) via visual press-states so the user knows their tap was registered.

guangshinhaha:
## Implemented on `claude/core-review-ai-y2o9ow` + code review

Branch: [`claude/core-review-ai-y2o9ow`](https://github.com/String-dxd/reliefcher/tree/claude/core-review-ai-y2o9ow) · commits `247eca8` (feature) and `ecb9516` (review fixes).

### Scope note — #591 shipped as a prerequisite

[...]

Issue #591, opened 2026-07-30
feat(mobile UI) Add Display Filter (All / Uncovered / Covered) to 'Periods to Fill' View


J1a3n:
### **User Story**

As a school relief coordinator, I want to toggle the "Periods to Fill" view between All, Uncovered, and Covered periods, So that I can quickly hide already-assigned slots and focus strictly on periods that still urgently need relief teachers.

### **Context**

Currently, the "Periods to Fill" list displays all absent periods mixed together regardless of their fulfillment status. During high-volume absence days, coordinators get overwhelmed scrolling past periods that have already been resolved. Adding a prominent, single-tap filter bar allows users to isolate actionable items instantly.

### **Acceptance Criteria**

 **Filter Control UI**
- Segmented Control / Filter Bar: Place a clean 3-option toggle control at the top of the "Periods to Fill" section:
[ Uncovered ] (Default active state)
[ Covered ]
[ All ]
- Badge Counts: Display live item counts inside each filter chip/button (e.g., Uncovered (6), Covered (12), All (18)).

**Filtering Behavior**
- Uncovered Filter: Shows only absent periods that do not currently have a relief teacher assigned.
- Covered Filter: Shows only absent periods where a relief teacher has already been successfully allocated.
- All Filter: Displays the full list of absent periods (both covered and uncovered) grouped chronologically.
- Real-Time Update: When a user assigns a relief teacher to an uncovered period while in the Uncovered tab, that period should gracefully animate/transition out of the list, updating the badge counts automatically.

**Empty States**
- Display clear, friendly empty state messages for each filter view when zero items match:
Uncovered (0): "🎉 All absent periods are covered for today!"
Covered (0): "No periods have been assigned relief yet."

### **Out of Scope**
Advanced multi-level filtering (e.g., filtering by specific grade, subject, or department simultaneously—can be a future enhancement).
Persisting filter selection across user sessions (defaults back to Uncovered on fresh page load).

### **UX / Design Notes**
Default View: Set Uncovered as the default active tab when navigating to this view, as it represents the actionable work list.
Touch Optimization: Filter segments should span full width on mobile viewports for easy thumb-tapping.

PR #589, opened 2026-07-30
feat(dashboard): mobile dashboard redesign (Ticket #583)

J1a3n:
Resolves #583

### **Description**
This PR implements a mobile-optimized redesign for the ReliefCher coordinator dashboard, focusing on maximizing vertical real estate and providing a clean, accordion-style user interface with scrollable candidate sub-panels.

A detailed walkthrough of the layout choices, file edits, and verification results can be found in the newly added documentation:
👉 [docs/walkthrough-583-mobile-dashboard-redesign.md](docs/walkthrough-583-mobile-dashboard-redesign.md)

### **Key Changes Included**
1. **Collapsible Absent Teacher Cards (Accordion)**: Cards are collapsed by default. Tapping a card expands it and automatically collapses all other open cards to save space.
2. **Trimmed Header Metadata**: Removed extra metadata from the card header in both states, keeping the header dimensions identical and compact.
3. **Two-Row Status Pills**: Period status pills are grouped into covered (top) and uncovered (bottom) rows. The `P` prefix automatically hides on screens below 380px.
4. **Horizontal Scrollable Period Pills**: Under the expanded view, period cards are trimmed of secondary details and placed on a single horizontally scrollable row.
5. **Anchored Viewport Scrolling**: Candidate lists are wrapped in maximum-height scrollable containers, keeping period details anchored on-screen while scrolling.

### **Verification**
- **Unit Tests**: All 837 tests passed.
- **Production Build**: Successfully built production bundle without errors.

Issue #583, opened 2026-07-27
(UX/UI) Mobile Dashboard Redesign: Compact Cards, Collapsible Sections & Scrollable Sub-Tabs


J1a3n:
### **User Story**

As a school relief coordinator using a mobile device, I want a compact, touch-optimized dashboard with inline card expanding and sub-tab navigation, So that I can manage relief assignments quickly without excessive vertical scrolling or losing my context on screen.

### **Context & Problem Statement**

In the current mobile view absent teacher cards and period blocks take up too much vertical screen real estate, forcing heavy scrolling. This redesign focuses on high information density, touch-friendly interactions, and keeping actions contained within the current viewport.

### **Acceptance Criteria**

**Compact Absent Period Cards**

- Reduced Vertical Footprint: Redesign period/teacher cards into a condensed row format showing key details at a glance (e.g., Teacher Name, Number of Unassigned Periods, Urgent/Status Indicator).
- Information Hierarchy: Prioritize critical status text while hiding tertiary metadata until expanded.

**Tap-to-Expand / Tap-to-Collapse Interaction**

- Accordion Behavior: Tapping a teacher/absent card expands it inline to reveal details (e.g., specific periods needing relief, assigned status, quick actions).
- In-Screen Toggle: Tapping the card again (or a clear collapse icon) closes it. Currently only the absent period cards are toggle-able.

 **Sub-Tab Scrolling**

- scrollable relief teacher sub-tab while the absent teacher and period to cover remains on screen as anchor.

### **UX / Design Notes**

Target Viewports: Optimized for mobile screen widths between 360px and 430px.
Touch Targets: Ensure all tappable card headers and collapse triggers meet minimum touch target guidelines (at least 44x44 pt).
Visual Cues: Include subtle visual indicators (chevrons ^ / v or expand icons) so users immediately know cards are interactive.

J1a3n:
Hi @J1a3n,

I have completed the implementation of **Ticket #583 (Mobile Dashboard Redesign)**. Below is a walkthrough of the changes:

### 1. Collapsible Absent Teacher Cards with Accordion Behavior
- **File**: `AbsentTeachersTab.tsx`
- **Logic**:
  - Cards are collapsed by default, drastically reducing vertical scrolling.
  - **Accordion Behavior**: Expanding a card automatically collapses any other open cards to keep the view focused and clean.
[...]

PR #593, opened 2026-07-30
feat(dashboard): display filters for periods to fill and interactive metric shortcuts (#591, #592)

J1a3n:
# Walkthrough: Mobile Dashboard Display Filters & Interactive Metric Shortcuts (#591, #592)

Implemented display filter controls for the "Periods to Fill" view and transformed dashboard summary metric cards into interactive action shortcuts.

---

### 1. Display Filter Bar for 'Periods to Fill' View (#591)
- **File Modified**: [PeriodsToFillTab.tsx](file:///c:/Users/hejia/Documents/Antigravity/Reliefcher/src/app/[slug]/dashboard/PeriodsToFillTab.tsx)
- Added a full-width segmented filter bar (`Uncovered`, `Covered`, `All`) with live item count badges.
- Set `Uncovered` as the default active tab.
- Added dynamic filtering for the period list in real time as relief assignments change.
- Provided custom empty states per filter view:
  - `Uncovered (0)`: *"🎉 All absent periods are covered for today!"*
  - `Covered (0)`: *"No periods have been assigned relief yet."*
  - `All (0)`: *"No absent periods recorded for today."*

### 2. Interactive Metric Card Shortcuts (#592)
- **File Modified**: [ReliefOverview.tsx](file:///c:/Users/hejia/Documents/Antigravity/Reliefcher/src/app/[slug]/dashboard/ReliefOverview.tsx)
- Refactored summary cards into interactive shortcut buttons with hover borders, active press scaling (`active:scale-[0.98]`), and right chevrons (`ChevronRight`):
  - **Absent Teachers Card**: Switches to `Absent Teachers` tab and scrolls down to the list.
  - **Uncovered Periods Card**: Switches to `Periods to Fill` tab, sets filter state to `Uncovered`, and scrolls down.
  - **Covered Periods Card**: Switches to `Periods to Fill` tab, sets filter state to `Covered`, and scrolls down.
- **Coverage Rate Visual Differentiation**:
  - Removed background card block (`border`, `bg-primary/10`, `shadow-sm`) from `% Coverage Rate`.
  - Formatted as a clean readout containing only 3 elements: SVG Donut Sector Chart, % percentage number, and text label.

---

### 1. Automated Tests
- Ran `npm test -- --run`: All **837 tests passed** across 82 test files.
- Ran `npx tsc --noEmit`: **0 TypeScript errors**.

### 2. Production Build Check
- Ran `npm run build`: Production build compiled and generated static pages successfully.


src/app/[slug]/dashboard/AbsentTeachersTab.tsx
```diff
@@ -338,69 +343,116 @@ export default function AbsentTeachersTab({
...
+                            className="inline-flex items-center justify-center rounded px-1.5 py-0.5 text-[10px] font-bold border border-destructive/20 bg-destructive/10 text-destructive shrink-0"
+                          >
+                            <span className="hidden min-[380px]:inline">P</span>
+                            {slot.periodNumber}
+                          </span>
+                        ))}
+                    </div>
+                  )}
+                </div>
+                {isCardExpanded && (
```
Comment (guangshinhaha) [inline src/app/[slug]/dashboard/AbsentTeachersTab.tsx]: **[High] Branch predates #589 — merging this reverts shipped #583 desktop behaviour**

The merge base is `7a78303` (PR #586), not current `main` (`9c9ef28`), so this branch never saw #583 land. `git merge-tree` against `main` produces **6 conflict regions**, and the conflicting content is #583's desktop rules.

```tsx
{isCardExpanded && (                                            // ←
  <button
    onClick={(e) => { e.stopPropagation(); handleDeleteReport(...) }}
```

**Problem:** `main` currently renders the delete button unconditionally with `${isCardExpanded ? "" : "hidden md:inline-block"}` — deliberately, because the card is never collapsed on md+. This branch gates it behind `isCardExpanded`, so a desktop admin loses the delete control entirely until they expand the card. The same revert hits three more places in this file:

- L350 — `md:cursor-default md:select-auto md:hover:bg-transparent` dropped, making desktop headers clickable/collapsible again
- L453 — period pills go from `${isCardExpanded ? "block" : "hidden"} md:block` back to `{isCardExpanded && …}`, so pills vanish on desktop unless expanded (undoing "always shown on desktop so the card keeps full density")
- L469 — `snap-start` reverted to `snap-align-start`, which is not a Tailwind utility and does nothing

None of this belongs to #591/#592 — it's just stale history riding along.

**Suggestion:** rebase onto current `main` and drop every `AbsentTeachersTab.tsx` / `PreferredTeachersPanel.tsx` hunk and the `docs/walkthrough-583-*.md` file from this branch. They already shipped in #589.

```bash
git fetch origin main
git rebase origin/main   # resolve in favour of main for the #583 files
```

Worth doing before anything else — the review below is on the #591/#592 code, which is a much smaller surface once this noise is gone.

---

---

src/app/[slug]/dashboard/ReliefOverview.tsx
```diff
@@ -356,37 +364,95 @@ export default function ReliefOverview({
...
       <div className="grid grid-cols-2 gap-3 md:grid-cols-4">
-        <div className="rounded-xl border border-border bg-destructive/10 p-4 shadow-sm">
-          <p className="text-xs font-medium uppercase tracking-wide text-destructive">
-            Absent Teachers
-          </p>
+        <button
+          type="button"
+          onClick={() => {
+            setActiveTab("absent");
+            scrollToTabSection();
```
Comment (guangshinhaha) [inline src/app/[slug]/dashboard/ReliefOverview.tsx:376]: **[Medium] Metric cards are tappable when there is nothing to scroll to — the same dead tap #592 set out to remove**

```tsx
onClick={() => {
  setActiveTab("absent");
  scrollToTabSection();   // ← tabContentRef.current is null here in two real states
}}
```

**Problem:** `tabContentRef` is attached to the tab section, which renders only under `{!isLoading && sickTeachers.length > 0 && (…)}` (L601). So in two reachable states the ref is null and `scrollToTabSection` silently no-ops while the button still reports a press:

1. **During load** — the cards render immediately showing `-`, the tab section does not exist yet. Tapping does nothing.
2. **A day with no absences** — cards show `0`, the "All clear for today" empty state is shown instead of the tab section. Tapping does nothing.

That is precisely the broken affordance the ticket describes ("users naturally tap them expecting a reaction, but nothing happens"), just relocated from the static card to the loading/empty states.

**Suggestion:** disable the shortcuts when their target does not exist, so the press state never fires on a dead tap.

```tsx
<button
  type="button"
  disabled={isLoading || totalAbsent === 0}
  onClick={() => { setActiveTab("absent"); scrollToTabSection(); }}
  className="group … disabled:pointer-events-none disabled:opacity-70"
>
```

Applies to all three shortcut cards.

---

---

src/app/[slug]/dashboard/ReliefOverview.tsx
```diff
@@ -97,9 +97,17 @@ export default function ReliefOverview({
   const [notifyPanelOpen, setNotifyPanelOpen] = useState(false);
   const [autoAssignLoading, setAutoAssignLoading] = useState(false);
   const [activeTab, setActiveTab] = useState<"absent" | "periods">("absent");
+  const [periodsFilterStatus, setPeriodsFilterStatus] = useState<"uncovered" | "covered" | "all">("uncovered");
```
Comment (guangshinhaha) [inline src/app/[slug]/dashboard/ReliefOverview.tsx:105]: **[Medium] Filter is never reset on date change — the work list can stay hidden after navigating days**

```tsx
const [periodsFilterStatus, setPeriodsFilterStatus] = useState<"uncovered" | "covered" | "all">("uncovered");  // ←
```

`handleDateChange` (L~150) resets `previewData` and `hintDismissed`, but not this.

**Problem:** coordinator taps **Covered Periods** → filter is `covered` → they move to tomorrow via the week strip. The filter is still `covered`. Tomorrow has no assignments yet, so the Periods tab shows *"No periods have been assigned relief yet."* while every uncovered period for that day sits behind a segment they aren't looking at. The one screen that exists to show outstanding work reads as if there is none.

#591 is explicit that `Uncovered` is the default because it is the actionable list, and that selection is not persisted. Persisting it across a date change is the same failure as persisting it across a session.

**Suggestion:**

```tsx
function handleDateChange(newDate: string) {
  hasDataRef.current = false;
  setDate(newDate);
  setPreviewData(null);
  setHintDismissed(false);
  setPeriodsFilterStatus("uncovered");   // a new day is a new work list
  window.history.replaceState(null, "", `${basePath}?date=${newDate}`);
}
```

---

---

src/app/[slug]/dashboard/PeriodsToFillTab.tsx
```diff
@@ -61,6 +65,13 @@ export default function PeriodsToFillTab({
   const [removingCoverageFor, setRemovingCoverageFor] = useState<string | null>(null);
   const [tagFilter, setTagFilter] = useState("");
   const [tagOptions, setTagOptions] = useState<string[]>([]);
+  const [localFilterStatus, setLocalFilterStatus] = useState<"uncovered" | "covered" | "all">("uncovered");
+
+  const filterStatus = activeFilterStatus ?? localFilterStatus;
+  const setFilterStatus = (status: "uncovered" | "covered" | "all") => {
```
Comment (guangshinhaha) [inline src/app/[slug]/dashboard/PeriodsToFillTab.tsx:71]: **[Medium] Filter change leaves the picker open, so the shortcut never reaches the list on mobile**

```tsx
const setFilterStatus = (status: "uncovered" | "covered" | "all") => {
  setLocalFilterStatus(status);
  onFilterStatusChange?.(status);      // ← selectedPeriod is left untouched
};
```

**Problem:** the period list lives in the left column, which is `hidden md:block` while `selectedPeriod` is set (L~372). Flow on mobile: coordinator taps a period → picker opens, list hidden → scrolls up → taps **Uncovered Periods**. The segment highlight flips (good, the bar is outside the hidden column) but the list never appears — they are still looking at the picker for the period they opened. #592's criterion is "takes user to the *period to fill* panel", and they don't get there.

Desktop has a milder version of the same thing: filter to `Covered` while an uncovered period is open and the picker stays mounted on a period that is no longer in the list beside it.

**Suggestion:** switching the view should return to the list — the selected period may not even belong to the new bucket.

```tsx
useEffect(() => {
  setSelectedPeriod(null);
}, [filterStatus]);
```

---

---

src/app/[slug]/dashboard/PeriodsToFillTab.tsx
```diff
@@ -296,23 +286,97 @@ export default function PeriodsToFillTab({
...
+    }
+    return null;
+  };
 
-        <div className="space-y-4">
-          {periodsByTeacher.map(({ teacherName, absenceReportId, periods }) => (
-            <div key={absenceReportId}>
+  return (
+    <div className="space-y-4">
+      {/* Segmented Filter Control Bar */}
```
Comment (guangshinhaha) [inline src/app/[slug]/dashboard/PeriodsToFillTab.tsx:334]: **[Medium] Segmented control drops the accessible-tabs pattern this same file already uses**

```tsx
{/* Segmented Filter Control Bar */}
<div className="flex w-full rounded-xl border border-border bg-muted/50 p-1">   // ←
  <button type="button" onClick={() => setFilterStatus("uncovered")} …>
    Uncovered ({uncoveredPeriods.length})
  </button>
```

**Problem:** three bare buttons — no `role="tablist"` / `role="tab"`, no `aria-selected`, no `aria-controls`, no roving `tabIndex`, no arrow-key navigation. A screen-reader user hears three buttons with no indication which view is active; the active state is conveyed by background colour alone.

What makes this worth fixing rather than deferring: the assignment-source segmented control ~90 lines below in this same component already implements all of it (`role="tablist"`, `aria-selected`, `aria-controls`, `tabIndex={isActive ? 0 : -1}`, ArrowLeft/ArrowRight). Two segmented controls on one screen behaving differently is the maintenance cost here, not the a11y gap alone.

**Suggestion:** mirror the existing pattern rather than writing a second idiom.

```tsx
<div role="tablist" aria-label="Filter periods by coverage"
     className="flex w-full rounded-xl border border-border bg-muted/50 p-1">
  {(["uncovered", "covered", "all"] as const).map((status, i) => {
    const isActive = filterStatus === status;
    return (
      <button
        key={status}
        role="tab"
        type="button"
        aria-selected={isActive}
        aria-controls="ptf-period-list"
        tabIndex={isActive ? 0 : -1}
        onClick={() => setFilterStatus(status)}
        className={`flex-1 rounded-lg py-2 text-xs font-semibold …`}
      >
        …
      </button>
    );
  })}
</div>
```

…with the list container carrying `id="ptf-period-list" role="tabpanel"`.

---

---

src/app/[slug]/dashboard/PeriodsToFillTab.tsx
```diff
@@ -61,6 +65,13 @@ export default function PeriodsToFillTab({
   const [removingCoverageFor, setRemovingCoverageFor] = useState<string | null>(null);
   const [tagFilter, setTagFilter] = useState("");
   const [tagOptions, setTagOptions] = useState<string[]>([]);
+  const [localFilterStatus, setLocalFilterStatus] = useState<"uncovered" | "covered" | "all">("uncovered");
+
+  const filterStatus = activeFilterStatus ?? localFilterStatus;
```
Comment (guangshinhaha) [inline src/app/[slug]/dashboard/PeriodsToFillTab.tsx:70]: **[Low] Two sources of truth for one filter value**

```tsx
const [localFilterStatus, setLocalFilterStatus] = useState<…>("uncovered");
const filterStatus = activeFilterStatus ?? localFilterStatus;    // ←
const setFilterStatus = (status) => {
  setLocalFilterStatus(status);
  onFilterStatusChange?.(status);
};
```

**Problem:** the only caller (`ReliefOverview`, L640–643) always passes both props, so `activeFilterStatus` always wins and `localFilterStatus` is written but never read. It's dead state that looks live. The failure mode if someone later mounts this tab without the props: the filter works, but the date-change reset (which lives in the parent) silently stops applying, and the bug is invisible because the component still "has" a filter.

**Suggestion:** make it plainly controlled — the parent owns it because the metric-card shortcuts need to set it.

```tsx
type Props = {
  …
  filterStatus: "uncovered" | "covered" | "all";
  onFilterStatusChange: (status: "uncovered" | "covered" | "all") => void;
};
```

Dropping `localFilterStatus` and the `??` removes ~6 lines and one class of desync.

> If you'd rather keep the tab standalone-capable, the other direction works too: make it fully uncontrolled and expose an imperative handle. But half-and-half is the option with no upside.

---

---

src/app/[slug]/dashboard/ReliefOverview.tsx
```diff
@@ -356,37 +364,95 @@ export default function ReliefOverview({
...
+                <path
+                  className="text-muted/60"
+                  strokeWidth="4"
+                  stroke="currentColor"
+                  fill="none"
+                  d="M18 2.0845 a 15.9155 15.9155 0 0 1 0 31.831 a 15.9155 15.9155 0 0 1 0 -31.831"
+                />
+                <path
+                  className="text-primary transition-all duration-500 ease-out"
+                  strokeDasharray={`${isLoading ? 0 : coverageRate}, 100`}
```
Comment (guangshinhaha) [inline src/app/[slug]/dashboard/ReliefOverview.tsx:451]: **[Low] Donut shows a coloured mark at 0% coverage**

```tsx
<path
  className="text-primary transition-all duration-500 ease-out"
  strokeDasharray={`${isLoading ? 0 : coverageRate}, 100`}   // ←
  strokeWidth="4"
  strokeLinecap="round"
  stroke="currentColor"
  fill="none"
  d="M18 2.0845 a 15.9155 15.9155 0 0 1 0 31.831 a 15.9155 15.9155 0 0 1 0 -31.831"
/>
```

**Problem:** with `coverageRate === 0` the dash length is 0, but `strokeLinecap="round"` renders a round cap on a zero-length dash — Chrome and Firefox both paint a filled dot of stroke-width diameter at the top of the ring. On a day where nothing is covered, the chart shows a primary-coloured mark, reading as "a little bit covered" next to a `0%` label. Same at load, where `isLoading` forces `0`.

(The `15.9155` radius is a nice touch — circumference lands on 100.0003, so the `, 100` denominator is right.)

**Suggestion:** drop the round cap when there is nothing to draw.

```tsx
strokeLinecap={coverageRate > 0 ? "round" : "butt"}
```

Or skip rendering the progress path entirely when `isLoading || coverageRate === 0`.

---

---

src/app/[slug]/dashboard/PeriodsToFillTab.tsx
```diff
@@ -296,23 +286,97 @@ export default function PeriodsToFillTab({
...
+            <UserX className="h-6 w-6 text-muted-foreground" />
+          </div>
+          <p className="text-sm font-medium text-foreground">No periods have been assigned relief yet.</p>
+          <p className="mt-1 text-xs text-muted-foreground">
+            Select an uncovered period to assign a relief teacher.
+          </p>
+        </div>
+      );
+    }
+    return null;
```
Comment (guangshinhaha) [inline src/app/[slug]/dashboard/PeriodsToFillTab.tsx:329]: **[Low] Unreachable `return null`, and a "(0) (0) (0)" filter bar over the no-periods state**

```tsx
if (filterStatus === "covered" && coveredPeriods.length === 0) {
  return ( … );
}
return null;                                   // ←
```

**Problem (a):** `renderEmptyState()` is only called when `displayPeriods.length === 0`. Enumerating the cases: `allPeriods.length === 0` hits the first branch; `uncovered` empty hits the second; `covered` empty hits the third; and `all` empty implies `allPeriods.length === 0`, which is already the first branch. So `return null` cannot be reached — but it reads like a real fallback, and the next person will treat it as one.

**Problem (b):** on a day with no periods at all, the segmented bar now renders `Uncovered (0) Covered (0) All (0)` above the "No periods to show" card. Previously that state early-returned a clean card. Low impact given `ReliefOverview` only mounts the tabs when `sickTeachers.length > 0`, but it is a visible oddity when it does occur.

**Suggestion:** keep the whole-day-empty case as an early return above the filter bar, so `renderEmptyState` only handles the two cases a filter can produce:

```tsx
if (allPeriods.length === 0) {
  return ( /* the existing "No periods to show" card, full width, no filter bar */ );
}

const renderEmptyState = () => {
  if (filterStatus === "uncovered") return ( /* 🎉 all covered */ );
  return ( /* nothing assigned yet */ );   // only "covered" can reach here
};
```

---

---

guangshinhaha (commented):
## Code Review Summary

| Severity | Count |
|----------|-------|
| Critical | 0 |
| High     | 1 |
| Medium   | 4 |
| Low      | 3 |

The one to handle first is structural rather than a code defect: this branch's merge base is `7a78303`, so it predates #589 and carries an older `AbsentTeachersTab.tsx` / `PreferredTeachersPanel.tsx`. `git merge-tree` against `main` yields **6 conflict regions**, and the conflicting content is #583's desktop rules — the always-visible delete button, the always-expanded period pills, `md:cursor-default`, and the `snap-start` fix. Rebase onto current `main` and drop those files from the branch; everything below is on the #591/#592 code, which is a much smaller surface once that noise is gone.

Of the Medium findings, two are worth calling out together because they undercut the ticket's own premise: the shortcut cards stay tappable while loading and on zero-absence days when there is nothing to scroll to, and changing the filter leaves the picker open so on mobile the shortcut never actually reaches the list. #592 exists to remove dead taps, so it's worth making sure the new ones don't reintroduce them in different states.

## What Looks Good

- Moving the segmented bar **outside** the `md:flex` two-column wrapper is the right call — it keeps the filter reachable on mobile while a period is selected, which is easy to get wrong by nesting it in the left column.
- The `15.9155` radius on the donut path is deliberate: circumference lands on 100.0003, so `strokeDasharray={`${coverageRate}, 100`}` needs no conversion. Nice.
- Bumping `md:h-[calc(100dvh-16rem)]` to `20rem` to pay for the filter bar's height, rather than letting the list overflow, shows the layout was actually checked rather than assumed.
- Per-filter empty-state copy matches the ticket's wording exactly, including the `🎉`.

---

---

src/app/[slug]/dashboard/ReliefOverview.tsx
```diff
@@ -356,37 +364,95 @@ export default function ReliefOverview({
...
-          <p className="mt-1 font-display text-2xl font-bold text-primary">
-            {isLoading ? "-" : `${coverageRate}%`}
-          </p>
+          <div className="mt-1 flex items-center justify-between">
+            <p className="font-display text-2xl font-bold text-foreground">
+              {isLoading ? "-" : `${coverageRate}%`}
+            </p>
+            {/* SVG Donut sector chart */}
+            <div className="relative flex h-9 w-9 items-center justify-center">
+              <svg className="h-full w-full -rotate-90" viewBox="0 0 36 36">
```
Comment (guangshinhaha) [inline src/app/[slug]/dashboard/ReliefOverview.tsx]: **[Low] `-rotate-90` makes the coverage donut start filling at 9 o'clock, not 12**

```tsx
<svg className="h-full w-full -rotate-90" viewBox="0 0 36 36">   // ←
  <path
    className="text-primary …"
    strokeDasharray={`${isLoading ? 0 : coverageRate}, 100`}
    d="M18 2.0845 a 15.9155 15.9155 0 0 1 0 31.831 a 15.9155 15.9155 0 0 1 0 -31.831"
  />
</svg>
```

**Problem:** `-rotate-90` is the idiom that goes with `<circle cx="18" cy="18" r="15.9155">`, whose stroke starts at 3 o'clock and needs rotating back to the top. This is a `<path>`, and it already starts at `M18 2.0845` — i.e. top-centre — sweeping clockwise (`sweep-flag 1`). The extra `-rotate-90` therefore moves the origin *off* 12 o'clock to 9 o'clock.

Concrete result: at `coverageRate === 25`, the filled quarter sits across the **left** side of the ring (9→12 o'clock) instead of the top-right quadrant (12→3). Every partial value reads as rotated a quarter-turn, and combined with the round cap it puts the 0% dot on the left edge rather than the top.

**Suggestion:** the path needs no rotation at all.

```tsx
<svg className="h-full w-full" viewBox="0 0 36 36">
```

---

src/app/[slug]/dashboard/PeriodsToFillTab.tsx
```diff
@@ -255,38 +266,17 @@ export default function PeriodsToFillTab({
 
   const referencePeriods = selectedTeachers.find((t) => t.daySchedule?.length)?.daySchedule ?? [];
```
Comment (guangshinhaha) [inline src/app/[slug]/dashboard/PeriodsToFillTab.tsx:268]: **[Low] `Badge` is now an unused import — its only call site was deleted with the "Periods" heading** (flagged here; the import itself is L12, outside the diff)

```tsx
import { Button, Badge } from "@flow/core";   // ← Badge no longer used anywhere in this file
```

**Problem:** the one `Badge` usage in this component was the uncovered count in the removed left-column heading:

```tsx
-        <div className="mb-3 flex items-center gap-2">
-          <h2 className="text-sm font-semibold text-foreground">Periods</h2>
-          <Badge variant="critical">{uncoveredPeriods.length}</Badge>
```

`git grep -n Badge` on this file at head returns only the import line. `eslint-config-next/typescript` has `@typescript-eslint/no-unused-vars` on, so this warns on every `next lint` run and ships a dead symbol into the module graph. It won't fail the build (it's a warning, not an error), which is exactly why it will sit there.

**Suggestion:**

```tsx
import { Button } from "@flow/core";
```

---

guangshinhaha:
## Code Review Summary

Re-review at `c602653`. **The branch has not moved since the previous pass** — merge base is still `7a78303`, and all 8 open threads from that review are still unaddressed, so nothing was resolved. Only 2 genuinely new findings this round.

| Severity | Count (new this pass) |
|----------|-------|
| Critical | 0 |
| High     | 0 |
| Medium   | 0 |
| Low      | 2 |

**Still open from the previous pass** — 1 High, 4 Medium, 3 Low. The High one gates the rest: the branch predates #589, so merging as-is reverts shipped #583 desktop behaviour in `AbsentTeachersTab.tsx`. `git diff origin/main...HEAD` confirms it — the delete button goes back behind `isCardExpanded`, the `md:cursor-default` / `md:block` period-pill rules are dropped, and `snap-start` reverts to the non-existent `snap-align-start`. `PreferredTeachersPanel.tsx` and `docs/walkthrough-583-*.md` are now **byte-identical to `main`**, so those two are pure merge-base noise and will drop out on a rebase. Rebasing onto `main` should shrink this PR to just the `PeriodsToFillTab.tsx` + `ReliefOverview.tsx` work.

## What Looks Good

- **The filter is lifted to `ReliefOverview` rather than reached into via a ref.** The metric-card shortcut needs to set the tab's filter, and `activeFilterStatus` / `onFilterStatusChange` is the right shape for that — parent owns the cross-component state, child stays declarative. (The leftover `localFilterStatus` fallback is the only thing muddying it — see the existing thread.)
- **Per-filter empty states instead of one generic "nothing here".** `Uncovered (0)` reading "🎉 All absent periods are covered" and `Covered (0)` reading "No periods have been assigned relief yet" carry genuinely different meaning for a coordinator, and the copy distinguishes them correctly.
- **`15.9155` as the donut radius** so the circumference lands on 100.0003 — makes `strokeDasharray={`${coverageRate}, 100`}` exact with no magic multiplier.
- **`md:h-[calc(100dvh-20rem)]` was bumped from `16rem`** to pay for the new filter bar, rather than left to overflow. Easy thing to forget.

---

guangshinhaha:
## Rebased onto `main` — merge-base issue resolved

This branch forked at `7a78303`, **before** #589 landed, and carried its own 6-commit copy of the #583 mobile dashboard redesign. Since `main` already ships #583 (via `9c9ef28`), merging as-was would have reverted the shipped desktop behaviour in `AbsentTeachersTab.tsx`.

Replaying all 8 commits would have re-applied that stale copy, so the rebase kept only the #591/#592 work and dropped the duplicated #583 commits — `main`'s versions win.

**Before → after**

| | Before | After |
|---|---|---|
| Commits | 8 | 2 |
| Files changed | 5 | 2 |
| Diff | +516 / −269 | +202 / −70 |

Dropped as already-in-`main`: `AbsentTeachersTab.tsx`, `PreferredTeachersPanel.tsx`, `docs/walkthrough-583-mobile-dashboard-redesign.md`.

Also picked up from `main` and verified intact after the rebase:
- #589's shipped `AbsentTeachersTab.tsx` — now byte-identical to `main`, so the delete-button re-gating and the `snap-align-start` typo are gone.
- #588's `WEEKLY` → `WINDOW` cap-mode rename in both remaining files.

**Verification:** `npm run build` clean (`✓ Generating static pages (93/93)`, no `⨯`); `npm test` 82 files / 837 passed. The failing Vercel check is a plan limitation (Hobby plan cannot deploy a private org repo), not a code failure.

**Still open:** the 6 remaining review threads on `PeriodsToFillTab.tsx` / `ReliefOverview.tsx` are unaffected by the rebase and still need attention — notably the dead taps on loading/empty metric cards, the filter not resetting on date change, and the filter change leaving the mobile picker open.

---

PR #594, opened 2026-07-30
feat(teacher role): Visual Timetable View & Quick Access for Teachers (#567)

### LINKED ISSUE #567: feat(teacher role) Visual Timetable View & Quick Access for Teachers
### **User story:**

As a teacher moving between classrooms, I want to view my daily/weekly schedule in a clean, visually distinct, and easily accessible digital format, So that I can instantly see where I need to be without relying on a tiny physical lanyard card or searching through my photos.

### **Background:**

Currently, timetable data exists in the system, but teachers lack a fast, high-visibility mobile view tailored for on-the-go movement. Relying on physical prints or gallery screenshots creates friction between class transitions. The primary goal of this view is glanceability and speed.

This feature could potentially improved adoption with teachers.

### **Acceptance Criteria:**

**Glanceable visual** 
•	Prominent visual showing current and upcoming classes of today
•	Colour coded by class
•	High contrast, big font to improve readability

**Quick Launch and Persistency**
•	Display directly upon log-in or accessible to one tap action
•	Locally cached to reduce impact of connectivity issues

**Data Integration**
•	Assigned relief duties are dynamically integrated into the base timetable

### **Out of Scope**
•	Customised colour coding by the teacher.
•	Interactive editing of timetable

J1a3n:
Closes #567

### Summary of Changes:
- **Sticky Countdown Bar**: Always-visible slim sticky bar with client-side live minute ticker (tracks lesson remaining time & break duration progress).
- **Collapsed Past Lessons**: Past/finished lesson cards automatically collapse into compact rows showing only time and class.
- **Omission of Free & Break Periods**: Unassigned free periods and recess/break slots are hidden from mobile list view.
- **Persistent Color Hashing**: Deterministic color swatches derived per user from \userId + className + subject\.
- **Light Theme**: High contrast typography and separate lines for Class & Subject.
- **Mobile Only Scope**: Scoped to \md:hidden\ with 0 impact on desktop or backend models.


src/app/[slug]/dashboard/AbsentTeachersTab.tsx
```diff
@@ -338,69 +343,116 @@ export default function AbsentTeachersTab({
...
                   )}
-                </button>
+                </div>
               </div>
             </div>
 
             {/* Period pills */}
-            <div className="px-4 py-3 md:px-5">
-              <div className="flex flex-wrap gap-2">
+            {isCardExpanded && (
```
Comment (guangshinhaha) [inline src/app/[slug]/dashboard/AbsentTeachersTab.tsx]: **[High] This branch was cut before #589 merged and re-implements the same files — merging it reverts the desktop guards that landed on `main`.**

The PR is currently `mergeable: CONFLICTING` / `mergeStateStatus: DIRTY`. Diffing this head against current `main` shows the accordion work here is an *older* variant of what #589 already shipped:

```tsx
// main (post-#589):
{/* Period pills — collapsed behind the accordion on mobile, but
    always shown on desktop so the card keeps full density. */}
<div className={`px-4 py-3 md:px-5 ${isCardExpanded ? "block" : "hidden"} md:block`}>
  <div className="flex md:flex-wrap gap-2 overflow-x-auto pb-2 md:pb-0 snap-x snap-mandatory">

// this branch:
{isCardExpanded && (                                    // ←  no md:block escape hatch
  <div className="px-4 py-3 md:px-5">
  <div className="flex md:flex-wrap gap-2 ... scrollbar-thin snap-x snap-mandatory">
```

**Problem:** Four fixes from #589 are undone if this lands as-is:

1. **Period pills vanish on desktop.** `main` keeps `md:block` so the pills are always rendered on md+; here they are gated on `isCardExpanded` at every breakpoint. A coordinator on a laptop opens the dashboard and every absent-teacher card is an empty header — they must click each one to see which periods need filling.
2. **Delete-report button vanishes on desktop.** `main` uses `className={... ${isCardExpanded ? "" : "hidden md:inline-block"}}`; here the button is `{isCardExpanded && (<button .../>)}`, so on desktop the only way to delete an absence report is to first expand the card.
3. **Header becomes a click target on desktop.** `main` has `md:cursor-default md:select-auto md:hover:bg-transparent` on the header; those three classes are dropped here.
4. **`snap-start` → `snap-align-start`** (lines 469 and 538). `snap-align-start` is not a Tailwind utility, so it compiles to nothing — combined with `snap-x snap-mandatory` on the parent, the horizontal pill strip has mandatory snapping enabled with zero snap points.

**Suggestion:** Rebase onto current `main` and drop the `AbsentTeachersTab.tsx` / `PreferredTeachersPanel.tsx` / walkthrough-doc changes entirely — `PreferredTeachersPanel.tsx` and `docs/walkthrough-583-mobile-dashboard-redesign.md` are already byte-identical to `main`, and `AbsentTeachersTab.tsx` only differs by these regressions. That leaves the PR scoped to what #567 actually asks for:

```bash
git fetch origin main
git rebase origin/main
# resolve by taking main's version of AbsentTeachersTab.tsx
git checkout origin/main -- "src/app/[slug]/dashboard/AbsentTeachersTab.tsx" \
  "src/app/[slug]/dashboard/PreferredTeachersPanel.tsx" \
  docs/walkthrough-583-mobile-dashboard-redesign.md
```

---

src/lib/timetableColorUtils.ts
```diff
@@ -0,0 +1,160 @@
...
+}
+
+/**
+ * Checks if a period/subject is a non-teaching break.
+ */
+export function isBreakPeriod(subjectName?: string, className?: string): boolean {
+  const s = (subjectName || "").toLowerCase().trim();
+  const c = (className || "").toLowerCase().trim();
+  const breakKeywords = ["break", "recess", "lunch", "free period", "gap"];
+  return breakKeywords.some((kw) => s.includes(kw) || c.includes(kw));
```
guangshinhaha: **[High] `isBreakPeriod` does an unanchored substring match, so any class or subject containing "gap" — including "Singapore" — is silently deleted from the mobile timetable.**

```ts
export function isBreakPeriod(subjectName?: string, className?: string): boolean {
  const s = (subjectName || "").toLowerCase().trim();
  const c = (className || "").toLowerCase().trim();
  const breakKeywords = ["break", "recess", "lunch", "free period", "gap"];
  return breakKeywords.some((kw) => s.includes(kw) || c.includes(kw)); // ←
}
```

**Problem:** `"singapore".includes("gap")` is `true` (`sin-**gap**-ore`). In an MOE school, `Singapore Studies`, `Singapore History` and `Contemporary Singapore` are ordinary subjects. `TimetableGrid.tsx:622` uses this predicate to `return null` for the whole period card, so a teacher opening their mobile timetable on a day with a Singapore Studies lesson sees that period **completely absent** from the list — no card, no row, no indication anything was filtered. The same predicate also feeds `hasTeachingEntryMap` in `getContextualPeriodTimer`, so the countdown bar skips that lesson too and reports the *following* period as "next class".

`"break"` has the same shape of problem (`Breakdance`, `Breakfast Club`), as does `"lunch"` (`Lunch Duty` is a real assigned duty, not a free slot).

**Suggestion:** Match whole normalised values rather than substrings, and drop `"gap"` (it is not a term that appears in imported timetables):

```ts
const BREAK_LABELS = new Set([
  "break", "recess", "lunch", "free period", "free", "assembly break",
]);

export function isBreakPeriod(subjectName?: string, className?: string): boolean {
  const s = (subjectName || "").toLowerCase().trim();
  const c = (className || "").toLowerCase().trim();
  // Whole-value match only — substring matching mis-classifies real subjects
  // (e.g. "Singapore Studies" contains "gap").
  return BREAK_LABELS.has(s) || BREAK_LABELS.has(c);
}
```

> If partial matching is genuinely needed, use word-boundary regexes (`/\bbreak\b/`) instead — but a whole-value set is safer here because break rows come from a small, known set of import labels.

---

src/app/[slug]/dashboard/timetable/TimetableGrid.tsx
```diff
@@ -616,123 +699,120 @@ export default function TimetableGrid({
...
+                          {mobileSwap.className}
+                        </p>
+                        <p className="text-base font-extrabold text-teal-950 mt-0.5">
+                          {mobileSwap.subject}
+                        </p>
+                        <p className="mt-1 text-xs font-semibold text-teal-800">
+                          Covering for {mobileSwap.otherTeacherName}
+                        </p>
+                      </div>
+                    ) : entry && colorScheme ? (
```
Comment (guangshinhaha) [inline src/app/[slug]/dashboard/timetable/TimetableGrid.tsx:805]: **[High] The `swapCovered` branch was dropped from the mobile card, so a period a teacher has swapped away now renders as a normal lesson they still have to teach.**

```tsx
) : mobileSwap?.type === "swapCovering" ? (
  <div className="mt-2 p-3 rounded-lg bg-teal-50 border border-teal-200">
    ...Covering for {mobileSwap.otherTeacherName}
  </div>
) : entry && colorScheme ? (          // ← swapCovered falls through to here
  <div className="mt-2 p-3 rounded-lg border transition-all" style={{...}}>
    <div ...>{entry.className}</div>
    <div ...>{entry.subject}</div>
  </div>
) : mobileBlock ? (
```

**Problem:** The pre-change mobile branch chain had a `mobileSwap?.type === "swapCovered"` case that rendered the class/subject struck through plus a teal `Swap` badge and `Covered by {otherTeacherName}`. It is gone. `swapCovered` rows are built in `timetable/page.tsx:224` from `periodSubstitution.findMany({ where: { timetableEntry: { teacherId: selectedTeacherId } } })`, so the teacher’s **own** `TimetableEntry` always exists for those slots and `entryMap` always has a hit — meaning every `swapCovered` period now falls straight into the `entry && colorScheme` branch.

Concrete failure: Ms Tan agrees a mutual swap giving away her Wed P3 4E Maths lesson. On desktop she still sees "Covered by Mr Lim · Swap". On her phone she sees a plain coloured `4E / MATHEMATICS` card identical to every lesson she is actually teaching, and turns up to a class that is not hers. Mutual swaps are in scope for the pilot (#585, merged in `c540117`), so this is a live path.

**Suggestion:** Restore the branch ahead of the generic `entry` case:

```tsx
) : mobileSwap?.type === "swapCovering" ? (
  ...
) : mobileSwap?.type === "swapCovered" ? (
  <div className="mt-2 p-3 rounded-lg bg-slate-100 border border-slate-200 opacity-60">
    <div className="flex items-center gap-1.5">
      <p className="text-xs font-bold tracking-wide uppercase text-slate-500 line-through">
        {mobileSwap.className}
      </p>
      <span className="shrink-0 rounded-full bg-teal-500 px-1.5 py-0.5 text-[9px] font-bold uppercase tracking-wide text-teal-950">
        Swap
      </span>
    </div>
    <p className="text-base font-extrabold text-slate-600 line-through mt-0.5">
      {mobileSwap.subject}
    </p>
    <p className="mt-1 text-xs font-semibold text-teal-700">
      Covered by {mobileSwap.otherTeacherName}
    </p>
  </div>
) : entry && colorScheme ? (
```

---

src/app/[slug]/dashboard/timetable/TimetableGrid.tsx
```diff
@@ -616,123 +699,120 @@ export default function TimetableGrid({
...
+                    <div className="flex items-center justify-between mb-2">
+                      <div className="flex items-baseline gap-2">
+                        <span className={`text-lg font-bold tracking-tight ${isPast ? "text-slate-500" : "text-slate-900"}`}>
+                          {period.startTime} – {period.endTime}
+                        </span>
+                      </div>
+                      <span className={`text-xs font-bold uppercase tracking-wider px-2.5 py-1 rounded-md border ${
+                        isPast ? "bg-slate-100 text-slate-400 border-slate-200" : "bg-slate-100 text-slate-800 border-slate-300"
+                      }`}>
+                        {entry?.className || mobileOverlay?.className ? "VENUE / ROOM" : "FREE"}
```
Comment (guangshinhaha) [inline src/app/[slug]/dashboard/timetable/TimetableGrid.tsx]: **[High] The period badge renders the literal placeholder string `VENUE / ROOM` on every mobile lesson card.**

```tsx
<span className={`text-xs font-bold uppercase tracking-wider px-2.5 py-1 rounded-md border ${
  isPast ? "bg-slate-100 text-slate-400 border-slate-200" : "bg-slate-100 text-slate-800 border-slate-300"
}`}>
  {entry?.className || mobileOverlay?.className ? "VENUE / ROOM" : "FREE"}   {/* ← */}
</span>
```

**Problem:** Two defects in one expression.

1. The truthy branch is a hard-coded design placeholder, not data. Every teacher opening the mobile timetable sees a badge reading `VENUE / ROOM` on every lesson — there is no `venue`/`room` field on `TimetableEntry` (`prisma/schema.prisma` has only `className` and `subject`), so this can never resolve to anything real.
2. The condition ignores `mobileSwap` and `mobileBlock`. For a `swapCovering` period (`entry` undefined, `mobileOverlay` undefined, `mobileSwap` set) the badge reads **`FREE`** while the body directly underneath renders `4E / MATHEMATICS · Covering for Ms Tan`. Same for a recurring `Unavailable` block — badge says `FREE`, body says `Unavailable`.

**Suggestion:** The card already leads with the time range, so use the badge for the period number — the one piece of information the new layout dropped and which the edit-mode variant still shows (`Period {period.number}`):

```tsx
<span className={`text-xs font-bold uppercase tracking-wider px-2.5 py-1 rounded-md border ${
  isPast ? "bg-slate-100 text-slate-400 border-slate-200" : "bg-slate-100 text-slate-800 border-slate-300"
}`}>
  Period {period.number}
</span>
```

> If a status word is wanted instead, derive it from the same value the body branches on rather than a second, narrower condition — e.g. `const hasContent = !!(effectiveClass || effectiveSubject || mobileBlock);` and render `{hasContent ? `P${period.number}` : "FREE"}` — so the badge can never contradict the card body.

---

src/lib/timetableColorUtils.ts
```diff
@@ -0,0 +1,160 @@
...
+/**
+ * Checks if a period's end time has passed for today.
+ */
+export function isPeriodPast(endTimeStr?: string, dayOfWeek?: number, selectedDateStr?: string): boolean {
+  if (!endTimeStr) return false;
+  
+  const now = new Date();
+  
+  if (selectedDateStr) {
+    const todayStr = now.toISOString().split("T")[0];
```
guangshinhaha: **[Medium] `isPeriodPast` compares a local-calendar date string against a UTC date string, so between 00:00 and 08:00 SGT the previous school day is treated as "today".**

```ts
const now = new Date();

if (selectedDateStr) {
  const todayStr = now.toISOString().split("T")[0];   // ← UTC date, not local
  if (selectedDateStr < todayStr) return true;
  if (selectedDateStr > todayStr) return false;
} else if (dayOfWeek) {
  ...
}
// falls through to a LOCAL hour/minute comparison
const currentHours = now.getHours();
```

**Problem:** `selectedDateStr` comes from `weekDates`, which `src/lib/timetable/weekDates.ts` builds with `Date.UTC(d.getFullYear(), d.getMonth(), d.getDate())` — i.e. the **local** calendar date rendered as `YYYY-MM-DD`. `now.toISOString()` is the **UTC** date. In SGT (UTC+8) those disagree for every local time between 00:00 and 07:59.

Concrete failure: it is 01:30 SGT on Tue 28 Jul. `now.toISOString()` is `2026-07-27T17:30:00Z`, so `todayStr = "2026-07-27"`. A teacher taps back to **Monday** (`selectedDateStr = "2026-07-27"`): the strings are equal, so the early return is skipped and the function compares Monday’s `endTime` against the current *local* clock — `01 > 15` is false, `01 === 15` is false — and returns `false`. Every Monday lesson renders as an active, fully-coloured card instead of the collapsed grey "finished" row, and `getPersistentClassColor(..., isPast)` gives them live colours.

**Suggestion:** Derive today’s string from local calendar parts, matching how `weekDates` is built:

```ts
const now = new Date();

if (selectedDateStr) {
  const todayStr = `${now.getFullYear()}-${String(now.getMonth() + 1).padStart(2, "0")}-${String(now.getDate()).padStart(2, "0")}`;
  if (selectedDateStr < todayStr) return true;
  if (selectedDateStr > todayStr) return false;
} else if (dayOfWeek) {
```

---

src/app/[slug]/dashboard/timetable/TimetableGrid.tsx
```diff
@@ -551,8 +561,45 @@ export default function TimetableGrid({
           })}
         </div>
 
+        {/* Sticky Slim contextual countdown bar (Only on Mobile) */}
+        {(() => {
+          const nowDayOfWeek = nowDate.getDay() === 0 ? 7 : nowDate.getDay();
+          const isSelectedDayToday = selectedDay === nowDayOfWeek;
```
Comment (guangshinhaha) [inline src/app/[slug]/dashboard/timetable/TimetableGrid.tsx]: **[Medium] The countdown bar keys "is this today" off day-of-week only, so it shows a live lesson timer while the teacher is browsing a different week.**

```tsx
const nowDayOfWeek = nowDate.getDay() === 0 ? 7 : nowDate.getDay();
const isSelectedDayToday = selectedDay === nowDayOfWeek;   // ←  ignores which week is loaded
const timerInfo = getContextualPeriodTimer(
  periods, (pid) => {...}, isSelectedDayToday, nowDate
);
```

**Problem:** `TimetableGrid` has week navigation (`goToWeek` → `?week=<monday>`, `weekMonday`/`weekDates` props), so `selectedDay` is a weekday *within the currently loaded week*, not necessarily this week. `isSelectedDayToday` only compares weekday numbers.

Concrete failure: it is Wednesday 10:15. A teacher taps "previous week" to check what they taught last Wednesday. `selectedDay === 3 === nowDayOfWeek`, so the sticky blue bar renders `"25m remaining in current class (ends at 10:40)"` with a live progress bar over a week that ended five days ago. The same happens forward — planning next Wednesday shows a countdown to a class that is seven days away. Note `isPeriodPast` *does* use `weekDates`, so the cards below the bar correctly collapse as "finished" while the bar above claims a lesson is in progress.

**Suggestion:** Compare the actual date of the selected column against today, reusing the date string already computed for the cards:

```tsx
const todayStr = `${nowDate.getFullYear()}-${String(nowDate.getMonth() + 1).padStart(2, "0")}-${String(nowDate.getDate()).padStart(2, "0")}`;
const selectedDateStr = weekDates?.[selectedDay - 1];
// Fall back to weekday comparison only when the parent did not supply weekDates.
const isSelectedDayToday = selectedDateStr
  ? selectedDateStr === todayStr
  : selectedDay === (nowDate.getDay() === 0 ? 7 : nowDate.getDay());
```

---

src/app/[slug]/dashboard/timetable/TimetableGrid.tsx
```diff
@@ -566,20 +613,56 @@ export default function TimetableGrid({
               ? blockMap.get(`${period.id}-${selectedDay}`)
               : undefined;
 
+            // Subject / Class names
+            const effectiveSubject = mobileOverlay?.subject || mobileSwap?.subject || entry?.subject;
+            const effectiveClass = mobileOverlay?.className || mobileSwap?.className || entry?.className;
+
+            // Omit breaks AND free periods completely in mobile view unless editing
+            const isFreePeriod = !effectiveSubject && !effectiveClass && !mobileBlock;
+            if (!isEditMode && (isBreakPeriod(effectiveSubject, effectiveClass) || isFreePeriod)) {
```
Comment (guangshinhaha) [inline src/app/[slug]/dashboard/timetable/TimetableGrid.tsx]: **[Medium] Filtering out free and break periods with no empty state leaves a completely blank mobile timetable for any day with no lessons.**

```tsx
// Omit breaks AND free periods completely in mobile view unless editing
const isFreePeriod = !effectiveSubject && !effectiveClass && !mobileBlock;
if (!isEditMode && (isBreakPeriod(effectiveSubject, effectiveClass) || isFreePeriod)) {
  return null;   // ←  every period can return null, and nothing catches that case
}
```

**Problem:** `periods.map(...)` is rendered straight into `<div className="space-y-3">` with no check for "did anything survive the filter". When every period on the selected day is free or a break, the map produces an array of nothing but `null` and the mobile view renders the day tabs, then empty space, then the end of the page.

Concrete failure: a teacher with no Friday timetable (common for part-timers, and the default state for a school mid-import) taps **Fri** and gets a blank screen with no explanation — indistinguishable from a failed load. Before this change those slots rendered `Free period` rows, so the day was visibly empty-but-loaded. The countdown bar does not fill the gap either: `getContextualPeriodTimer` returns `null` when no period passes `hasTeachingEntryMap`.

**Suggestion:** Compute the visible set once, then branch:

```tsx
const visiblePeriods = isEditMode
  ? periods
  : periods.filter((p) => {
      const e = entryMap.get(`${p.id}-${selectedDay}`);
      const ov = overlayMap.get(`${selectedDay}-${p.id}`);
      const sw = !ov ? swapOverlayMap.get(`${selectedDay}-${p.id}`) : undefined;
      const blk = !e && !ov && !sw ? blockMap.get(`${p.id}-${selectedDay}`) : undefined;
      const subj = ov?.subject || sw?.subject || e?.subject;
      const cls = ov?.className || sw?.className || e?.className;
      if (isBreakPeriod(subj, cls)) return false;
      return !!(subj || cls || blk);
    });

// ...
<div className="space-y-3">
  {visiblePeriods.length === 0 ? (
    <div className="rounded-xl border border-border bg-card px-6 py-12 text-center">
      <p className="text-sm font-medium text-foreground">No lessons scheduled</p>
      <p className="mt-1 text-xs text-muted-foreground">
        Nothing on this teacher&apos;s timetable for {DAY_NAMES[selectedDay - 1]}.
      </p>
    </div>
  ) : (
    visiblePeriods.map((period) => { /* existing body, minus the early return */ })
  )}
</div>
```

> This also removes the duplicated overlay/entry lookup — the filter and the card body currently recompute `effectiveSubject`/`effectiveClass` from the same three maps.

---

src/lib/timetableColorUtils.ts
```diff
@@ -0,0 +1,160 @@
...
+ */
+export function timeToMinutes(timeStr: string): number {
+  const [h, m] = timeStr.split(":").map((n) => parseInt(n, 10));
+  return (h || 0) * 60 + (m || 0);
+}
+
+/**
+ * Returns contextual timer status for current time vs today's periods.
+ */
+export function getContextualPeriodTimer(
```
guangshinhaha, line 117: **[Medium] New pure-logic module ships with zero tests, against the repo’s test-first rule for `src/lib/` functions.**

```ts
export function getContextualPeriodTimer(
  periods: Array<{ id: string; startTime: string; endTime: string }>,
  hasTeachingEntryMap: (periodId: string) => boolean,
  isToday: boolean,
  nowDate: Date = new Date()
): { type: "CURRENT_LESSON" | "NEXT_LESSON" | "DAY_ENDED" | "BEFORE_DAY"; ... } | null {   // ←
```

**Problem:** `CLAUDE.md` › Testing requires spec-driven tests written before or alongside new `src/lib/` logic, in `src/__tests__/**/*.test.ts` (the suite enforced by `.husky/pre-push`). `src/lib/timetableColorUtils.ts` adds four exported pure functions — all time- and string-sensitive, all trivially unit-testable because `getContextualPeriodTimer` already accepts an injectable `nowDate` — and no test file was added. That is exactly why the two boundary bugs flagged elsewhere in this review (the `"gap"` substring match and the UTC/local date comparison in `isPeriodPast`) reached review undetected.

Note also that `"BEFORE_DAY"` is in the return union but is never produced by any branch — a test enumerating the four states would have caught that too.

**Suggestion:** Add `src/__tests__/timetableColorUtils.test.ts` covering the boundaries:

```ts
import { describe, it, expect } from "vitest";
import { isBreakPeriod, isPeriodPast, timeToMinutes, getContextualPeriodTimer } from "@/lib/timetableColorUtils";

const periods = [
  { id: "p1", startTime: "08:00", endTime: "09:00" },
  { id: "p2", startTime: "09:30", endTime: "10:30" },
];
const teaching = (id: string) => id === "p1" || id === "p2";

describe("isBreakPeriod", () => {
  it("flags real break rows", () => {
    expect(isBreakPeriod("Recess")).toBe(true);
    expect(isBreakPeriod(undefined, "Lunch")).toBe(true);
  });
  it("does not flag subjects that merely contain a keyword", () => {
    expect(isBreakPeriod("Singapore Studies", "4E")).toBe(false); // "gap" substring
  });
});

describe("isPeriodPast", () => {
  it("treats an earlier calendar date as past regardless of clock time", () => {
    // pin the clock inside the 00:00-08:00 SGT window that breaks the UTC compare
    expect(isPeriodPast("15:00", 1, "2026-07-27")).toBe(true);
  });
});

describe("getContextualPeriodTimer", () => {
  it("returns null when the selected day is not today", () => {
    expect(getContextualPeriodTimer(periods, teaching, false, new Date())).toBeNull();
  });
  it("reports remaining minutes inside a lesson", () => {
    const r = getContextualPeriodTimer(periods, teaching, true, new Date(2026, 6, 27, 8, 40));
    expect(r).toMatchObject({ type: "CURRENT_LESSON", progressPercent: 67 });
  });
  it("reports the break countdown between lessons", () => {
    const r = getContextualPeriodTimer(periods, teaching, true, new Date(2026, 6, 27, 9, 15));
    expect(r).toMatchObject({ type: "NEXT_LESSON", targetPeriodId: "p2" });
  });
  it("reports day ended after the last period", () => {
    const r = getContextualPeriodTimer(periods, teaching, true, new Date(2026, 6, 27, 17, 0));
    expect(r).toMatchObject({ type: "DAY_ENDED" });
  });
});
```

---

src/app/[slug]/dashboard/AbsentTeachersTab.tsx
```diff
@@ -338,69 +343,116 @@ export default function AbsentTeachersTab({
...
+              <div className="px-4 py-3 md:px-5">
+              <div className="flex md:flex-wrap gap-2 overflow-x-auto pb-2 md:pb-0 scrollbar-thin snap-x snap-mandatory">
```
Comment (guangshinhaha) [inline src/app/[slug]/dashboard/AbsentTeachersTab.tsx]: **[Low] `scrollbar-thin` is not a utility this project has — it compiles to nothing.**

```tsx
<div className="flex md:flex-wrap gap-2 overflow-x-auto pb-2 md:pb-0 scrollbar-thin snap-x snap-mandatory">  {/* ← */}
```

**Problem:** `scrollbar-thin` comes from the `tailwind-scrollbar` plugin. `package.json` has only `tailwindcss@^4` and `@tailwindcss/postcss@^4` — no scrollbar plugin — and `src/app/globals.css` defines no `.scrollbar-thin` utility. Tailwind v4 silently emits nothing for an unknown class, so the pill strip keeps the platform-default scrollbar and a future reader is left believing a thin scrollbar is configured somewhere.

**Suggestion:** Drop the class, or define the utility explicitly in `globals.css` if a thin scrollbar is actually wanted:

```tsx
<div className="flex md:flex-wrap gap-2 overflow-x-auto pb-2 md:pb-0 snap-x snap-mandatory">
```

```css
/* or, in src/app/globals.css */
@utility scrollbar-thin {
  scrollbar-width: thin;
}
```

---

src/app/[slug]/dashboard/PeriodsToFillTab.tsx
```diff
@@ -61,6 +65,13 @@ export default function PeriodsToFillTab({
   const [removingCoverageFor, setRemovingCoverageFor] = useState<string | null>(null);
   const [tagFilter, setTagFilter] = useState("");
   const [tagOptions, setTagOptions] = useState<string[]>([]);
+  const [localFilterStatus, setLocalFilterStatus] = useState<"uncovered" | "covered" | "all">("uncovered");
+
+  const filterStatus = activeFilterStatus ?? localFilterStatus;
```
Comment (guangshinhaha) [inline src/app/[slug]/dashboard/PeriodsToFillTab.tsx]: **[Low] Controlled/uncontrolled hybrid: `localFilterStatus` is written on every change but can never be read while the only caller passes `activeFilterStatus`.**

```tsx
const [localFilterStatus, setLocalFilterStatus] = useState<"uncovered" | "covered" | "all">("uncovered");

const filterStatus = activeFilterStatus ?? localFilterStatus;   // ←  always the prop today
const setFilterStatus = (status: "uncovered" | "covered" | "all") => {
  setLocalFilterStatus(status);       // dead write
  onFilterStatusChange?.(status);
};
```

**Problem:** `ReliefOverview.tsx:642` always passes `activeFilterStatus={periodsFilterStatus}`, so `activeFilterStatus` is never `undefined` and `localFilterStatus` is never read. It still triggers a state update and re-render on every tab press, and it leaves two sources of truth that will diverge the moment someone adds a second call site that passes only `onFilterStatusChange` — the segmented control would then appear to respond (local state moves) while the parent’s value drives the actual list.

**Suggestion:** Make the prop required and drop the local copy, matching how `activeTab` is already owned by `ReliefOverview`:

```tsx
type Props = {
  ...
  filterStatus: "uncovered" | "covered" | "all";
  onFilterStatusChange: (status: "uncovered" | "covered" | "all") => void;
};

export default function PeriodsToFillTab({ ..., filterStatus, onFilterStatusChange }: Props) {
  // no localFilterStatus; call onFilterStatusChange directly from the segmented buttons
```

> If the component genuinely needs to work standalone, keep it uncontrolled-capable but guard the write: `if (activeFilterStatus === undefined) setLocalFilterStatus(status);`

---

src/app/[slug]/dashboard/ReliefOverview.tsx
```diff
@@ -356,37 +364,95 @@ export default function ReliefOverview({
...
       <div className="grid grid-cols-2 gap-3 md:grid-cols-4">
-        <div className="rounded-xl border border-border bg-destructive/10 p-4 shadow-sm">
-          <p className="text-xs font-medium uppercase tracking-wide text-destructive">
-            Absent Teachers
-          </p>
+        <button
+          type="button"
+          onClick={() => {
+            setActiveTab("absent");
+            scrollToTabSection();
```
Comment (guangshinhaha) [inline src/app/[slug]/dashboard/ReliefOverview.tsx:383]: **[Low] The summary cards are always interactive, but the section they scroll to is only rendered when there are absent teachers — so on a clean day the buttons are visibly dead.**

```tsx
onClick={() => {
  setActiveTab("absent");
  scrollToTabSection();   // ←  tabContentRef.current is null when sickTeachers.length === 0
}}
className="group ... transition-all hover:border-destructive/50 hover:shadow active:scale-[0.98] min-h-[44px]"
```

**Problem:** `tabContentRef` is attached at line 601 inside `{!isLoading && sickTeachers.length > 0 && (...)}`. When nobody is absent, all three cards still render as buttons with hover borders, a chevron affordance and an `active:scale-[0.98]` press animation — but `scrollToTabSection()` short-circuits on the null ref and `setActiveTab` changes nothing visible. A coordinator taps "Uncovered Periods · 0", feels the press animation, and nothing happens; the same tap on a day with absences jumps them down the page. The affordance promises navigation that does not exist.

**Suggestion:** Gate the interactivity on the same condition that gates the target:

```tsx
const hasTabSection = !isLoading && sickTeachers.length > 0;

<button
  type="button"
  disabled={!hasTabSection}
  onClick={() => { setActiveTab("absent"); scrollToTabSection(); }}
  className={`group flex flex-col justify-between rounded-xl border border-border bg-destructive/10 p-4 text-left shadow-sm min-h-[44px] ${
    hasTabSection
      ? "transition-all hover:border-destructive/50 hover:shadow active:scale-[0.98]"
      : "cursor-default"
  }`}
>
  <div className="flex items-center justify-between">
    <p className="text-xs font-medium uppercase tracking-wide text-destructive">Absent Teachers</p>
    {hasTabSection && <ChevronRight className="h-4 w-4 text-destructive/60 transition-transform group-hover:translate-x-0.5" />}
  </div>
```

---

guangshinhaha:
## Code Review Summary

| Severity | Count |
|----------|-------|
| Critical | 0 |
| High     | 4 |
| Medium   | 4 |
| Low      | 3 |

**Blocking first:** this branch was cut from `7a78303`, before #589 merged, and re-implements the same four dashboard files. GitHub reports `mergeable: CONFLICTING`. `PreferredTeachersPanel.tsx` and the #583 walkthrough doc are already byte-identical to `main`, and `AbsentTeachersTab.tsx` differs from `main` only by reverting #589’s desktop guards. Rebasing onto `main` and dropping those files would shrink this PR to just the #567 timetable work — and would resolve the largest finding on its own.

## What Looks Good

- **`getContextualPeriodTimer` takes `nowDate` as an injectable parameter** rather than calling `new Date()` internally — the countdown logic is fully deterministic and unit-testable without fake timers. Pairing that with the 60s `setInterval` ticker in `TimetableGrid` (rather than a server refetch) keeps the live bar free of network cost.
- **The colour hash is seeded with `userId:className:subject`** and normalises with `.trim().toLowerCase()`, so a class keeps the same hue across sessions and devices without any persisted state — and two teachers viewing the same class get different palettes, which is the right call for a per-user mnemonic.
- **The mobile/desktop split in `AbsentTeachersTab` uses `hidden md:flex` / `flex md:hidden` on sibling blocks** instead of a JS breakpoint hook, so both variants stay server-renderable with no hydration mismatch and no layout flash.
- **`PeriodsToFillTab`’s empty states are now filter-aware** — "all covered" vs "nothing assigned yet" vs "no periods recorded" are three distinct messages keyed to `filterStatus`, replacing two early returns that could only express the first and third.

---

Issue #19, opened 2026-04-09
[Feature Request] test


guangshinhaha:
## Feature Request

**Submitted by:** tan_guang_shin@moe.edu.sg
**Category:** other

### Description
testt

### Use Case
test

---
*Submitted via the feedback form*

guangshinhaha:
Closing — junk/test issue.

PR #603, opened 2026-08-01
fix(security): remove hardcoded OAuth `hd=*` wildcard from Google auth URL (MAIA #19)

### LINKED ISSUE #601: fix(security): remove hardcoded OAuth `hd=*` wildcard from Google auth URL (MAIA pentest #19)
## Finding

MAIA Penetration Test (2026-07), finding **#19** — *"OAuth `hd=*` wildcard hardcoded in Google auth URL"*.
Severity: **Informational (CVSS 0.0)**. The last open item from the report; findings #1–#18 (2 High, 16 Medium) are all closed.

`src/app/api/auth/google/route.ts:48` hardcodes `hd: "*"` in the Google authorization URL:

```ts
const params = new URLSearchParams({
  client_id: clientId,
  redirect_uri: `${appUrl}/api/auth/google/callback`,
  response_type: "code",
  scope: "openid email profile",
  state,
  hd: "*",
  prompt: "select_account",
});
```

## Why this is not a vulnerability

`hd` (hosted domain) is a **hint to Google's account-picker UI only**. Google's own documentation states it does not control access to the application and that the relying party must verify the domain server-side. A user can authenticate with any account regardless of the value sent.

Access is actually gated in the callback (`src/app/api/auth/google/callback/route.ts`):

1. `userInfo.email_verified !== true` → reject (`email_not_verified`)
2. `isAllowedEmailDomain(domain)` against `ALLOWED_EMAIL_DOMAINS` in `src/lib/authDomains.ts` → reject (`domain_not_allowed`)
3. No matching `User` row → reject (`user_not_found`)

Only after all three does `createSession()` run. MAIA's own risk note agrees: the wildcard "only matters if the callback doesn't validate the email domain — and it does."

## Decision: remove the parameter (do not risk-accept)

Considered and rejected:

- **Pin a specific domain** (`hd: "moe.edu.sg"`) — not viable. `hd` accepts a single value, but the pilot allowlists three domains (`moe.edu.sg`, `hci.edu.sg`, `gmail.com`). Pinning one breaks Hwa Chong and stakeholder logins.
- **Make it env-configurable** (`GOOGLE_OAUTH_HD`) — satisfies the finding literally, but adds a config surface with no consumer during the pilot. Revisit only if a single-tenant Workspace deployment appears.
- **Risk-accept** — rejected. Removal is a one-line change; carrying an open pentest finding into launch for a line of code that does nothing is not worth the paperwork.

Removing it also fixes a latent inconsistency: `hd=*` restricts the account picker to Google Workspace accounts, which filters out consumer accounts — working against `gmail.com` being deliberately allowlisted for pilot/stakeholder logins (see `CLAUDE.md`, "Permitted SSO exception — MOE iCON").

## Scope

- `src/app/api/auth/google/route.ts` — drop `hd` from the params and update the stale comment.
- `tests/api/google-oauth.test.ts:58,74` — the one assertion expecting `hd=*`; assert the param is absent instead.

No env vars, docs, or other call sites reference it.

## Acceptance criteria

- [ ] `hd` is absent from the Google authorization URL
- [ ] Test asserts `url.searchParams.has("hd") === false`
- [ ] Callback domain-allowlist behaviour is unchanged and still covered by tests
- [ ] Google SSO login verified working for a `moe.edu.sg` account after deploy

## Out of scope

Verifying the `hd` **claim** returned by Google on the callback (i.e. requiring institutional emails to originate from the Workspace tenant rather than an unmanaged consumer account at that domain) is a genuine defence-in-depth improvement, but it is a separate change and not required to close #19. Track separately if wanted.

## Follow-up

Report to GovTech (@Cyril CHEN) that #19 is **fixed, not risk-accepted**, so the retest covers all 19 findings.

guangshinhaha:
Closes #601.

## Summary

Removes the hardcoded `hd: "*"` parameter from the Google OAuth authorization URL in `src/app/api/auth/google/route.ts`, and updates the one test that asserted on it.

## Why

MAIA penetration test (2026-07) finding **#19**, rated **Informational (CVSS 0.0)** — the last open item on the report. Findings #1–#18 (2 High, 16 Medium) are already closed.

`hd` (hosted domain) is a hint to Google's **account-picker UI only**. Google's documentation states it does not control access to the application and that the relying party must verify the domain server-side. Access is actually gated in the callback:

1. `email_verified !== true` → reject
2. domain not in `ALLOWED_EMAIL_DOMAINS` (`src/lib/authDomains.ts`) → reject
3. no matching `User` row → reject

Only then does `createSession()` run. **None of that changes in this PR** — the parameter carried no security weight in either direction. MAIA's own risk note agrees: the wildcard "only matters if the callback doesn't validate the email domain — and it does."

### Why remove rather than risk-accept

- **Pinning a specific domain isn't viable** — `hd` takes a single value, but the pilot allowlists three (`moe.edu.sg`, `hci.edu.sg`, `gmail.com`). Pinning one breaks Hwa Chong and stakeholder logins.
- **Env-configurable (`GOOGLE_OAUTH_HD`)** — satisfies the finding literally but adds a config surface with no consumer during the pilot. Revisit if a single-tenant Workspace deployment appears.
- **Risk acceptance** — not worth the paperwork when the fix is one line.

Removal also clears a latent inconsistency: `hd=*` restricts the picker to Google Workspace accounts, hiding the consumer accounts that `gmail.com` is deliberately allowlisted for (see `CLAUDE.md`, "Permitted SSO exception — MOE iCON").

## Test plan

- `npm test` — 864 passed, 1 skipped (83 files)
- `npm run test:api -- google-oauth` — the `hd` assertion now expects the param to be absent and passes
- `npm run build` — clean, `✓ Generating static pages (93/93)`

### Pre-existing failures (not introduced here)

Two callback tests in `tests/api/google-oauth.test.ts` fail identically on unmodified `main`; both are stale tests, not defects:

1. `error=domain_not_allowed for non-moe.edu.sg email` — uses `teacher@gmail.com`, but `gmail.com` was later added to `ALLOWED_EMAIL_DOMAINS`, so the request correctly falls through to `user_not_found`.
2. `redirects to /[slug]/reporting for TEACHER role` — the route now redirects TEACHERs to `/[slug]/teacher`.

Left out of scope to keep this PR to the security finding. Happy to fix in a follow-up.

## Post-merge

Report to GovTech that #19 is **fixed, not risk-accepted**, so the retest covers all 19 findings. Verify a real `moe.edu.sg` SSO login on the deploy.


guangshinhaha:
**[Medium] The only test covering the `domain_not_allowed` gate — the exact control this PR leans on — is currently broken, so removing `hd` lands with zero passing coverage of the server-side domain check**

📍 `tests/api/google-oauth.test.ts:192` (outside the diff hunk, so posted as a PR comment)

```ts
it("redirects to login with error=domain_not_allowed for non-moe.edu.sg email", async () => {
    await mockCookiesForCallback("test-state");
    mockGoogleApiSuccess("teacher@gmail.com");   // ← gmail.com IS allowlisted

    const { GET } = await import("@/app/api/auth/google/callback/route");
    const req = new NextRequest(
      "http://localhost/api/auth/google/callback?code=authcode&state=test-state"
    );
    const res = await GET(req);

    expect(res.status).toBe(307);
    expect(res.headers.get("location")).toContain(
      "/login?error=domain_not_allowed"                 // ← never reached
    );
  });
```

**Problem:** This PR's justification for dropping `hd` is that access is gated server-side in the callback (`isAllowedEmailDomain`, `src/app/api/auth/google/callback/route.ts:95`). That is correct — but the regression test guarding it no longer exercises it. `gmail.com` is in `ALLOWED_EMAIL_DOMAINS` (`src/lib/authDomains.ts:8`), so `teacher@gmail.com` passes the domain check, falls through to `prisma.user.findUnique` against a truncated DB, and redirects to `?error=user_not_found`. The assertion fails — and more importantly, `isAllowedEmailDomain` returning `false` is never asserted anywhere: there is no unit test either (`grep -rn isAllowedEmailDomain src/__tests__` → no hits; `src/__tests__/access/trust-tier.test.ts` only covers the trust-tier helpers).

Net effect of merging as-is: the client-side picker hint is gone, and the server-side check that fully replaces it has no green test. If someone later widens `ALLOWED_EMAIL_DOMAINS` or reorders the callback guards, nothing fails.

Fully acknowledged that this is pre-existing on `main` and explicitly scoped out in the PR description — but it is a one-string fix in a file this PR already touches, and it is the single test that makes this PR's security argument verifiable.

**Suggestion:**
```ts
it("redirects to login with error=domain_not_allowed for a non-allowlisted domain", async () => {
    await mockCookiesForCallback("test-state");
    // Must be a domain absent from ALLOWED_EMAIL_DOMAINS — gmail.com is
    // allowlisted for pilot logins, so it exercises user_not_found instead.
    mockGoogleApiSuccess("teacher@outlook.com");

    const { GET } = await import("@/app/api/auth/google/callback/route");
    const req = new NextRequest(
      "http://localhost/api/auth/google/callback?code=authcode&state=test-state"
    );
    const res = await GET(req);

    expect(res.status).toBe(307);
    expect(res.headers.get("location")).toContain(
      "/login?error=domain_not_allowed"
    );
  });
```

The sibling stale test (~line 284, expecting `/[slug]/reporting` when the callback now sends `TEACHER` to `/[slug]/teacher` — `callback/route.ts:158`) is a one-line string swap in the same file. Fixing both keeps `npm run test:api -- google-oauth` green rather than leaving CI red on the very file this PR edits.

---

guangshinhaha:
## Code Review Summary

| Severity | Count |
|----------|-------|
| Critical | 0 |
| High     | 0 |
| Medium   | 1 |
| Low      | 0 |

## What Looks Good
- **The security reasoning is correct and verified against the code.** `hd` is a picker hint with no enforcement weight; the real gate is the callback's three-step chain — `email_verified !== true` → `isAllowedEmailDomain(domain)` → `User` row lookup — and `createSession()` only runs after all three (`src/app/api/auth/google/callback/route.ts:80–129`). Removing the parameter changes no access decision.
- **Removal was the right call over the alternatives.** `hd` accepts a single value while the pilot allowlists three domains (`moe.edu.sg`, `hci.edu.sg`, `gmail.com`), so pinning would have broken Hwa Chong and stakeholder logins; an env-configurable `GOOGLE_OAUTH_HD` would have added a config surface with no consumer. Choosing "fixed, not risk-accepted" also keeps the GovTech retest covering all 19 findings.
- **Fixes a latent UX inconsistency, not just the finding.** `hd=*` restricts the picker to Workspace accounts, which silently contradicted `gmail.com` being deliberately allowlisted per `CLAUDE.md` — so the change is a net improvement beyond satisfying the pentest.
- **The replacement comment explains *why the parameter is absent*,** which is exactly the kind of thing a future contributor would otherwise re-add "for security". Better than a silent deletion, and the test comment mirrors it so the intent survives in both places.
- **Tightly scoped.** One parameter, one assertion, no drive-by refactoring in an auth path — the right blast radius for a security fix that has to be re-verified against a live `moe.edu.sg` login post-merge.

---

guangshinhaha:
Addressed in 48a40f3 — both broken assertions repaired.

- `domain_not_allowed`: swapped `teacher@gmail.com` → `teacher@outlook.com`, so the domain gate is actually exercised (gmail.com is allowlisted, so the old case fell through to `user_not_found`).
- TEACHER landing: expectation and test name updated `/[slug]/reporting` → `/[slug]/teacher` to match `callback/route.ts:158`.

Verified before/after: `npm run test:api -- tests/api/google-oauth.test.ts` was **2 failed | 16 passed**, now **18 passed**.

PR #625, opened 2026-08-05
chore(scripts): add idempotent prod seed for product-team superadmins

guangshinhaha:
## Summary

Adds `scripts/seed-prod-superadmins.ts` (+ an `npm run seed:superadmins` entry) to grant `SUPERADMIN` to the three product-team accounts in one idempotent, previewable run:

- `tan_guang_shin@moe.edu.sg`
- `calebazx@gmail.com`
- `goh_jun_hong_gary@moe.edu.sg`

## Why

`scripts/upsert-superadmin.ts` already exists but takes **one email per invocation**, so bootstrapping the team in a new environment means three separate runs with a hand-pasted prod connection string each time — easy to typo, and with no way to preview what it will do before it writes.

This is the fixed-list equivalent, written to be safe to point at prod:

- **Requires exactly one of `--dry-run` / `--apply`.** A bare invocation is a no-op, so a mistyped command can't write to whatever DB it was pointed at.
- **No connection-string default.** Exits if neither `DIRECT_URL` nor `DATABASE_URL` is set, rather than silently falling back to a local database.
- **Lifts, never overwrites.** Only `role` and `onboardingComplete` are written. An account that already exists as a `TEACHER` attached to a school keeps its `schoolId` and `name`.
- **Logs the DB host only**, never the connection string (it carries a password).
- **Pre-flight domain check** against `isAllowedEmailDomain`, so we can't seed a `SUPERADMIN` who then has no way to sign in.

`upsert-superadmin.ts` is intentionally kept — it's the generic one-off tool, this is the fixed roster. Happy to fold them into one if reviewers would rather not have both.

### On `calebazx@gmail.com`

Valid as a superadmin: `assertRoleAllowedForEmail` exempts `SUPERADMIN` from the institutional-domain rule, and `gmail.com` is in `ALLOWED_EMAIL_DOMAINS`. The gmail restriction only blocks `SCHOOL_ADMIN`. Flagging it explicitly since it looks like a violation at a glance.

## How to run

```bash
DIRECT_URL="<prod connection string>" npm run seed:superadmins -- --dry-run   # preview
DIRECT_URL="<prod connection string>" npm run seed:superadmins -- --apply     # commit
```

Imports stay within `src/lib`, per the `scripts/` COPY constraint documented in the Dockerfile, so this also runs from the deployed image via `docker exec`.

## Test plan

Verified against a real Postgres (local `reliefcher-dev-db`, brought up to date with `prisma migrate deploy`):

- [x] No flag, and both flags together → refuses to run, exits with usage.
- [x] `--dry-run` → correctly read existing state (2 already superadmin, 1 to create) and wrote nothing.
- [x] `--apply` → created the missing user.
- [x] `--apply` a second time → all three reported `SKIP`. Idempotent.
- [x] **Role-lift preservation:** downgraded one account to `TEACHER` with a `schoolId` and a custom name, re-ran `--apply` → role became `SUPERADMIN` while `name` and `schoolId` were preserved unchanged.
- [x] `npm test` → 883 passed, 1 skipped.
- [x] `npm run build` → clean, `✓ Generating static pages (93/93)`, exit 0.
- [x] `npm audit` → 0 vulnerabilities.
- [x] `tsc --noEmit` → clean.

No automated test accompanies this — it's a one-off ops script with no app-code callers, so the coverage above is manual-against-a-real-DB. Say the word if you'd rather it had a test in `tests/api/`.

## Human verification still needed

Nothing in this diff touches an auth path, but the *effect* is an auth one. After `--apply` runs against prod, each of the three should confirm they can actually sign in and reach the superadmin surfaces — the script can create the row but can't prove the login works end-to-end.


guangshinhaha:
Ops follow-up tracked in #626 — @chadinwork / @fatin-suhaimi to run the seed against the AWS prod DB once this merges and the merge-commit image is live in ECR.

Note the sequencing there: the script ships inside the image, so the ECS task being exec'd into has to be running an image built from this merge commit or later.

PR #627, opened 2026-08-05
fix(reporting): support multi-day per-period relief instructions for FULL_DAY absences

### LINKED ISSUE #622: fix(reporting): support multi-day per-period relief instructions for FULL_DAY absences
## User story

As a teacher reporting a multi-day full-day absence, I want to enter relief instructions for each period on each day of my absence, so that relief teachers covering any day of my absence have the guidance they need for that specific day.

## Background

Multi-day `PARTIAL_DAY` absences already support this: a "Select Date" dropdown lets the teacher pick a day within the absence range, and relief instructions are entered and stored per (date, period). `FULL_DAY` absences spanning multiple days have no equivalent. The date dropdown is gated to `coverageType === "PARTIAL_DAY"` only (`AbsenceReportForm.tsx:1318`), so for a multi-day `FULL_DAY` absence the Relief Instructions section can only ever show and collect notes for a single fixed day (whichever day `currDate` happened to initialize to), leaving every other day in the range with no way to receive instructions at all.

Confirmed by direct code review, not assumption:
- `currDate` state only changes via the `PARTIAL_DAY`-only dropdown's `onChange`; nothing else updates it for `FULL_DAY`.
- The Relief Instructions section's `periodsForNotes` computation (`AbsenceReportForm.tsx` ~line 1449-1455) filters by `currDate` for `PARTIAL_DAY`, but for `FULL_DAY` returns `teacherPeriods` unfiltered, and `teacherPeriods` itself only ever reflects one calendar date (fetched via an effect keyed on `[schoolId, teacherId, currDate]`, ~line 540).

### Relief instructions are entered separately per day for a multi-day full-day absence

- **Given** a teacher is reporting a `FULL_DAY` absence spanning multiple days, with relief needed
- **When** they select a specific date from the date list and enter relief instructions for a period on that date
- **Then** those instructions are saved against that specific date only, and do not appear when a different date within the same absence range is selected

### Single-day full-day absences are unaffected

- **Given** a teacher is reporting a `FULL_DAY` absence for a single day
- **When** they fill in relief instructions
- **Then** no date selector is shown, and the flow matches current single-day behaviour exactly (no regression)

### Date selector defaults inside the reported range

- **Given** a teacher is reporting a future-dated, multi-day `FULL_DAY` absence
- **When** the form loads
- **Then** the date selector's initial value falls within the absence's own start/end range, not an unrelated default such as today's date

## Out of scope

- Changes to the existing `PARTIAL_DAY` multi-day flow, which already works and is the reference pattern for this fix.
- `NO_RELIEF` coverage type, which collects no relief instructions and is unaffected.
- A "copy instructions to all days" convenience/bulk-entry feature. Worth considering later, but not required to close this gap.

## Design assets

N/A. No new UI pattern is being introduced. The reference implementation to mirror already exists and is shipped: `AbsenceReportForm.tsx`'s existing `PARTIAL_DAY` date-selector + per-day/per-period instructions flow (`AbsenceReportForm.tsx:1318-1340` for the selector, `selectedPeriods`/`lessonNotes` state keyed by date throughout the same file).

---

## Technical context

Starting points for whoever grooms this (from this session's investigation, not verified as complete):
- `AbsenceReportForm.tsx:1318` — widen the date-selector's gating condition to include `FULL_DAY` when `spanDays > 1`, not just `PARTIAL_DAY`.
- `AbsenceReportForm.tsx` ~1449-1455 — the `FULL_DAY` branch of `periodsForNotes` needs to filter by `currDate` the same way the `PARTIAL_DAY` branch already does.
- `AbsenceReportForm.tsx` ~685-687 — the `FULL_DAY` payload-construction branch (`periodsPayload = Object.fromEntries(dateWindow.map((d) => [d.value, []]))`) is a separate code path from `PARTIAL_DAY`'s and needs checking against the fix, not assumed correct.
- The `lessonNotes`/`lessonNotesPayload` state is already keyed `Record<date, Record<periodId, string>>` and the submit-payload loop already iterates multiple dates (~line 674-732) — this may already carry multi-date `FULL_DAY` notes through correctly once the UI populates it for more than one date, but confirm rather than assume.

`POST /api/absence-reports` (`src/app/api/absence-reports/route.ts:347-411`) already reads `multiDayLessonNotes`/`multiDayPeriods` from the request body and persists them date-scoped via `writeAbsencePeriodRequirements`. Confirm this already handles a `FULL_DAY` payload with multiple populated dates correctly; the `FULL_DAY` payload shape is constructed by a different branch on the frontend than `PARTIAL_DAY`'s (see Technical context), so don't assume parity without checking.

---

guangshinhaha:
Fixes #622

## Summary

A multi-day `FULL_DAY` absence could only ever collect relief instructions for **one** day of the range. Every other day silently got nothing.

The "Select Date" dropdown was rendered **inside** the `PARTIAL_DAY` block, so for a `FULL_DAY` absence nothing could move `currDate`. Since the absent teacher's periods are fetched one calendar date at a time (keyed on `currDate`), the Relief Instructions section could only ever show — and store notes for — whichever day `currDate` happened to initialise to.

A relief teacher covering day 2+ of a colleague's absence was therefore guaranteed to receive no instructions, with nothing on the reporting side hinting anything was missing.

## Why the fix is frontend-only

The backend already supports this. `writeAbsencePeriodRequirements`'s `FULL_DAY` branch iterates every date key in `datePeriods` and resolves `lessonNotes[dateStr][period.id]` per day, and the route's `multiDayLessonNotes` cleaner is date-keyed already. The form was the only thing that couldn't produce a multi-date payload.

## Changes

**1. The date selector renders for any multi-day absence that needs relief** (the actual fix)

Extracted into a shared `dateSelector` element gated on `showDateSelector = needsRelief && spanDays > 1`:
- `PARTIAL_DAY` keeps it inside its existing period-picker card — no visual change.
- `FULL_DAY` gets its own card directly above Relief Instructions, which is the only section it scopes.

**2. Per-date period cache (`periodsByDate`)**

`teacherPeriods` only ever describes the day currently on screen. Submit iterates *every* dated note, so it needs its own per-date view — otherwise a note for Tuesday gets resolved against Monday's period ids. This was latent before (only one day was reachable); enabling multi-day makes it live, so it's fixed in the same change. Cleared when the teacher changes, kept when the date changes.

**3. Payload builder extracted to `src/lib/absenceNotesPayload.ts`**

Moves ~45 lines out of `handleSubmit` so the per-date resolution is unit-testable without mounting the form. Behaviour-preserving apart from the `periodsByDate` fix above and one related correction: the fallback-notes loop now also drops dates outside the reported window (the per-period loop already did). Without that, typing a note and then *shortening* the range sends an out-of-window date key, and the API rejects the whole submission.

## Acceptance criteria

| Criterion | How it's met |
|---|---|
| Instructions saved against the selected date only, not visible on other dates | `lessonNotes`/`generalLessonNotes` are already keyed by date and the textarea reads `lessonNotes[currDate]`; the selector now makes more than one date reachable. Covered by tests. |
| Single-day `FULL_DAY` unchanged — no selector, no regression | `showDateSelector` requires `spanDays > 1`. Covered by tests. |
| Selector defaults inside the absence's own range | Already handled by the existing `dateWindow` effect, which snaps `currDate` to `startDate` whenever it falls outside the window. Not modified. |

## Test plan

**Automated** — `src/__tests__/absenceNotesPayload.test.ts`, 13 cases derived from the issue's acceptance criteria:
- multi-day `FULL_DAY`: each day's instructions stay scoped to that day; no bleed onto other days in the range
- the fallback note resolves against *that date's* own periods, not the visible day's (the specific regression)
- out-of-window dates dropped; whitespace-only notes skipped; trimming
- single-day `FULL_DAY` passes through unchanged
- `PARTIAL_DAY` reference flow unchanged (period filtering, orphaned days, fallback expansion)

```
npm test        →  86 files, 896 passed | 1 skipped
tsc --noEmit    →  exit 0
npm run build   →  exit 0, ✓ Generating static pages
eslint (changed files) → 0 errors (1 pre-existing exhaustive-deps warning, untouched)
```

**Manual — please verify before merge.** I did not drive the real form in a browser: setting a multi-day range means driving the `DatePicker` popover (Radix + react-day-picker), which is too brittle to pin in jsdom, and the change isn't deployed anywhere I could reach. Worth ~2 minutes on a preview deploy:

1. Report an absence spanning 3 weekdays, **Full day**, relief needed.
2. Confirm a **Select Date** dropdown appears above Relief Instructions, listing all 3 weekdays.
3. Click **+ Add relief instructions**, enter a note on day 1.
4. Switch to day 2 → the period rows should reload for day 2's timetable and the note boxes should be **empty**.
5. Enter a different note on day 2, submit.
6. On the coordinator dashboard, check day 1 and day 2 each show their own instructions.
7. Regression: a **single-day** full-day absence shows **no** date selector, and a multi-day **partial-day** absence looks exactly as it did before.

## Notes for the reviewer

- The `FULL_DAY` selector renders whenever the absence is multi-day, even before the teacher clicks "+ Add relief instructions". Kept deliberately consistent with `PARTIAL_DAY`, which is also always-visible when multi-day — happy to gate it on `includeHandover` instead if you'd prefer less chrome.
- Out of scope per the issue: the `PARTIAL_DAY` flow, `NO_RELIEF`, and a "copy instructions to all days" bulk-entry convenience.
- The legacy single-string `notes` field (used only when the teacher has no timetabled periods at all) is still read from the DOM at submit time, so it reflects the last-viewed day only. Pre-existing, untouched, and separate from the per-period path this issue is about.


src/app/[slug]/reporting/AbsenceReportForm.tsx
```diff
@@ -1407,6 +1405,12 @@ export default function AbsenceReportForm({
             </div>
           )}
  
+          {/* FULL_DAY has no period picker to hang the day selector off, so it
+              gets its own card immediately above the instructions it scopes. */}
+          {showDateSelector && coverageType === "FULL_DAY" && (
```
Comment (guangshinhaha) [inline src/app/[slug]/reporting/AbsenceReportForm.tsx:1459]: **[Medium] For a `FULL_DAY` day with no timetabled periods, the new date selector lets a teacher type instructions per day but only the last-viewed day is ever saved**

```tsx
{/* FULL_DAY has no period picker to hang the day selector off, so it
    gets its own card immediately above the instructions it scopes. */}
{showDateSelector && coverageType === "FULL_DAY" && (   // ←
  <div className="rounded-lg bg-muted/20 p-3">{dateSelector}</div>
)}
```

**Problem:** The relief-instructions panel falls back to the single `name="notes"` textarea whenever `periodsForNotes.length === 0` — for `FULL_DAY` that is exactly `teacherPeriods` being empty (line 1453-1459), i.e. an FJT/new hire, or a day the teacher has no lessons. Its value is stored per date in `generalLessonNotes[currDate]`, but neither sink keeps more than one day:

- `buildLessonNotesPayload`'s fallback loop does `targetPeriodIds = periodsByDate[date]` and `if (targetPeriodIds.length === 0) continue`. The fallback textarea renders **iff** that array is empty for `FULL_DAY`, so every dated fallback note is dropped — the branch is unreachable for `FULL_DAY`.
- `generalNotes` (line 720-723) reads `formData.get("notes")` off the live DOM, which only ever holds `generalLessonNotes[currDate]` — the day currently on screen.

Repro: 3-day `FULL_DAY` absence for a teacher with no timetable → type "Mon: cover in Hall 4B" on day 1 → switch to day 2 (only possible because of this PR) → type "Tue: bring lab keys" → submit. Day 1's text is gone from both `notes` and `multiDayLessonNotes`, with no warning. Mixed case is worse: teacher has lessons Mon but not Tue, types a Tue fallback note, switches back to Mon, submits → `generalNotes` is `""` (Mon has periods) and the Tue note is dropped by the builder, so it vanishes entirely.

**Suggestion:** Since there is no period to hang a dated note on, fold every dated fallback note into the single `notes` string instead of letting the DOM read decide:

```tsx
// AbsenceReportForm.handleSubmit — replace the formData.get("notes") read
const datedFallback = Object.entries(generalLessonNotes)
  .filter(([d, t]) => dateWindow.some((dw) => dw.value === d) && t.trim())
  .sort(([a], [b]) => a.localeCompare(b));

const generalNotes =
  includeHandover && (!teacherPeriods || teacherPeriods.length === 0)
    ? datedFallback.map(([d, t]) => `${d}: ${t.trim()}`).join("\n")
    : "";
```

Alternative if that changes the `notes` contract too much: render the fallback textarea only when `spanDays === 1`, and for multi-day show an explicit "no timetable on file — instructions apply to the whole absence" single box, so nothing is silently per-date.

---

src/app/[slug]/reporting/AbsenceReportForm.tsx
```diff
@@ -539,6 +551,13 @@ export default function AbsenceReportForm({
     };
   }, [schoolId, teacherId, startDate, currDate]);
 
+  // A different teacher has a different timetable, so nothing cached above
+  // still applies. Cleared separately from the fetch effect so switching
+  // *dates* keeps the cache (that's the whole point of it).
+  useEffect(() => {
+    setPeriodsByDate({});
```
Comment (guangshinhaha) [inline src/app/[slug]/reporting/AbsenceReportForm.tsx:567]: **[Medium] Switching the absent teacher clears `periodsByDate` but leaves `lessonNotes` / `generalLessonNotes` / `selectedPeriods` behind, so the previous teacher's instructions get written onto the new teacher's absence**

```tsx
// A different teacher has a different timetable, so nothing cached above
// still applies. Cleared separately from the fetch effect so switching
// *dates* keeps the cache (that's the whole point of it).
useEffect(() => {
  setPeriodsByDate({});           // ← only the period cache is reset
}, [schoolId, teacherId]);
```

**Problem:** The comment's own premise — "a different teacher has a different timetable, so nothing cached above still applies" — is equally true of `lessonNotes`, `generalLessonNotes` and `selectedPeriods`, none of which are reset here (the only resets are the "Submit Another" handler at line 836/845-846 and `setSelectedPeriods({})` on "No relief needed" at line 1127).

Reachable in the coordinator/admin flow, where `lockedTeacherId` is absent and the teacher Combobox is live (line 985-992): pick teacher A → multi-day `FULL_DAY` → "+ Add relief instructions" → type notes on Mon and Tue → realise it's the wrong person → pick teacher B → submit. `lessonNotes` still holds `{MON: {p1: "A's note"}, TUE: {...}}`. The `FULL_DAY` branch of `buildLessonNotesPayload` applies **no** period filter, and `writeAbsencePeriodRequirements`'s `FULL_DAY` branch resolves `lessonNotes[dateStr][period.id]` against *all school periods* — A's period ids are valid school periods, so B's relief teacher receives A's instructions. This PR widens the blast radius (previously only one day's worth of stale notes could be carried over; now every visited day's can).

`selectedPeriods` has the same problem on the `PARTIAL_DAY` path, where the stale ids go straight into `multiDayPeriods`.

**Suggestion:**
```tsx
// A different teacher has a different timetable, so nothing captured against
// the old one still applies — cached periods *or* the notes/periods keyed to
// them. Cleared separately from the fetch effect so switching *dates* keeps
// the cache (that's the whole point of it).
useEffect(() => {
  setPeriodsByDate({});
  setLessonNotes({});
  setGeneralLessonNotes({});
  setSelectedPeriods({});
}, [schoolId, teacherId]);
```

(Guard with a `useRef` first-run check if wiping on mount is a concern — it is a no-op here since all four already start empty.)

---

src/app/[slug]/reporting/AbsenceReportForm.tsx
```diff
@@ -687,49 +738,16 @@ export default function AbsenceReportForm({
...
-
-          for (const pid of targetPeriodIds) {
-            filteredNotes[date][pid] = trimmed;
-          }
-        }
-
-        if (Object.keys(filteredNotes).length > 0) lessonNotesPayload = filteredNotes;
+        // #622 — resolves each date's notes against that date's own periods
+        // (`periodsByDate`), not against whichever day is currently on screen.
+        lessonNotesPayload = buildLessonNotesPayload({
```
Comment (guangshinhaha) [inline src/app/[slug]/reporting/AbsenceReportForm.tsx:787]: **[Medium] The out-of-window 400 the new filter is meant to prevent is still fully reachable — `parseDateKey` runs over `multiDayPeriods`, not over the lesson notes**

```tsx
if (coverageType === "PARTIAL_DAY") {
  periodsPayload = selectedPeriods; // ← never filtered against dateWindow
} else if (coverageType === "FULL_DAY") {
  periodsPayload = Object.fromEntries(dateWindow.map((d) => [d.value, []]));
}

if (includeHandover) {
  lessonNotesPayload = buildLessonNotesPayload({
    coverageType,
    dateWindow: dateWindow.map((d) => d.value),   // ← window filter applied here only
    ...
```

**Problem:** The PR description justifies the new `inWindow` filter on the fallback-notes loop with *"typing a note and then shortening the range sends an out-of-window date key, and the API rejects the whole submission."* That rejection does not actually come from the notes. `writeAbsencePeriodRequirements` only calls `parseDateKey(dateStr, windowStart, windowEnd)` over `Object.entries(input.datePeriods)` / `Object.keys(input.datePeriods)` (`src/lib/absencePeriodRequirements.ts:87,115,145`); `lessonNotes` is only ever read as a lookup `input.lessonNotes?.[dateStr]?.[periodId]`, so a stray note key is silently ignored, never thrown on.

The key that *does* throw is `multiDayPeriods`, which is still sent unfiltered from `selectedPeriods`. Repro: `PARTIAL_DAY`, Mon–Wed, tick a period on Wed, then move End Date back to Tue. `selectedPeriods` keeps `{..., "2026-08-12": ["p1"]}`, `parseDateKey` throws `AbsencePeriodValidationError`, the whole `$transaction` rolls back and the route returns 400 `Date "2026-08-12" is after the absence end.` — the absence is not created and the message is meaningless to a teacher.

(`FULL_DAY` is safe because its `periodsPayload` is derived from `dateWindow`, so this bites the `PARTIAL_DAY` path only — but it is the exact failure the new filter was added to close.)

**Suggestion:** Filter the period payload the same way the notes payload is now filtered:

```tsx
if (coverageType === "PARTIAL_DAY") {
  const inWindow = new Set(dateWindow.map((d) => d.value));
  periodsPayload = Object.fromEntries(
    Object.entries(selectedPeriods).filter(
      ([date, ids]) => inWindow.has(date) && ids.length > 0
    )
  );
} else if (coverageType === "FULL_DAY") {
  periodsPayload = Object.fromEntries(dateWindow.map((d) => [d.value, []]));
}
```

Note the submit-button guard (line 1548-1550) counts periods across *all* dates, so it will not stop a submission whose only selections are out-of-window — that check needs the same filter, or the API returns "Periods required for partial-day absence."

---

src/app/[slug]/reporting/AbsenceReportForm.tsx
```diff
@@ -589,6 +608,38 @@ export default function AbsenceReportForm({
   const isLongTerm =
     spanDays !== null && spanDays > LONG_TERM_THRESHOLD && !spanInvalid;
 
+  // #622 — a multi-day absence needs a day picker so relief instructions are
+  // entered against the day they apply to. This used to render only inside
+  // the PARTIAL_DAY block, which meant a multi-day FULL_DAY absence had no
+  // way to move off whichever day `currDate` initialised to — every other day
+  // in the range could never receive instructions at all.
+  const showDateSelector = needsRelief && spanDays !== null && spanDays > 1;
```
Comment (guangshinhaha) [inline src/app/[slug]/reporting/AbsenceReportForm.tsx]: **[Low] `showDateSelector` gates on calendar days but the dropdown is populated with weekdays only, so a weekend-spanning absence renders an empty "Select Date" card**

```tsx
const showDateSelector = needsRelief && spanDays !== null && spanDays > 1;  // ←

const dateSelector = (
  ...
    <Select items={dateWindow.map(...)} value={currDate} onChange={setCurrDate} />
```

**Problem:** `spanDays` is a raw calendar-day count (line 599-605) while `dateWindow` comes from `generateWeekdaysRange`, which skips Sat/Sun (line 900). They disagree at the edges, and this PR gives the mismatch a second, more visible render site — a standalone card for `FULL_DAY`:

- Sat → Sun: `spanDays === 2`, `dateWindow === []` → the new card renders a "Select Date" label, helper text and a `Select` with **zero** options. The snap-to-start effect is also a no-op here (`if (windowDays.length > 0 && ...)`, line 516), so `value={currDate}` points at an item that does not exist.
- Fri → Sat: `spanDays === 2`, `dateWindow === [Fri]` → a "select a date" dropdown with exactly one choice.

Non-crashing, but the first case shows a control the teacher cannot use and the FULL_DAY submit then fails server-side with "Date window required for full-day absence." (`periodsPayload` is `{}`).

**Suggestion:** Gate on what actually populates the control:

```tsx
const showDateSelector = needsRelief && dateWindow.length > 1;
```

---

guangshinhaha:
## Code Review Summary

| Severity | Count |
|----------|-------|
| Critical | 0 |
| High     | 0 |
| Medium   | 3 |
| Low      | 1 |

The core fix is right: the root cause is correctly identified as "`currDate` had no way to move for `FULL_DAY`", the backend genuinely already supports date-keyed notes (`writeAbsencePeriodRequirements` iterates `datePeriods` keys and looks up `lessonNotes[dateStr][period.id]` per day), and the `periodsByDate` cache closes the latent per-date resolution bug that enabling multi-day would otherwise have made live. The findings below are all about state that the newly-reachable day-switching now exposes, plus one place where the PR's own stated rationale is not actually satisfied by the change.

## What Looks Good

- **Extracting `buildLessonNotesPayload` into `src/lib/` instead of leaving it inline** — the per-date resolution is the whole bug, and pulling it out of a 1500-line client component is what makes the 13 spec-derived cases possible without mounting the form or driving a Radix popover in jsdom.
- **`periodsByDate` fixes the real defect, not just the symptom.** Rendering the selector alone would have shipped a form that lets you type Tuesday notes and then resolves them against Monday's period ids. Catching that the cache was a *precondition* of the UI change, not a follow-up, is the right call.
- **Clearing the cache in its own effect keyed on `[schoolId, teacherId]` rather than folding it into the fetch effect** — the comment explains exactly why (date changes must *keep* the cache), which is the non-obvious half.
- **The negative tests carry their weight**: `MON_PERIODS`/`TUE_PERIODS` deliberately share no period ids, so the "resolved against the wrong day's timetable" regression cannot pass silently, and `periodsByDate: {}` is covered as its own case rather than assumed.
- **Honest test plan.** Explicitly stating that the browser path was not driven, and why, is more useful than a checkbox claiming it was.

---

src/app/[slug]/reporting/AbsenceReportForm.tsx
Comment (guangshinhaha) [reply src/app/[slug]/reporting/AbsenceReportForm.tsx:1459]: Fixed in 981572f — both halves.

Payload side: `buildLessonNotesPayload` now takes `schoolPeriodIds` and a `fallbackPeriodsForDate` that falls back to it when the date resolves to no timetabled periods. That is exactly the condition that renders the fallback textarea on `FULL_DAY`, and `writeAbsencePeriodRequirements` materialises the day against every school period anyway, so the note now has somewhere to land. `PARTIAL_DAY` is untouched — a note there can still only apply to periods that were actually ticked.

Flat `notes` side: dropped the `new FormData(form).get("notes")` read entirely. It is now built from `generalLessonNotes` state via `buildFlatNotesFallback`, which keeps every day in the window and labels each one when there is more than one — that string is the last-resort lessonNote for *every* date in `/api/teacher-dashboard`, so handing it one arbitrary day was the worse half of this. Removed `name="notes"` from the textarea so nothing reads it off the DOM again.

Covered by 4 new cases in `src/__tests__/absenceNotesPayload.test.ts` (`FULL_DAY, teacher with no timetable`) plus a `buildFlatNotesFallback` block. The old case that pinned the drop-the-note behaviour now pins it for `PARTIAL_DAY`, where it is correct.

src/app/[slug]/reporting/AbsenceReportForm.tsx
Comment (guangshinhaha) [reply src/app/[slug]/reporting/AbsenceReportForm.tsx:567]: Fixed in 981572f — the effect now clears `lessonNotes`, `generalLessonNotes` and `selectedPeriods` alongside `periodsByDate`. It still keys on `[schoolId, teacherId]` only, so switching *dates* keeps the cache.

Also caught the mirror gap on the way past: "Submit Another" resets notes and ticks but never learned about the new `periodsByDate`, and with `lockedTeacherId` set (teacher self-report) `teacherId` never changes, so the effect would not fire to cover for it. Added the explicit reset there.

src/app/[slug]/reporting/AbsenceReportForm.tsx
Comment (guangshinhaha) [reply src/app/[slug]/reporting/AbsenceReportForm.tsx:787]: Confirmed and fixed in 981572f — you are right that the window filter in the notes builder never touched this path. `periodsPayload` now uses `selectedPeriodsInDateWindow(selectedPeriods, dateWindowValues)`, which drops stranded date keys and any date left with nothing ticked.

The same value drives the "select at least one period" hint and the submit `disabled` check, which previously summed over *all* dates — so after shortening the range submit could stay enabled on ticks that were about to be discarded, and the request would then fail the other way with "Periods required for partial-day absence." Both now count only what will actually be sent.

Extracted to a lib function rather than left inline so it is unit-testable — 4 cases under `describe("selectedPeriodsInDateWindow")`.

src/app/[slug]/reporting/AbsenceReportForm.tsx
Comment (guangshinhaha) [reply src/app/[slug]/reporting/AbsenceReportForm.tsx]: Fixed in 981572f — `showDateSelector` is now `needsRelief && dateWindow.length > 1`, i.e. gated on the number of *selectable* days rather than the calendar span. Sat–Sun gives an empty `dateWindow` and Fri–Sat gives one entry, so neither renders the card now.

PR #628, opened 2026-08-05
fix(timetable): manual edits always 400 on WEEKLY-cycle schools, and the failure is silent

### LINKED ISSUE #624: fix(timetable): manual timetable edits always 400 on WEEKLY-cycle schools, and the failure is silent
## Summary

Manual timetable editing is **completely broken for every school whose timetable cycle is `WEEKLY`** — which is the default, and covers the pilot schools. Every cell add/edit returns `400`, and the UI shows no error at all: the editor just closes and the change silently vanishes.

Reproduced against `main` (`8303ca1`) on the test env.

---

## Reproduction

Driven through the real UI (Playwright, `maia-pri-1`, SCHOOL_ADMIN):

1. Dashboard → Timetable
2. **Edit Timetable** → **Enable Editing**
3. Click any empty cell, enter a class and subject, click **Save**

Captured request/response:

```
POST https://reliefcher.string.sg/api/timetable -> 400

REQ {"teacherId":"cmrkf1gzh0028194r5bxvncja","dayOfWeek":1,
     "periodId":"cmqrmo5to004p194fcj2pjlg1","className":"QA-CLASS",
     "subject":"QA-SUBJ","weekType":"ODD","date":"2026-08-03"}

RES {"error":"This school's timetable cycle is set to Weekly. Set it to
     Two-week or 10-day in Settings before saving an odd/even entry."}
```

The only thing visible to the user is a `400` in the browser console. The cell editor closes as if the save succeeded, and the entry is not there after a refresh.

---

## Root cause

Two halves that only combine into a bug once both are present.

### 1. The grid always sends `ODD` or `EVEN` — never `ALL`

`src/app/[slug]/dashboard/timetable/page.tsx` types the week type as a two-value union and always resolves it to one of them, regardless of the school's cycle:

```ts
const selectedWeekType: "ODD" | "EVEN" =
  weekType === "ODD" || weekType === "EVEN"
    ? weekType
    : weekTypeForSchool(
        new Date(),
        schoolCycle?.timetableCycle ?? "WEEKLY",
        schoolCycle?.weekOneAnchorDate ?? null
      );
```

`weekTypeForSchool` returns `"ODD" | "EVEN"`; there is no `"ALL"` in its return type. For a `WEEKLY` school the value is date-invariant and meaningless, but it is still `ODD` or `EVEN`.

`TimetableGrid.handleSave` forwards it verbatim:

```ts
body: JSON.stringify({ ..., weekType, date: cellDate }),
```

Note the ODD/EVEN toggle isn't even rendered for these schools (`hasWeekRotation` is false), so the user has no way to see or change what's being sent.

### 2. The API rejects `ODD`/`EVEN` when the school's cycle is `WEEKLY`

`src/app/api/timetable/route.ts`:

```ts
if (wt === "ODD" || wt === "EVEN") {
  const school = await prisma.school.findUnique({
    where: { id: schoolId },
    select: { timetableCycle: true },
  });
  if (school?.timetableCycle === "WEEKLY") {
    return NextResponse.json({ error: "This school's timetable cycle is set to Weekly. ..." }, { status: 400 });
  }
}
```

The guard itself is reasonable — an `ODD`/`EVEN` entry on a `WEEKLY` school would never resolve for any date. But the grid is the *only* writer to this endpoint, and it can never satisfy the guard on a `WEEKLY` school. The result is a 100% failure rate on the default configuration.

Reads are unaffected: the page loads entries with `weekType: { in: [selectedWeekType, "ALL"] }`, so existing `ALL` entries display normally. **Only writes break** — which is why this wasn't caught by viewing the page.

### 3. Compounding: `handleSave` ignores the response status

```ts
const res = await fetch("/api/timetable", { ... });
if (res.ok) {
  const saved = await res.json();
  setLocalEntries(...);
}
closeEditor();          // runs regardless
```

There is no `else`. A non-OK response produces no alert, no inline message, no retained editor state — the failure is invisible. Every other mutating handler in the dashboard (`handleAssign`, `handleReassign`, `handleRemoveCoverage`) surfaces `body.error`; this one does not.

---

## Suggested fix

1. **Send the correct week type for the school's cycle.** Derive `"ALL" | "ODD" | "EVEN"` from `School.timetableCycle` — `WEEKLY` ⇒ `"ALL"` — and widen `selectedWeekType`'s type so the compiler stops accepting the two-value union where three are possible. This is the actual fix; the write then matches how `WEEKLY` entries are stored and read.
2. **Surface the error in `handleSave`.** Add the `else` branch that reads `body.error` and shows it (and keeps the editor open so the input isn't lost), matching the other dashboard handlers.
3. **Regression test.** An API test asserting that a `WEEKLY`-cycle school can create a timetable entry through the same payload the grid sends. The existing coverage evidently exercises the endpoint only with rotation-cycle schools.

## Also visible in the console (separate, lower severity)

The screenshot shows a React **#418** hydration error on the same page. Likely `TimetableGrid`'s mobile-day initialiser, which reads the clock during render:

```ts
const [selectedDay, setSelectedDay] = useState(() => {
  const today = new Date().getDay();   // server (UTC) vs client (SGT) can disagree
  ...
});
```

Server and client can compute different days across the UTC/SGT boundary, producing a hydration mismatch. Worth splitting into its own issue if it isn't already tracked. (Unrelated aside: that initialiser's comment says `0=Mon .. 4=Fri` but the code produces `1=Mon .. 5=Fri`; the code is consistent with its call sites, the comment is wrong.)

---

## Environment

- Branch: `main` @ `8303ca1`
- Env: test (`reliefcher.string.sg`), verified on `maia-pri-1`
- Originally reported on the Keming pilot dashboard (also a `WEEKLY`-cycle school)
- Severity: blocks all manual timetable editing for default-configured schools

guangshinhaha:
Fixes #624

## Summary

Manual timetable editing was **completely broken on every `WEEKLY`-cycle school** — the default, and what the pilot schools run. Every cell add/edit returned `400`, and the UI showed nothing at all: the editor closed as if it had saved.

## Root cause

Two halves that only bite together.

**1. The grid sent its *view* week type as the *storage* week type.**

`selectedWeekType` is typed `"ODD" | "EVEN"` and always resolves to one of them — `weekTypeForSchool` has no third value. On a `WEEKLY` school that value is date-invariant and carries no meaning, but `TimetableGrid.handleSave` forwarded it verbatim. The ODD/EVEN toggle isn't even rendered for these schools, so there was no way to see or change what was being sent.

**2. `POST /api/timetable` rejects ODD/EVEN when the cycle is `WEEKLY`.**

That guard is correct — such an entry would never resolve for any date. But the grid is the *only* writer to that endpoint, so it could never satisfy it. 100% failure rate on the default configuration.

The server side was already right and already covered by #285's tests (`POST /api/timetable — timetableCycle guard`, which asserts both the ODD rejection and the ALL acceptance). Nothing tested what the *grid* sends, so the client half drifted unnoticed.

Reads were unaffected (`weekType: { in: [selectedWeekType, "ALL"] }`), which is exactly why this looked healthy on screen and nobody caught it.

**3. Why it presented as a mystery rather than an error.**

```ts
const res = await fetch("/api/timetable", { ... });
if (res.ok) { ...update local state... }
closeEditor();          // ran regardless
```

No `else`. No alert, no inline message, and the typed class/subject were discarded. Every other mutating handler in the dashboard surfaces `body.error`; this one didn't.

## Changes

- **`src/lib/timetable/writeWeekType.ts`** (new) — `weekTypeForWrite(cycle, viewWeekType)`: `WEEKLY` ⇒ `"ALL"`, any rotation cycle ⇒ the viewed week. Derived from the *cycle* so it mirrors the server guard directly and the two can't drift apart again.
- **`TimetableGrid`** now takes `timetableCycle` explicitly rather than inferring from `hasWeekRotation`. Those are currently equivalent (`hasWeekRotation === cycle !== "WEEKLY"`), but `hasWeekRotation` used to be derived from whether ODD/EVEN *entries* existed — leaning on that equivalence is the same class of coupling that caused this bug.
- **Save failures are surfaced inline** and the editor stays open with the typed values intact.

## Test plan

**New unit tests** — `src/__tests__/writeWeekType.test.ts` (4 cases): `WEEKLY` stores `ALL` for either viewed week; `TWO_WEEK` and `TEN_DAY` keep the viewed week; and a guard asserting `WEEKLY` never yields ODD/EVEN — the exact condition the API rejects.

No new API test: #285 already covers both sides of the server guard. The gap was on the client, which is where the new test sits.

```
npm test        →  86 files, 887 passed | 1 skipped
tsc --noEmit    →  exit 0
npm run build   →  exit 0
eslint (changed files) → no new problems
```

**Verified against the live test env** (`maia-pri-1`, a `WEEKLY`-cycle school) by POSTing both payloads to the real API:

```
what the CURRENT grid sends:  weekType "ODD"  -> 400
  {"error":"This school's timetable cycle is set to Weekly. ..."}

what the FIXED grid sends:    weekType "ALL"  -> 201
  {"id":"...","weekType":"ALL","className":"QA-624", ...}
```

The test entry was deleted afterwards. The original 400 was also reproduced through the real UI in Playwright (click cell → type → Save), recorded in #624.

**Manual check worth doing on the preview deploy:** open Timetable on a WEEKLY-cycle school → Edit Timetable → add an entry → it should persist across a refresh. Then confirm a `TWO_WEEK`/`TEN_DAY` school still writes to the correct odd/even week. To see the new error surfacing, temporarily point a WEEKLY school's cycle at something invalid, or just trust the `!res.ok` branch.

## Notes for the reviewer

- Not fixed here, out of scope but worth a follow-up: `TimetableGrid.tsx` has a pre-existing `react/jsx-key` **error** on the `timeoff` overlay `<td>` (line 978 before this change, 1008 after — the branch is missing `key={dayOfWeek}` where its three sibling branches have it). I left it alone to keep this branch to one logical change.
- The React #418 hydration error visible in the original report is separate — `new Date().getDay()` in a `useState` initialiser, so server (UTC) and client (SGT) can disagree. Noted at the bottom of #624, not addressed here.


src/app/[slug]/dashboard/timetable/TimetableGrid.tsx
```diff
@@ -239,21 +250,35 @@ export default function TimetableGrid({
           periodId: editingCell.periodId,
           className: formClassName,
           subject: formSubject,
-          weekType,
+          // #624 — the *stored* week type follows the school's cycle, not the
+          // week being viewed. A WEEKLY school stores `ALL`; sending the
+          // viewed ODD/EVEN made the server reject every edit.
+          weekType: weekTypeForWrite(timetableCycle, weekType),
```
Comment (guangshinhaha) [inline src/app/[slug]/dashboard/timetable/TimetableGrid.tsx]: **[High] On a `WEEKLY` school that already holds ODD/EVEN entries, a save now creates a duplicate row instead of updating the visible one**

```ts
// TimetableGrid.handleSave
weekType: weekTypeForWrite(timetableCycle, weekType),   // ← WEEKLY ⇒ always "ALL"
date: cellDate,
```

```ts
// src/app/api/timetable/route.ts — the upsert match is weekType-scoped
const existing = await tx.timetableEntry.findFirst({
  where: { schoolId, teacherId, dayOfWeek, periodId, weekType: wt, overrideId }, // ←
});
```

**Problem:** `School.timetableCycle === "WEEKLY"` does **not** imply the school's entries are stored as `ALL`. The aSc odd/even PDF import writes `weekType: "ODD"` / `"EVEN"` (`src/app/api/import-timetable/route.ts:434,447`) and never writes `timetableCycle`, so `WEEKLY` + ODD/EVEN rows is a live combination — as is any school that was set to `TWO_WEEK` and later switched back.

For such a school the grid *reads* `weekType: { in: [selectedWeekType, "ALL"] }` with `selectedWeekType === "ODD"`, so the ODD row is what the admin sees and clicks. The save now posts `ALL`, the server's `findFirst` keys on `weekType: "ALL"`, misses the ODD row, and `create`s a second row — the `@@unique([schoolId, teacherId, dayOfWeek, periodId, weekType, overrideId])` constraint permits it because `weekType` differs.

Concrete sequence on a WEEKLY-cycle school imported from odd/even PDFs:
1. Mon P3 shows `3A / Math` (row A, `weekType: "ODD"`).
2. Admin corrects it to `3A / Science` → row B created (`weekType: "ALL"`). DB now has two rows for Mon P3.
3. Both rows match the read filter. `timetableEntriesForDate` runs `findMany` with no `orderBy` for this call, and `entryMap` is last-write-wins — which of `Math` / `Science` renders is whatever Postgres returns last.
4. Admin deletes the cell → `handleDelete(entry.id)` removes only the row in `entryMap`; the sibling reappears on the next refresh.

Before this PR the POST 400'd, so no duplicate could be written. The fix makes this path reachable, and there's no backfill in the PR — the pre-existing ODD/EVEN rows on WEEKLY schools are exactly the population it bites.

**Suggestion:** collapse the slot server-side so one (teacher, day, period, override) ends up with exactly one row, whatever week type it previously carried:

```ts
// src/app/api/timetable/route.ts, inside the $transaction
const slot = { schoolId, teacherId, dayOfWeek, periodId, overrideId };

// Any sibling row for this slot under a different weekType is stale the
// moment we write the cycle-correct one — otherwise both survive and both
// match the read filter `weekType: { in: [viewed, "ALL"] }`.
await tx.timetableEntry.deleteMany({ where: { ...slot, weekType: { not: wt } } });

const existing = await tx.timetableEntry.findFirst({ where: { ...slot, weekType: wt } });
```

Alternative if deleting siblings is too blunt for rotation schools (a `TWO_WEEK` slot legitimately holds distinct ODD and EVEN rows): scope the `deleteMany` to the WEEKLY case only — `if (wt === "ALL") await tx.timetableEntry.deleteMany({ where: { ...slot, weekType: { not: "ALL" } } })` — which is the only direction this PR newly makes reachable. Either way, worth a covering test in `tests/api/` seeding a WEEKLY school with an `ODD` entry and asserting the save leaves one row.

---

src/app/[slug]/dashboard/timetable/TimetableGrid.tsx
```diff
@@ -239,21 +250,35 @@ export default function TimetableGrid({
...
+      const saved = await res.json();
+      setLocalEntries((prev) => {
+        const key = `${saved.periodId}-${saved.dayOfWeek}`;
+        const without = prev.filter(
+          (e) => `${e.periodId}-${e.dayOfWeek}` !== key
+        );
+        return [...without, saved];
+      });
       closeEditor();
+    } catch {
```
Comment (guangshinhaha) [inline src/app/[slug]/dashboard/timetable/TimetableGrid.tsx]: **[Low] The new `catch` discards the underlying error entirely and also covers the success path's `res.json()`**

```ts
      const saved = await res.json();          // ← inside the same try
      setLocalEntries((prev) => { ... });
      closeEditor();
    } catch {                                  // ← no binding, no log
      setSaveError("Could not save this entry. Check your connection and try again.");
    } finally {
```

**Problem:** Two small things, both in the spirit of what this PR is fixing:

1. The catch takes no binding and logs nothing, so a `TypeError` in the `setLocalEntries` reducer or a malformed-response parse failure leaves no trace anywhere — the admin sees "check your connection", and there is nothing in the console or in a session replay to tell you it wasn't the network. This PR exists because a failure was invisible; silently dropping the exception object re-creates a smaller version of that.
2. It spans `await res.json()` on the **201** path. A 201 whose body doesn't parse (proxy/edge truncation — cf. the nginx-edge behaviour on this deployment) shows "Could not save… check your connection" for a write that *did* land. Not harmful — a retry upserts — but it's a wrong message.

**Suggestion:**
```ts
    } catch (err) {
      console.error("[timetable save]", err);
      setSaveError("Could not save this entry. Check your connection and try again.");
    } finally {
```
and, if you want the 201-parse case to be honest, move the success-body read out of the failure story:
```ts
      const saved = await res.json().catch(() => null);
      if (saved) {
        setLocalEntries((prev) => { ... });
      }
      closeEditor();   // the write succeeded; the server response is just unusable
```

---

src/app/[slug]/dashboard/timetable/TimetableGrid.tsx
```diff
@@ -66,7 +69,10 @@ export default function TimetableGrid({
   periods: Period[];
   entries: TimetableEntry[];
   selectedTeacherId: string | null;
+  /** Week type being *viewed*. Never what an edit is stored as — see `weekTypeForWrite`. */
   weekType: "ODD" | "EVEN";
+  /** The school's declared cycle. Decides how a manual edit is stored (#624). */
+  timetableCycle: TimetableCycle;
```
Comment (guangshinhaha) [inline src/app/[slug]/dashboard/timetable/TimetableGrid.tsx:78]: **[Low] `hasWeekRotation` is now a second, independently-computed view of `timetableCycle` — the two can disagree**

```tsx
  timetableCycle: TimetableCycle;
  hasWeekRotation: boolean;   // ← now fully derivable from the line above
```

```tsx
// page.tsx — computed from the same source, but not from each other
const hasWeekRotation = schoolCycle?.timetableCycle !== "WEEKLY";        // null ⇒ true
...
timetableCycle={schoolCycle?.timetableCycle ?? "WEEKLY"}                 // null ⇒ "WEEKLY"
```

**Problem:** The PR's own rationale is that leaning on the `hasWeekRotation === (cycle !== "WEEKLY")` equivalence is the coupling that caused #624 — but the equivalence is now encoded twice, in two expressions with different null handling. If `schoolCycle` is ever null the props contradict each other: `hasWeekRotation` is `true` so the ODD/EVEN toggle renders, while `timetableCycle` is `"WEEKLY"` so every save stores `ALL` — an admin toggling to "Even week" and typing gets a silently week-agnostic write. It isn't reachable today (the school row is guaranteed by `requireSchoolForSlugOrRedirect`), but keeping two props that must stay consistent is the same drift shape this fix set out to close.

**Suggestion:** drop the prop and derive it inside the grid, so there is one source of truth:
```tsx
// TimetableGrid.tsx — remove `hasWeekRotation` from props
const hasWeekRotation = timetableCycle !== "WEEKLY";
```
```tsx
// page.tsx — remove the now-dead local and the prop
<TimetableGrid
  timetableCycle={schoolCycle?.timetableCycle ?? "WEEKLY"}
  ...
```

---

guangshinhaha:
## Code Review Summary

| Severity | Count |
|----------|-------|
| Critical | 0 |
| High     | 1 |
| Medium   | 0 |
| Low      | 2 |

The High finding is the one worth a decision before merge: the fix is correct for a WEEKLY school whose entries are all `ALL`, but it makes a previously-400'd write path reachable for WEEKLY schools that already hold ODD/EVEN rows (the odd/even PDF import writes those without ever setting `School.timetableCycle`). Those saves now create a second row rather than updating the visible one. Since this PR deliberately carries no backfill, the collapse has to happen in the write path.

## What Looks Good

- **Deriving the stored week type from the cycle rather than the view** is the right axis — `weekTypeForWrite` mirrors the server guard's own predicate (`wt === "ODD" || "EVEN"` × `timetableCycle === "WEEKLY"`), so the two are readable side by side instead of being related only by coincidence.
- **Passing `timetableCycle` explicitly instead of reusing `hasWeekRotation`** correctly refuses to lean on an equivalence that was historically derived from entry data rather than school config — the reasoning in the PR description is the actual root cause, not just the symptom.
- **Keeping the editor open with the typed values on failure** is the right failure mode for this surface: the old `if (res.ok) { … } closeEditor()` discarded the admin's input on every rejection, which is why the bug read as a mystery rather than an error.
- **The test lives in `src/__tests__/`** (pure-logic, no DB) rather than being bolted onto the API suite that already covers the server half — the new coverage sits exactly where the gap was.

---

src/app/[slug]/dashboard/timetable/TimetableGrid.tsx
Comment (guangshinhaha) [reply src/app/[slug]/dashboard/timetable/TimetableGrid.tsx]: Confirmed and fixed in 65b41ce — thanks, this was a real gap. `handlePdfConfirm` (`src/app/api/import-timetable/route.ts`) never passes `timetableCycle` to `executeImport`, so the odd/even PDF path writes ODD/EVEN rows while `School.timetableCycle` stays at its `WEEKLY` default. `weekTypeForSchool`'s WEEKLY branch is hardcoded to return `"ODD"`, so the ODD row is exactly what the grid renders — and the ALL write would have landed beside it.

The fix is server-side, in the `$transaction` in `POST /api/timetable`, since the endpoint has to hold that invariant whatever the client sends.

**Scoping chosen — narrower than "incoming `weekType` is ALL":** collapse only when the write is `ALL` **and** the effective cycle for the row's scope is `WEEKLY`:

- base rows (`overrideId === null`) → `School.timetableCycle`
- override-scoped rows → that override's own `cadence`

The second half matters: `timetableResolver.ts` matches override rows against `override.cadence`, independent of the school's. Gating on `wt === "ALL"` alone would let a `TWO_WEEK`-cadence suspended timetable inside a `WEEKLY` school lose its odd/even pair, because the grid derives its write week type from the *school* cycle. Gating on the cycle means a genuine rotation can't be destroyed by an edit that merely happens to arrive as `ALL` — the `TWO_WEEK` school case you flagged is now an explicit test.

**Promote, don't delete-and-recreate.** Rather than deleting siblings and inserting fresh, the row already occupying the slot is updated in place (preferring an existing `ALL`, else the `ODD` one the admin was looking at) and only the remaining siblings are deleted. `AbsencePeriodRequirement`, `PreferredReliefAssignment` and `PeriodSubstitution` all `onDelete: Cascade` off `timetableEntryId`, and `ReliefAssignment.timetableEntryId` is `SetNull` — deleting the visible row would have quietly taken those with it.

**Tests** — new `POST /api/timetable — non-rotating slot collapse (#628)` block in `tests/api/timetable.test.ts`, 7 cases: the duplicate that used to be written (asserting one row, same id, promoted to `ALL`); ODD+EVEN collapsing to one; an existing `ALL` winning over a stale `ODD`; other days/periods untouched; and the two that must **not** collapse — a `TWO_WEEK` school's pair surviving an explicit `ALL` write, and a `TWO_WEEK`-cadence override's pair surviving. Also added a `timetableCycle` option to `createSchool` in `tests/api/helpers.ts`.

I could not execute `npm run test:api` here: the `reliefcher-test-db` container is not running, and port 5433 is currently held by an unrelated project's Postgres with different credentials (`prisma db push` fails `P1000`). Per the branch instructions I did not start it — CI should be the check on these.

src/app/[slug]/dashboard/timetable/TimetableGrid.tsx
Comment (guangshinhaha) [reply src/app/[slug]/dashboard/timetable/TimetableGrid.tsx]: Fair — swallowing the error was exactly the failure mode this branch exists to remove. Fixed in 65b41ce: the blanket `catch {}` is gone and `handleSave` now has three narrow handlers, one per thing that can actually fail, each logging the exception:

1. **the `fetch` itself** — wrapped alone, since a non-2xx response resolves normally and only the transport throws. `console.error("[TimetableGrid] save request failed", err)` + `"Could not reach the server. Check your connection and try again."`
2. **the error-body parse** on `!res.ok` — `.catch()` on that `res.json()` logs before falling back to the generic message, instead of silently returning `null`.
3. **the success-body parse** — no longer inside the same net as the network call, so a 201 with an unreadable body reports that, not a bogus connection error.

`try { … } finally { setIsMutating(false) }` is all that remains at the top level, so the button can't get stuck.

src/app/[slug]/dashboard/timetable/TimetableGrid.tsx
Comment (guangshinhaha) [reply src/app/[slug]/dashboard/timetable/TimetableGrid.tsx:78]: Agreed, and the divergence was real: `page.tsx` computed `hasWeekRotation = schoolCycle?.timetableCycle !== "WEEKLY"` (true when the school row is missing, since `undefined !== "WEEKLY"`) while the cycle prop used `?? "WEEKLY"` (false). Opposite answers from the same missing row.

Collapsed to a single derivation in 65b41ce:

- `hasWeekRotation` is dropped as a prop; `TimetableGrid` derives it once at the top of the component from the prop it already receives: `const hasWeekRotation = timetableCycle !== "WEEKLY";`. The three call sites (the odd/even toggle and the two `ml-auto` classnames) are unchanged.
- `page.tsx` now derives `const timetableCycle = schoolCycle?.timetableCycle ?? "WEEKLY"` once, immediately after the query, and uses it both for `weekTypeForSchool(...)` and for the prop — that `?? "WEEKLY"` was previously spelled out twice as well.

One prop in, one derivation, no way for them to disagree.

Issue #623, opened 2026-08-05
fix(dashboard): cannot re-assign relief twice — internal list goes empty, external 500s and wedges the slot


guangshinhaha:
## Summary

A coordinator cannot re-assign relief for the same period more than once. Both the internal and external paths break, for two unrelated root causes:

- **Internal (school staff):** after one re-assignment, re-opening the assignment panel for that period shows an **empty candidate list** forever.
- **External (preferred relief):** the **second** re-assignment returns **500 Internal server error**, and leaves the slot permanently stuck as covered — the absence report then can't be deleted either.

Both reproduced against `main` (`8303ca1`) on the test env.

---

### Root cause

`AbsentTeachersTab.tsx` guards candidate fetches with a `fetchingRef` set, which doubles as an "already fetched" cache key:

```ts
const fetchAvailableTeachers = useCallback(async (timetableEntryId: string) => {
  if (fetchingRef.current.has(timetableEntryId)) return;   // early-return guard
  fetchingRef.current.add(timetableEntryId);
  ...
  setAvailableByEntryId((prev) => ({ ...prev, [timetableEntryId]: json.availableTeachers }));
}, [date, weekType, tagFilter]);
```

Every other path that invalidates `availableByEntryId[entryId]` also clears the matching `fetchingRef` entry:

| Path | clears `availableByEntryId` | clears `fetchingRef` |
|---|---|---|
| date / weekType change | ✅ | ✅ |
| fetch error | — | ✅ |
| tag-filter change | ✅ | ✅ |
| `handleAssign` | ✅ | ✅ |
| **`handleReassign`** | ✅ | ❌ **missing** |

`handleReassign` deletes the cached candidate list but leaves the entry id in `fetchingRef`. On the next expand, `fetchAvailableTeachers` early-returns at the guard, nothing re-populates `availableByEntryId[entryId]`, and the panel renders an empty list. It stays empty until the coordinator changes the date, the week type, or the tag filter.

`handleAssign` gets this right — it calls `fetchingRef.current.delete(timetableEntryId)` before clearing the cache. `handleReassign` is missing the same line.

### Repro

1. Dashboard → Absent Teachers → expand an absent teacher.
2. Assign an internal relief teacher to a period.
3. Click the same period again → choose a different teacher (first re-assign works).
4. Click the same period a third time → **candidate list is empty**.

### Suggested fix

Add the missing invalidation in `handleReassign`, mirroring `handleAssign`:

```ts
[...]

PR #629, opened 2026-08-05
fix(dashboard): candidate list goes permanently empty after one re-assignment

guangshinhaha:
Addresses the **internal-teacher half** of #623. The external (preferred) half is a schema-level problem and is deliberately **not** in this PR — see below.

## Summary

After re-assigning a period once, re-opening that period's assignment panel showed an **empty candidate list, permanently**. A coordinator could assign, re-assign once, and then never pick a different teacher for that slot again — until they changed the date, the week type, or the tag filter.

## Root cause

`fetchingRef` in `AbsentTeachersTab` does double duty. It's an in-flight guard, but it also marks a slot as *already fetched*, because `fetchAvailableTeachers` early-returns on it:

```ts
if (fetchingRef.current.has(timetableEntryId)) return;
fetchingRef.current.add(timetableEntryId);
```

So clearing a slot's cached list **without** also clearing its marker leaves that slot empty forever — the early-return fires and nothing ever refills `availableByEntryId[entryId]`.

`handleReassign` did exactly that: it deleted the cached list and left the marker behind.

Every other invalidation path already got this right:

| Path | clears cache | clears marker |
|---|---|---|
| date / weekType change | ✅ | ✅ |
| fetch error | — | ✅ |
| tag-filter change | ✅ | ✅ |
| `handleAssign` | ✅ | ✅ |
| `PeriodsToFillTab.handleReassign` | ✅ | ✅ |
| **`AbsentTeachersTab.handleReassign`** | ✅ | ❌ |

It was the single outlier — `PeriodsToFillTab`'s equivalent handler is correct, which is what confirms the intended pattern rather than me guessing at it.

## Changes

Rather than adding the one missing `delete` line, all three invalidation points now go through a single `invalidateCandidates(entryId)` helper that clears the cached list and the marker together. The half-invalidated state is no longer reachable, so a future call site can't reintroduce this.

Remaining direct `fetchingRef` uses are the ones that should stay direct: the full reset on date/week change, the guard itself, and the store/clear inside the fetch.

## Test plan

```
npm test        →  85 files, 883 passed | 1 skipped
tsc --noEmit    →  exit 0
npm run build   →  exit 0
eslint          →  0 errors (5 pre-existing unused-var warnings, untouched)
```

**No new automated test, deliberately** — flagging this against CLAUDE.md's "write a covering test alongside the fix", rather than quietly skipping it.

The bug lives in the coordination between a `useRef` and a `useState` inside a ~900-line client component. Covering it honestly means mounting `AbsentTeachersTab` with a `SchoolProvider`, `@flow/core`, `PeriodStrip` and `PreferredTeachersPanel`, a full `SickTeacherCard` fixture, and mocked `fetch`, then clicking through nested expand state to assign → re-assign → re-expand. That's a large, brittle test for one assertion. The alternatives I considered and rejected:

- **Extract a `useCandidateCache` hook** and test it with `renderHook` (the pattern `useAutoRefresh.test.ts` uses). Clean in principle, but it pulls four pieces of state and the response side-effects out of a busy component — a much larger diff and regression surface than the bug.
- **Move `fetchingRef` into state** so the invariant becomes a pure function. That changes concurrency semantics: the ref is read synchronously precisely so two rapid clicks can't both pass the guard. Not a change I could verify.

I'd rather ship the small, structural fix and say plainly that it's covered by review + manual check than bolt on a fragile test. Happy to do the hook extraction as a follow-up if you'd prefer the coverage.

**Manual verification (~1 min on a preview deploy):**

1. Dashboard → Absent Teachers → expand an absent teacher.
2. Assign an internal relief teacher to a period.
3. Click the same period → pick a different teacher (this first re-assign works today too).
4. Click the same period a **third** time.
   - **Before:** candidate list is empty.
   - **After:** candidates load normally, and you can keep re-assigning indefinitely.

## Not in this PR

- **The external / preferred half of #623.** `@@unique([timetableEntryId, date, status])` on `PreferredReliefAssignment` also blocks a second `CANCELLED` row, so the second external re-assign hits P2002 → unhandled → 500, and wedges the slot (no `DELETE` handler, and the absence report then won't delete either). Fixing it needs a schema change plus a `prisma migrate dev` run, and there's a real choice to make between a nullable `activeSlotKey` column and a partial index. Left for a separate PR.
- **Adjacent staleness, not fixed:** `handleRemoveCoverage` (in both `AbsentTeachersTab` and `PeriodsToFillTab`) doesn't invalidate either. That's milder — the list goes stale rather than empty, so a teacher freed by removing coverage won't reappear as available until something else invalidates. Kept out to hold this branch to one logical change; worth its own issue.


src/app/[slug]/dashboard/AbsentTeachersTab.tsx
```diff
@@ -231,11 +237,9 @@ export default function AbsentTeachersTab({
...
   ) {
     setExpandedKey(null);
-    setAvailableByEntryId((prev) => {
-      const next = { ...prev };
-      delete next[timetableEntryId];
-      return next;
-    });
+    // #623 — this used to clear only the cached list, leaving the slot's
+    // `fetchingRef` marker behind and the candidate list empty forever.
+    invalidateCandidates(timetableEntryId);
```
Comment (guangshinhaha) [inline src/app/[slug]/dashboard/AbsentTeachersTab.tsx:243]: **[High] `handleReassign` hard-DELETEs via the internal route even when the slot is covered by a PREFERRED assignment, so external → internal reassignment always fails**

Not introduced by this diff, but it lives in the function this PR edits and it sits exactly on the seam with the external half of #623 (#630), so flagging it here rather than letting it fall between the two PRs.

```tsx
    setExpandedKey(null);
    // #623 — this used to clear only the cached list, leaving the slot's
    // `fetchingRef` marker behind and the candidate list empty forever.
    invalidateCandidates(timetableEntryId);   // ← fix is correct; the step below is not

    try {
      const del = await fetch(`/api/relief-assignments/${existingAssignmentId}`, {
        method: "DELETE",                     // ← always the INTERNAL endpoint
      });
      if (!del.ok) { ...; return; }
```

**Problem:** `slot.assignmentId` is the `PreferredReliefAssignment` id whenever `slot.assignmentSource === "PREFERRED"` — `src/app/api/dashboard/route.ts:260-262` maps `assignmentId` and `assignmentSource` off the same row. The reassign panel renders the **Internal Staff** tab unconditionally for a covered slot (`isReassign = slot.isCovered`), so a coordinator can open a period covered by an external/preferred teacher, switch to Internal Staff, and click **Reassign**. That calls `DELETE /api/relief-assignments/{preferredAssignmentId}`, which does `prisma.reliefAssignment.findFirst({ where: { id, schoolId } })` → no row → `404 "Assignment not found."` (`src/app/api/relief-assignments/[id]/route.ts:17-23`). The handler alerts and `return`s before the POST, so switching a period from an external back to an internal teacher is impossible through this panel — the coordinator just sees "Assignment not found." with no way forward.

`cancelCoverage()` already exists for precisely this and is used correctly two functions below in `handleRemoveCoverage` — `handleReassign` is the one cancel path that bypasses it.

**Suggestion:**
```tsx
  async function handleReassign(
    existingAssignmentId: string,
    existingAssignmentSource: "INTERNAL" | "PREFERRED" | null,
    absenceReportId: string,
    timetableEntryId: string,
    reliefTeacherId: string,
    ...
  ) {
    setExpandedKey(null);
    invalidateCandidates(timetableEntryId);

    try {
      const cancelled = await cancelCoverage(slug, existingAssignmentId, existingAssignmentSource);
      if (!cancelled.ok) {
        alert(cancelled.error ?? "Could not remove the current relief teacher. Refresh and try again.");
        return;
      }
      // ...unchanged POST /api/relief-assignments
```
and at the call site (~line 799):
```tsx
    ? handleReassign(
        slot.assignmentId,
        slot.assignmentSource,   // ← new
        card.absenceReportId,
        ...
```

`PeriodsToFillTab.handleReassign` has the identical gap, so whichever PR takes this should fix both. If you'd rather this land in #630 alongside the `PreferredReliefAssignment` unique-constraint change, that's reasonable — it's the same user-visible flow — but note it fails *before* reaching the schema problem, so #630's fix alone won't unblock it.

---

guangshinhaha:
## Code Review Summary

| Severity | Count |
|----------|-------|
| Critical | 0 |
| High     | 1 |
| Medium   | 0 |
| Low      | 0 |

The fix itself is correct. I traced every path that touches `fetchingRef` / `availableByEntryId` and the internal half of #623 is genuinely closed: assign → reassign → re-expand now refetches, and no remaining call site clears one without the other. The one `fetchingRef.current.delete` left without a matching cache clear (the `catch` in `fetchAvailableTeachers`, line 170) is the safe direction — it re-enables a retry rather than stranding an empty list — so the invariant holds.

The single High finding is pre-existing, not introduced here, but it lives inside `handleReassign` and sits on the seam with #630, so it's easy for it to fall between the two PRs.

### Not filed, but worth a line

- **`handleRemoveCoverage` still doesn't invalidate** (both tabs). You already called this out and deferred it — agreed it's milder (stale, not empty), and now that `invalidateCandidates` exists it's a one-liner whenever you pick it up.
- **`PeriodsToFillTab` keeps the raw two-step pattern** in three places (lines 141, 197, 221). The "can't drift apart" guarantee this PR establishes only covers `AbsentTeachersTab`. Not asking for the hook extraction — just noting the other half of the surface is still hand-rolled.
- **On the missing test:** your reasoning holds up. Mounting this component to assert one ref/state invariant would be mostly fixture, and moving `fetchingRef` into state genuinely does change the double-click semantics. I wouldn't block on it. If you do the `useCandidateCache` extraction later, doing it as a shared hook across both tabs would pay for itself twice — coverage *and* the duplication above.

## What Looks Good

- **Fixing the invariant instead of the symptom.** Adding the missing `delete` would have worked; routing all three call sites through `invalidateCandidates` makes the half-invalidated state unrepresentable, which is the difference between fixing this bug and fixing this bug class.
- **The root-cause writeup is the actual value here.** The observation that `fetchingRef` is doing double duty — in-flight guard *and* already-fetched marker — is the non-obvious part, and the invalidation-path table in the description is what makes it reviewable in two minutes instead of twenty.
- **Using `PeriodsToFillTab.handleReassign` as the correctness oracle** rather than inferring the intended pattern from scratch — that's how you tell a real outlier from a deliberate asymmetry.
- **The scope boundary is drawn honestly**, with the external half's P2002 mechanism and the `handleRemoveCoverage` gap both named rather than silently left out.

---

src/app/[slug]/dashboard/AbsentTeachersTab.tsx
Comment (guangshinhaha) [reply src/app/[slug]/dashboard/AbsentTeachersTab.tsx:243]: Confirmed and fixed in 4fd0922 — both tabs.

Traced it end to end before changing anything: `src/app/api/dashboard/route.ts:177-185` sets `id: pra.id` together with `source: "PREFERRED"` on the same map entry, so `slot.assignmentId` really is a `PreferredReliefAssignment` id for external coverage, and `isReassign = slot.isCovered` (AbsentTeachersTab.tsx:595) does render the Internal Staff tab for those slots. The internal `DELETE` 404s at `relief-assignments/[id]/route.ts:17-23` and the handler returns before the POST. Reachable, and the flow is a dead end for the coordinator.

**What changed**

Rather than patch the `cancelCoverage()` call into two places and leave the cancel-then-assign pair duplicated, I moved the whole sequence into `reassignCoverage()` in `src/lib/coverageActions.ts`, next to `cancelCoverage()`. It cancels by source, and returns early without assigning if the cancel fails — so a failed cancel can no longer drop coverage without replacing it. Both `handleReassign`s now take `assignmentSource` and delegate to it. That keeps the two tabs from drifting apart on this endpoint again, which is the same class of drift this PR already exists to remove.

**Yes, `PeriodsToFillTab` too.** It needed the same substitution plus `async`/`await`, since its version was a bare `.then()` chain. One incidental improvement falls out: it previously chained the POST unconditionally and ignored every response, so it both swallowed errors *and* fired the assign after a failed cancel. It now surfaces the error and stops, matching `AbsentTeachersTab`.

**Test.** The unit suite is `environment: "node"` and `.ts`-only so it cannot mount the tabs, and the reachable contract here is client-side fetch routing, not a server endpoint — a `tests/api/` test would not have covered the actual defect. Extracting `reassignCoverage()` is what made the branch decision unit-testable: `src/__tests__/coverageActions.test.ts` (8 tests, mocked `fetch`) asserts PREFERRED cancels via `PATCH /api/{slug}/preferred-relief-assignments/{id}` then POSTs, INTERNAL/`null` cancels via the internal `DELETE`, and no assign is attempted when the cancel fails. I verified it fails against the old hardcoded-DELETE behaviour before the fix (`expected "/api/relief-assignments/cover-1" to be "/api/acme/preferred-relief-assignments/cover-1"`), so it is a real regression test rather than a restatement.

**Verification:** `npx tsc --noEmit` exit 0, clean. `npm test` 86 files / 891 passed, 1 skipped. `npm run build` exit 0, `✓ Generating static pages using 11 workers (93/93)`, zero `⨯`. `npm audit` 0 vulnerabilities via the pre-push hook. The integration suite was not run — `reliefcher-test-db` is not up locally and I did not start it; CI covers it.

Nothing owned by #630 was touched (no schema, no `preferred-relief-assignments/route.ts`, no `relief-assignments/route.ts`). And confirming your seam note: this failure is upstream of the P2002 constraint, so the two fixes are independent.

Issue #455, opened 2026-07-03
bug: preferred-relief-assignment date not normalized to midnight UTC — breaks per-slot invariant and dashboard visibility


guangshinhaha:
## Summary
`POST /api/[slug]/preferred-relief-assignments` stores the `date` **verbatim** (`z.coerce.date()`) instead of normalizing to midnight UTC like the rest of the system. A non-midnight value never matches the dashboard's exact-equality date query and bypasses all three duplicate/race guards, breaking the "one NOTIFIED row per slot per day" invariant.

## Location
`src/app/api/[slug]/preferred-relief-assignments/schema.ts:23` (+ route.ts create/checks)

```ts
date: z.coerce.date(),   // accepts any ISO datetime, stored as-is
```

Everything else normalizes to midnight UTC: the internal route uses `new Date(dateStr + "T00:00:00.000Z")` (`relief-assignments/route.ts:37`), and the dashboard reads with exact equality `where: { date: dateUTC }` (`dashboard/route.ts`). The `@@unique([timetableEntryId, date, status])` race guard, the `existingInternal` check, and the concurrent-period check all compare exact timestamps.

## Failure scenario
A client sends `date: "2026-07-03T00:00:00.000+08:00"` (a plausible SGT-browser value, or any caller via the `from-preferred` shim which forwards `date` verbatim). The row is stored at `2026-07-02T16:00Z`. It never appears on the July-3 dashboard (slot still shows uncovered), doesn't collide with the unique constraint, and a second row can be created for the same slot — double-booking with a phantom `assignmentCount` increment. Same-day duplicates at different times also bypass all three 409 guards.

## Fix direction
Normalize `date` to midnight UTC at the schema/route boundary (`YYYY-MM-DD` → `…T00:00:00.000Z`), consistent with `relief-assignments` and the dashboard read.

## Confidence
Behavior verified; real-world trigger frequency depends on caller date formatting.

---
*Filed from the 2026-07 full-app audit. Verified against code.*

PR #630, opened 2026-08-05
fix(relief): external relief can only be re-assigned once — second cancel 500s and wedges the slot

guangshinhaha:
Fixes the **external (preferred) half** of #623. The internal half is PR #629.

⚠️ **Contains a migration.** See "Deploy notes" at the bottom — there's a backfill script to run after it lands.

## Summary

Swapping the external relief teacher on a slot worked exactly once. The second cancel returned **500**, and left the assignment stuck `NOTIFIED` with no way to clear it: there's no `DELETE` route, `PATCH → CANCELLED` is the only exit, and it was the broken thing. The absence report then refused to delete too (`409 COVERAGE_EXISTS`), so the slot could only be freed by hand in the database.

Reproduced against the test env before the fix:

```
assign A -> 201 | cancel A -> 200 | assign B -> 201 | cancel B -> 500
delete absence report -> 409 COVERAGE_EXISTS
```

## Root cause

`PreferredReliefAssignment` carried:

```prisma
// Only one NOTIFIED row per (timetableEntry × date). CANCELLED
// rows don't collide so KPs can cycle through preferred teachers ...
@@unique([timetableEntryId, date, status])
```

**The comment was wrong.** A composite unique constrains every status equally — it permits one `NOTIFIED` row **and one `CANCELLED` row** per slot. It is not a partial index on `status = 'NOTIFIED'`.

| step | action | row becomes | result |
|---|---|---|---|
| 1 | assign A | `(E, D, NOTIFIED)` | 201 |
| 2 | cancel A | `(E, D, CANCELLED)` | 200 |
| 3 | assign B | `(E, D, NOTIFIED)` | 201 |
| 4 | cancel B | `(E, D, CANCELLED)` — A already holds it | **P2002 → 500** |

## Approach

What the model needs is *"at most one **live** assignment per (timetableEntry × date), keeping cancellation history"*. A partial unique index (`WHERE status = 'NOTIFIED'`) says that directly, but it isn't expressible in `schema.prisma`, and hand-writing migration SQL is out per CLAUDE.md.

So: a nullable key. `activeSlotKey` (`"{timetableEntryId}|{YYYY-MM-DD}"`) is set while a row is live and nulled on cancel, with `@@unique([activeSlotKey])`. Postgres treats NULLs in a unique index as distinct, so any number of cancelled rows coexist while still allowing only one live row per slot.

## Changes

- **`prisma/schema.prisma`** — `activeSlotKey String?` + `@@unique([activeSlotKey])`, replacing the old triple-key unique. Migration `20260805080151_preferred_relief_active_slot_key` generated by `prisma migrate dev`; the SQL is Prisma's, not hand-authored.
- **`src/lib/preferredReliefSlot.ts`** — `activeSlotKey()`, the single place the key format is defined. It's persisted and the backfill reproduces it, so the format is a contract.
- **All three writers updated:**
  1. `preferred-relief-assignments/route.ts` — sets the key on create.
  2. `preferred-relief-assignments/[id]/route.ts` — nulls it on cancel.
  3. **`relief-assignments/route.ts`** — cancels a preferred assignment when internal staff supersede it. Easy to miss (I only found it by grepping for writers); leaving its key set would have blocked every later external claim on that slot.
- **P2002 → 409** on the cancel route, so a constraint failure returns something legible instead of the bare 500 this bug surfaced as.
- **`scripts/backfill-preferred-relief-slot-keys.ts`** (+ `npm run backfill:preferred-slot-keys`) — dry-run/apply, idempotent, and refuses with a report if it finds a slot that somehow already has two live rows.
- **`tests/api/helpers.ts`** — `createPreferredReliefAssignment` now mirrors the routes (live rows hold the key, cancelled rows don't), so helper-built rows participate in the unique rather than passing where prod would fail.

## Test plan

**New integration test** — `tests/api/preferred-relief-assignments.test.ts` cycles cancel→reassign **three times** and asserts every cancel is 200, exactly one row stays live, and only that row holds a key. The existing test stopped after **one** cycle, which is precisely why this shipped broken. Plus a negative control asserting a second *live* assignment on the same slot is still rejected with 409 — releasing the key on cancel must not weaken the guarantee the unique exists for.

**New unit tests** — 4 cases pinning the key format, including that a timestamp drifted off midnight still maps to the same slot-day (the POST schema normalises to midnight UTC per #455, and a second key for one day would mean two live assignments).

```
npm test        →  86 files, 887 passed | 1 skipped
tsc --noEmit    →  exit 0
npm run build   →  exit 0
eslint          →  clean on changed files
npm run test:api →  99 files, 964 tests, 924 passed, 40 failed
```

**On those 40 API failures — they are pre-existing on `main`, not from this PR.** I ran the same suite on clean `main` and got a **byte-identical** list of 40 failures (962 tests there vs 964 here — the 2 extra are this PR's, both passing). They're concentrated in `absence-reports`, `admin`, `auth-otp-bypass`, `leads-bulk-import` and `cron-*`, none of which this PR touches.

While chasing that I found a likely root cause for a chunk of them, **unrelated to this PR and not fixed here**: [`absence-reports/route.ts:47`](src/app/api/absence-reports/route.ts#L47) defaults `reason = "MEDICAL_LEAVE"`, which is not a member of the `AbsenceReason` enum (`SICK_LEAVE_MC`, `SICK_LEAVE_NO_MC`, `TIME_OFF`, …). Any request omitting `reason` therefore fails `VALID_REASONS.has(reason)` and 400s. Worth its own issue.

## Deploy notes

1. The migration adds `activeSlotKey` as NULL for every existing row and moves the unique onto it. Rows written before the migration are therefore **not** covered by the new unique until they're next written.
2. Current data is consistent — the old constraint guaranteed at most one `NOTIFIED` row per slot — so this is a narrow window, not a correctness break. But until the backfill runs, a new concurrent claim against a pre-existing row wouldn't be caught by the DB.
3. After the migration deploys, run once per environment:

```bash
DIRECT_URL=... npm run backfill:preferred-slot-keys -- --dry-run   # review
DIRECT_URL=... npm run backfill:preferred-slot-keys -- --apply
```

`prisma migrate dev` generates DDL, not DML, and I'm not hand-editing the migration file — hence a script rather than folding the UPDATE into the migration.

Also: this unsticks the row my #623 repro left on `maia-pri-1` (`cmsfr7i8d003i195r5t3h7ch9`) — once deployed, cancelling it through the UI will work, and the absence report becomes deletable.


scripts/backfill-preferred-relief-slot-keys.ts
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  let updated = 0;
+  for (const row of pending) {
+    const key = activeSlotKey(row.timetableEntryId, row.date);
+    if (dryRun) {
+      console.log(`SET  ${row.id} -> ${key}`);
+      continue;
+    }
+    await prisma.preferredReliefAssignment.update({
```
guangshinhaha: **[High] Backfill can stamp a slot key onto a row that was cancelled mid-run, permanently wedging that slot**

```ts
const pending = await prisma.preferredReliefAssignment.findMany({
  where: { status: "NOTIFIED", activeSlotKey: null },   // read
  ...
});
...
for (const row of pending) {
  const key = activeSlotKey(row.timetableEntryId, row.date);
  ...
  await prisma.preferredReliefAssignment.update({
    where: { id: row.id },        // ← no status/key re-check between read and write
    data: { activeSlotKey: key },
  });
```

**Problem:** The deploy notes say to run this "once per environment" *after* the migration deploys — i.e. against a live app. The read and the write are separated by an unbounded window (one round-trip per row, no transaction, no re-check).

Sequence:
1. Backfill reads row `X` — `NOTIFIED`, key `NULL`.
2. A KP hits `PATCH .../preferred-relief-assignments/X` ("Mark uncovered"), or an admin assigns internal staff over it via `relief-assignments` POST. Either path sets `status = CANCELLED, activeSlotKey = NULL`.
3. Backfill's `update` fires unconditionally and writes `activeSlotKey = "{entry}|{date}"` back onto the now-`CANCELLED` row.

The slot is now held by a cancelled row. Every subsequent external claim on that slot hits P2002 and gets a 409 ("another teacher was just assigned to this period" — the winner lookup finds no `NOTIFIED` row, so the message is wrong too). And there is **no route that can clear it**: `PATCH` early-returns at `[id]/route.ts:56` on `status === "CANCELLED"` without touching `activeSlotKey`, and there is no `DELETE`. That is exactly the "stuck slot, only fixable by hand in the DB" failure this PR exists to remove — reintroduced by the tool meant to close it out.

**Suggestion:** Make the write conditional on the row still being in the state that was read. `updateMany` is a no-op (`count: 0`) if it isn't, so a concurrent cancel is simply skipped:

```ts
let updated = 0;
let skipped = 0;
for (const row of pending) {
  const key = activeSlotKey(row.timetableEntryId, row.date);
  if (dryRun) {
    console.log(`SET  ${row.id} -> ${key}`);
    continue;
  }
  // Re-assert the precondition at write time — the row may have been
  // cancelled since the read, and stamping a key onto a CANCELLED row
  // wedges the slot with no route able to release it.
  const { count } = await prisma.preferredReliefAssignment.updateMany({
    where: { id: row.id, status: "NOTIFIED", activeSlotKey: null },
    data: { activeSlotKey: key },
  });
  if (count === 1) updated += 1;
  else {
    skipped += 1;
    console.log(`SKIP ${row.id} — no longer live (cancelled since read)`);
  }
}
```

Worth also noting in the deploy notes that the safest run is with no KP traffic, since the dry-run output can go stale the same way.

---

src/app/api/[slug]/preferred-relief-assignments/route.ts
```diff
@@ -169,6 +170,9 @@ export async function POST(req: Request, { params }: Params) {
         date: input.date,
         status: "NOTIFIED",
         createdByUserId: session.user.id,
+        // Holds the slot while this row is live; nulled on cancel so the
+        // unique lets the KP cycle through teachers indefinitely (#623).
+        activeSlotKey: activeSlotKey(input.timetableEntryId, input.date),
```
Comment (guangshinhaha) [inline src/app/api/[slug]/preferred-relief-assignments/route.ts:203]: **[High] No application-level same-slot guard — the DB unique is the only thing preventing double external coverage, and it does not cover pre-backfill rows**

```ts
// Race shape 3 checks a *different* slot at the same period...
const concurrent = await prisma.preferredReliefAssignment.findFirst({
  where: {
    preferredTeacherId: input.preferredTeacherId,
    date: input.date,
    status: "NOTIFIED",
    timetableEntryId: { not: input.timetableEntryId },  // ← same slot explicitly excluded
    timetableEntry: { periodId: entry.periodId },
  },
});
...
const created = await prisma.preferredReliefAssignment.create({
  data: {
    ...
    activeSlotKey: activeSlotKey(input.timetableEntryId, input.date),  // ← sole guard
  },
});
```

**Problem:** There is no pre-check for "this slot already has a live preferred assignment". Race shape 1 covers internal `ReliefAssignment`; race shape 3 explicitly excludes the same `timetableEntryId`. So the *only* thing stopping a second live external assignment on one slot is the `activeSlotKey` unique — and by this PR's own deploy notes, every row written before the migration carries `activeSlotKey = NULL`, and Postgres treats NULLs in a unique index as distinct.

Concretely, in the window between the migration deploying and the backfill being run (and indefinitely if the backfill is ever forgotten on an environment):

- Slot S has a pre-migration `NOTIFIED` row for teacher A, `activeSlotKey = NULL`.
- A KP assigns teacher B to slot S. The create succeeds — the unique sees `NULL` vs `"S|2026-08-10"`, no collision.
- Slot S now has **two live rows**. No error, no 409, nothing in the logs.

Downstream this is silent bad data rather than a crash: `absenceCoverageGuard` counts both, `getPreferredAssignmentCountsForWindow` bills both teachers a shift, and the dashboard renders whichever `findFirst` returns — so two external teachers can be told to cover the same period.

This is also the precondition for the backfill's `P2002` failure mode (see the comment on the conflict pre-check).

**Suggestion:** Add a same-slot pre-check before the create. It is independent of `activeSlotKey` state, so it holds regardless of whether the backfill has run, and it produces a legible message instead of relying on the P2002 catch:

```ts
// Race shape 0: this slot already has a live preferred assignment.
// Checked in the app rather than leaning on the `activeSlotKey` unique
// alone — rows written before the #623 migration carry a NULL key and
// are therefore not covered by it until the backfill runs.
const existingPreferred = await prisma.preferredReliefAssignment.findFirst({
  where: {
    timetableEntryId: input.timetableEntryId,
    date: input.date,
    status: "NOTIFIED",
  },
  include: { preferredTeacher: { select: { name: true } } },
});
if (existingPreferred) {
  return NextResponse.json(
    {
      error: `${existingPreferred.preferredTeacher.name} is already covering this period. Mark it uncovered first.`,
      code: "ALREADY_ASSIGNED_PREFERRED",
    },
    { status: 409 },
  );
}
```

The existing P2002 catch stays as the race closer; this makes the common case deterministic and removes the dependency on backfill ordering.

---

scripts/backfill-preferred-relief-slot-keys.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  // Two live rows on one slot would mean the pre-migration data already
+  // violated the old constraint. Refuse rather than fail halfway through, and
+  // report the conflicts so they can be resolved by hand.
+  const byKey = new Map<string, string[]>();
+  for (const row of pending) {
+    const key = activeSlotKey(row.timetableEntryId, row.date);
+    byKey.set(key, [...(byKey.get(key) ?? []), row.id]);
+  }
+  const conflicts = [...byKey.entries()].filter(([, ids]) => ids.length > 1);
```
guangshinhaha, line 99: **[Medium] Conflict pre-check only compares NULL-key rows against each other, so the dry-run can report "clean" and `--apply` still dies with P2002 halfway through**

```ts
const byKey = new Map<string, string[]>();
for (const row of pending) {                       // pending = { status: NOTIFIED, activeSlotKey: null }
  const key = activeSlotKey(row.timetableEntryId, row.date);
  byKey.set(key, [...(byKey.get(key) ?? []), row.id]);
}
const conflicts = [...byKey.entries()].filter(([, ids]) => ids.length > 1);  // ←
```

**Problem:** The map is built from `pending` only. A key that is already held by a *different* live row — one written post-migration by the POST route, which does set the key — is invisible to this check.

That shape is reachable exactly because the POST route has no same-slot pre-check (see the other comment): pre-migration row A on slot S has a NULL key, a KP then assigns teacher B to S and B's row gets `activeSlotKey = "S|2026-08-10"`. Now:

- `--dry-run` prints `SET A -> S|2026-08-10`, `Dry run complete — 1 row(s) would be stamped.`, exit 0. Looks safe.
- `--apply` calls `update` on A, Postgres rejects it with `P2002`, the top-level `.catch` prints a raw Prisma stack trace, `process.exitCode = 1`.

Two problems with that outcome. The dry run gave false assurance before a production run, which is the one thing a dry run exists to prevent. And because there is no transaction wrapping the loop, whichever rows were processed before the throw stay committed while the rest do not — recoverable via re-run given idempotency, but the operator sees a stack trace with no statement of how far it got.

**Suggestion:** Build the conflict map from *all* live rows, not just the ones needing a stamp, and wrap the writes so partial application is reported rather than inferred:

```ts
// Every live row on a slot — including ones that already hold a key —
// so a pre-existing holder is detected before any write is attempted.
const allLive = await prisma.preferredReliefAssignment.findMany({
  where: { status: "NOTIFIED" },
  select: { id: true, timetableEntryId: true, date: true, activeSlotKey: true },
});

const byKey = new Map<string, string[]>();
for (const row of allLive) {
  const key = row.activeSlotKey ?? activeSlotKey(row.timetableEntryId, row.date);
  byKey.set(key, [...(byKey.get(key) ?? []), row.id]);
}
const conflicts = [...byKey.entries()].filter(([, ids]) => ids.length > 1);
```

and in the catch, say how much landed:

```ts
main().catch((err) => {
  console.error(`Failed after ${updated} row(s) stamped — re-run is safe (idempotent).`);
  console.error(err);
  process.exitCode = 1;
});
```

---

prisma/schema.prisma
```diff
@@ -1208,19 +1208,32 @@ model PreferredReliefAssignment {
...
+  // comment claimed CANCELLED rows didn't collide. They did: a composite
+  // unique constrains every status equally, so it allowed one NOTIFIED row
+  // *and one CANCELLED row* per slot. A KP could therefore swap external
+  // relief exactly once — the second cancel hit a unique violation, 500'd,
+  // and left the slot stuck NOTIFIED with no way out (#623).
+  //
+  // Postgres treats NULLs in a unique index as distinct, so nulling the key
+  // on cancel lets any number of cancelled rows coexist while still allowing
+  // only one live row per slot.
+  @@unique([activeSlotKey])
```
guangshinhaha, line 1236: **[Medium] Dropping the composite unique leaves no index on `timetableEntryId` at all — the internal-assignment hot path loses its index**

```prisma
-  @@unique([timetableEntryId, date, status])   // ← removed; was the only index on timetableEntryId
+  @@unique([activeSlotKey])
   @@index([schoolId, date])
   @@index([preferredTeacherId, date])
   @@index([absenceReportId])
```

**Problem:** The old `@@unique([timetableEntryId, date, status])` was doing double duty — it enforced the constraint *and* was the only index whose leading column is `timetableEntryId`. Nothing in the replacement set covers it (`activeSlotKey` is opaque, and the three `@@index` entries lead with `schoolId`, `preferredTeacherId`, `absenceReportId`).

The queries that used it are still there and still filter the same way:

- `src/app/api/relief-assignments/route.ts:310` — `findFirst({ where: { timetableEntryId, date, status: "NOTIFIED" } })`. This runs on **every** internal relief assignment POST, including each write in the auto-assign confirm loop. It has no `schoolId` in the predicate, so it can't fall back to `PreferredReliefAssignment_schoolId_date_idx` either — it degrades to a sequential scan over the whole table, across all tenants.
- `src/app/api/[slug]/preferred-relief-assignments/route.ts:200` — the winner lookup in the P2002 catch, same shape.

Small today; `PreferredReliefAssignment` accumulates one CANCELLED audit row per swap and is never pruned, and this PR's whole point is to make swapping unlimited — so the table is now designed to grow monotonically under exactly the traffic that triggers the scan.

**Suggestion:** Re-add the index explicitly (a plain `@@index`, so it carries no constraint semantics — then `prisma migrate dev` regenerates the SQL):

```prisma
  @@unique([activeSlotKey])
  // Retains the index the old `@@unique([timetableEntryId, date, status])`
  // incidentally provided. `relief-assignments` POST looks up the live
  // preferred row by (timetableEntryId, date) on every internal assignment.
  @@index([timetableEntryId, date])
  @@index([schoolId, date])
  @@index([preferredTeacherId, date])
  @@index([absenceReportId])
```

---

src/app/api/relief-assignments/route.ts
```diff
@@ -360,6 +360,10 @@ export async function POST(request: NextRequest) {
             status: "CANCELLED",
             cancelledAt: new Date(),
             cancelledByUserId: session.user.id,
+            // Same slot release as the explicit cancel route (#623) — without
+            // it, assigning internal staff over an external teacher would
+            // leave the slot key held and block any later external claim.
+            activeSlotKey: null,
```
guangshinhaha, line 366: **[Medium] The one new production line in this file has no covering test — deleting it would leave the suite green**

```ts
if (supersededPreferred) {
  await tx.preferredReliefAssignment.update({
    where: { id: supersededPreferred.id },
    data: {
      status: "CANCELLED",
      cancelledAt: new Date(),
      cancelledByUserId: session.user.id,
      activeSlotKey: null,   // ← the only new line here, and nothing asserts it
    },
  });
}
```

**Problem:** The PR body flags this writer as the easy-to-miss one ("leaving its key set would have blocked every later external claim on that slot") — and it's the one path with no regression cover.

The existing test for this path, `tests/api/relief-assignments.test.ts:643`, builds its preferred row with a raw `testPrisma.preferredReliefAssignment.create({ ... })` and no `activeSlotKey`, so the row starts at `NULL`. It then asserts only `status === "CANCELLED"` and `cancelledAt !== null`. Remove `activeSlotKey: null` from the route and the test still passes, because the field was already `NULL` going in.

This is the exact failure mode the helper change in `tests/api/helpers.ts` was made to prevent — "helper-built rows wouldn't participate in the `activeSlotKey` unique and tests would pass where prod fails" — but three call sites still bypass the helper and construct rows directly: `tests/api/relief-assignments.test.ts:643`, `tests/api/autoAssign.test.ts:544`, and `tests/api/preferred-relief-assignments.test.ts:440`. All three produce NULL-key live rows that prod would never write.

**Suggestion:** Route those three through the updated factory, and assert the release at `relief-assignments.test.ts:643`:

```ts
// tests/api/relief-assignments.test.ts — use the factory so the row holds
// its slot the way a route-written row does.
const preferredAssignment = await createPreferredReliefAssignment(school.id, {
  absenceReportId: absence.id,
  timetableEntryId: entry.id,
  preferredTeacherId: preferred.id,
  date: DATE_UTC,
});
expect(preferredAssignment.activeSlotKey).not.toBeNull();  // precondition

// ... after the internal POST:
const pref = await testPrisma.preferredReliefAssignment.findUnique({
  where: { id: preferredAssignment.id },
});
expect(pref?.status).toBe("CANCELLED");
expect(pref?.cancelledAt).not.toBeNull();
// Superseding must release the slot, or no external teacher can claim it again.
expect(pref?.activeSlotKey).toBeNull();
```

Then follow it with a second external assignment on the same slot asserting 201, which is the behaviour the line actually protects.

---

guangshinhaha:
## Code Review Summary

| Severity | Count |
|----------|-------|
| Critical | 0 |
| High     | 2 |
| Medium   | 3 |
| Low      | 0 |

The root-cause diagnosis is right and the nullable-key approach is the correct way to get partial-unique semantics without hand-written migration SQL. The findings cluster on one seam: **`activeSlotKey` is now the sole guarantor of "one live assignment per slot", and it is only as good as the rows that carry it.** Both High findings fall out of that — the POST route has no independent same-slot check, and the backfill that closes the NULL-key gap can itself write a key onto a row that was cancelled mid-run (which wedges the slot with no route able to release it, since `PATCH` early-returns on `CANCELLED` and there is no `DELETE`).

**On the backfill specifically**, against the criteria it needs to meet before a production run:

- **Idempotency on re-run** — good. `where: { status: "NOTIFIED", activeSlotKey: null }` narrows to exactly the unprocessed set, and the key is deterministic.
- **Multi-tenancy** — deliberately global (no `schoolId` scope), which is correct for a one-shot per-environment backfill. `timetableEntryId` is a cuid so the key can't collide across schools.
- **Rows silently skipped** — none unintentionally; `CANCELLED` rows staying NULL is the design, not an omission, and the doc comment says so.
- **Secrets** — clean. No default connection string, exits early if neither `DIRECT_URL` nor `DATABASE_URL` is set, and logs only the parsed host.
- **Partial failure** — the weak spot. No transaction, no re-check at write time, and a raw stack trace on abort with no statement of how far it got. See the two comments on the script.

Nothing in this diff touches a flagged-OFF area — preferred/external relief is in-pilot scope — and no CLAUDE.md migration rule is violated: `20260805080151` is Prisma-generated DDL (the `DROP INDEX` name matches what `20260523000000` created via `CREATE UNIQUE INDEX`, so it will apply cleanly), and the DML correctly lives in a script rather than being hand-edited into the migration.

**Cross-PR seam with #629:** checked — #629 touches only `src/app/[slug]/dashboard/AbsentTeachersTab.tsx`, so there is no file overlap and no shared contract between the two halves. They can land in either order.

## What Looks Good

- **Grepping for writers rather than fixing the reported one.** Catching `relief-assignments/route.ts:357` — the internal-supersede path — is the difference between a fix and a half-fix; leaving that key set would have reproduced #623 through a different door.
- **The test written against the bug's actual shape.** The old test cycled once and passed; cycling three times and asserting exactly one row holds a key is what makes this class of regression detectable at all. The negative control ("second *live* assignment still 409s") is the right companion — it pins that releasing the key didn't dissolve the guarantee.
- **Treating the key format as a contract and testing it as one.** It's persisted and independently reproduced by the backfill, so the drifted-off-midnight case in `preferredReliefSlot.test.ts` is guarding a real seam between two codebases, not padding coverage.
- **Correcting the schema comment instead of just the schema.** The old comment asserted the exact false invariant that caused the bug; replacing it with the reasoning and the incident reference means the next reader can't re-derive the same mistake.

---

scripts/backfill-preferred-relief-slot-keys.ts
guangshinhaha: Fixed in 4d01658 — agreed, and this was the worst of the five: the backfill could recreate the exact wedge it was written to clear.

The write is now an `updateMany` that re-asserts the precondition the read assumed:

```ts
const { count } = await prisma.preferredReliefAssignment.updateMany({
  where: { id: row.id, status: "NOTIFIED", activeSlotKey: null },
  data: { activeSlotKey: key },
});
if (count === 1) updated += 1;
else { skipped += 1; console.log(`SKIP ${row.id} — no longer live (cancelled since read)`); }
```

A row cancelled between the `findMany` and the write reports `count: 0` and is skipped rather than stamped, so a CANCELLED row can never end up holding a key. Skips are counted and included in the final line (`N row(s) stamped, M skipped (changed since read)`) so an operator sees drift instead of it passing silently.

Also took the deploy-notes point — the module header now says the cleanest run is one with no KP traffic, since dry-run output can go stale between the review and the apply.

src/app/api/[slug]/preferred-relief-assignments/route.ts
Comment (guangshinhaha) [reply src/app/api/[slug]/preferred-relief-assignments/route.ts:203]: Fixed in 4d01658. The analysis holds — correctness was resting on a unique index that structurally cannot see the rows it was meant to protect, and the failure mode is silent bad data rather than an error.

Added race shape 0 before the create, essentially as suggested:

```ts
const existingPreferred = await prisma.preferredReliefAssignment.findFirst({
  where: { timetableEntryId: input.timetableEntryId, date: input.date, status: "NOTIFIED" },
  include: { preferredTeacher: { select: { name: true } } },
});
if (existingPreferred) {
  return NextResponse.json(
    { error: `${existingPreferred.preferredTeacher.name} is already covering this period. Mark it uncovered first.`,
      code: "ALREADY_ASSIGNED_PREFERRED" },
    { status: 409 },
  );
}
```

It reads `status`, not `activeSlotKey`, so it holds regardless of whether the backfill has run. The P2002 catch stays as the concurrent-race closer.

Two knock-on changes worth flagging:

- `tests/api/preferred-relief-assignments.test.ts:122` (the sequential double-claim) now asserts `ALREADY_ASSIGNED_PREFERRED` instead of `STALE_DASHBOARD`. That test was never actually concurrent — the first row is committed before the second POST — so the deterministic pre-check is what should answer it. `STALE_DASHBOARD` remains correct for a genuine race. Renamed the test and left a comment recording the distinction.
- Added a covering test: seed a live row, NULL its `activeSlotKey` to reproduce a pre-migration row, then claim the slot with a different teacher — expects 409 `ALREADY_ASSIGNED_PREFERRED` and asserts exactly one live row remains. That test fails without the pre-check.

Note on verification: this test is in the integration suite, which I was not able to run — see the summary comment for why.

scripts/backfill-preferred-relief-slot-keys.ts
guangshinhaha: Fixed in 4d01658. The map is now built from every live row rather than only the ones needing a stamp, so a slot already held by a post-migration row is visible to `--dry-run`:

```ts
const allLive = await prisma.preferredReliefAssignment.findMany({
  where: { status: "NOTIFIED" },
  select: { id: true, timetableEntryId: true, date: true, activeSlotKey: true },
});
const byKey = new Map<string, string[]>();
for (const row of allLive) {
  const key = row.activeSlotKey ?? activeSlotKey(row.timetableEntryId, row.date);
  byKey.set(key, [...(byKey.get(key) ?? []), row.id]);
}
```

Using the stored key when present and deriving it otherwise means the two collision shapes both surface: a NULL-key row against a keyed holder, and two NULL-key rows against each other.

The catch now reports how far a failed run got, since the loop is deliberately not transactional:

```
Failed after N row(s) stamped — re-running is safe (idempotent).
```

`updated` moved to module scope so the top-level handler can read it.

Worth noting the two fixes compound: with the same-slot pre-check now on the POST route, the shape that made this reachable (a second live row on a slot already held by a pre-migration row) can no longer be created in the first place. This check is the belt to that suspenders — it still matters for data written before either fix ships.

prisma/schema.prisma
guangshinhaha: Agreed, and the schema edit is in 4d01658 — but **leaving this thread open: the migration is not generated yet and needs a human step.**

`prisma/schema.prisma` now carries the index as suggested:

```prisma
@@unique([activeSlotKey])
// Retains the index the old `@@unique([timetableEntryId, date, status])`
// incidentally provided — `activeSlotKey` is opaque and the other indexes
// lead with schoolId/preferredTeacherId/absenceReportId, so nothing else
// covers a lookup by timetable entry. `relief-assignments` POST resolves
// the live preferred row by (timetableEntryId, date) on every internal
// assignment, with no schoolId in the predicate — without this it is a
// sequential scan over a table this change deliberately lets grow (one
// CANCELLED audit row per swap, never pruned).
@@index([timetableEntryId, date])
@@index([schoolId, date])
@@index([preferredTeacherId, date])
@@index([absenceReportId])
```

**Required before merge — someone with the local dev DB running must generate the migration:**

```bash
docker compose up -d
npx prisma migrate dev --name preferred_relief_timetable_entry_date_index
```

Per CLAUDE.md I am not permitted to hand-write migration SQL or pick migration timestamps, so I made the schema edit only. Without that command the branch has schema drift: `schema.prisma` declares an index that no migration creates, so the index will be missing in every deployed environment even though Prisma Client and the type-checker are perfectly happy. The API test suite hides this too — `tests/api/setup.ts` uses `prisma db push`, which materialises the index straight from the schema without a migration.

Resolve this thread once the migration file exists.

src/app/api/relief-assignments/route.ts
guangshinhaha: Fixed in 4d01658 — the point about the assertion being vacuous was exactly right: the row was already NULL going in, so it proved nothing.

All three raw call sites now go through the factory:
- `tests/api/relief-assignments.test.ts:643`
- `tests/api/autoAssign.test.ts:544`
- `tests/api/preferred-relief-assignments.test.ts:440`

At the supersede test, the precondition is now explicit and the release is asserted:

```ts
const preferredAssignment = await createPreferredReliefAssignment(school.id, { ... });
expect(preferredAssignment.activeSlotKey).not.toBeNull();   // precondition — this is what was missing

// ... after the internal POST:
expect(pref?.status).toBe("CANCELLED");
expect(pref?.cancelledAt).not.toBeNull();
expect(pref?.activeSlotKey).toBeNull();
```

Plus the behavioural follow-up you asked for, adapted: a second *external* POST on that slot cannot return 201 here, because the internal assignment now holds it and race shape 1 (`ALREADY_ASSIGNED_INTERNAL`) refuses first. So the re-claim goes through the factory instead, which stamps the key and therefore throws P2002 if the supersede path left the old key set:

```ts
const reclaimed = await createPreferredReliefAssignment(school.id, { /* same entry + date */ });
expect(reclaimed.activeSlotKey).toBe(preferredAssignment.activeSlotKey);
```

Deleting `activeSlotKey: null` from the route now fails the suite twice over — once on the direct assertion, once on the re-claim.

Caveat on verification: these are integration tests and I could not run `npm run test:api` in this environment — details in the summary comment. Worth a run before merge.

guangshinhaha:
## Review fixes — 4d01658

All five findings were valid on inspection; none were refuted. Fixed in one commit.

| # | Finding | Status |
|---|---|---|
| 1 | Backfill can stamp a key onto a row cancelled mid-run | Fixed — `updateMany` re-asserts the precondition, skips counted |
| 2 | No application-level same-slot guard | Fixed — race shape 0 pre-check, independent of backfill state |
| 3 | Conflict pre-check only compares NULL-key rows | Fixed — map built from all live rows; catch reports rows written |
| 4 | Dropping the composite unique loses the `timetableEntryId` index | Schema edited — **migration still to be generated, see below** |
| 5 | `activeSlotKey: null` on the supersede path has no covering test | Fixed — 3 call sites routed through the factory + assertions |

### Verification

```
npx tsc --noEmit    exit 0, no output
npm test            86 files, 887 passed | 1 skipped, exit 0
npm run build       exit 0 — "Compiled successfully", "Generating static pages (93/93)", no ⨯
npm audit           found 0 vulnerabilities
```

Pre-push hooks ran on the push (audit + tsc + unit suite), not bypassed.

### ⚠️ `npm run test:api` was NOT run — two blockers

Stating this plainly because the changed tests are integration tests, so **the finding-5 assertions and the new finding-2 test are unverified by execution**. They type-check (`tsconfig` includes `**/*.ts`, and `tsc` is clean) but have not been executed.

1. Port 5433 on this machine is held by an unrelated project's Postgres (`kopi-db_test-1`, user `nurse`), not `reliefcher-test-db`. I stood up a dedicated `postgres:15` on a spare port and repointed the test config at it temporarily (reverted afterwards — not in the commit).
2. That run then failed in `tests/api/setup.ts`, which shells out to `prisma db push`. Prisma refuses schema-mutating commands when it detects an AI agent, requiring explicit user consent via `PRISMA_USER_CONSENT_FOR_DANGEROUS_AI_ACTION`. No such consent exists, so I stopped rather than work around the guardrail.

**Please run `npm run test:api` locally before merge.** Expected deltas beyond the new tests: `preferred-relief-assignments.test.ts` "returns 409 when a second KP claims the same slot+date" now asserts `ALREADY_ASSIGNED_PREFERRED` rather than `STALE_DASHBOARD` — the sequential case is answered by the new deterministic pre-check, with the P2002 catch still covering genuine races.

### 🔴 Human step required before merge — migration for finding 4

`prisma/schema.prisma` declares `@@index([timetableEntryId, date])` but **no migration creates it**. Per CLAUDE.md I cannot hand-write migration SQL or choose timestamps:

```bash
docker compose up -d
npx prisma migrate dev --name preferred_relief_timetable_entry_date_index
```

Until that runs, the branch has schema drift and the index will be absent in every deployed environment — silently, since Prisma Client and `tsc` are unaffected and the API suite uses `db push` (which materialises the index straight from the schema). That thread is left unresolved deliberately; the other four are resolved.

prisma/schema.prisma
guangshinhaha: Migration generated and pushed — `4bbd35f`.

`prisma migrate dev` produced `20260805093150_preferred_relief_timetable_entry_date_index`, containing exactly:

```sql
CREATE INDEX "PreferredReliefAssignment_timetableEntryId_date_idx" ON "PreferredReliefAssignment"("timetableEntryId", "date");
```

One statement and nothing else, which also confirms the branch carried no other schema drift. Generated by Prisma against a throwaway local dev DB, not hand-written, per the CLAUDE.md migrations rule.

Worth recording why this was easy to miss: `tsc` and Prisma Client are both unaffected by a missing migration, and `tests/api/setup.ts` uses `prisma db push`, which materialises the index straight from `schema.prisma` — so the integration suite would have gone green against an index that did not exist in any real environment.

`npm run test:api` now run on this branch: 40 failures, byte-identical to the `main` baseline (13 files, same test names — pre-existing date-window rot, not from this PR). 925 passed vs main's 922, i.e. the 3 tests this PR adds all pass and nothing regressed.

---

Issue #642, opened 2026-08-07
feat(analytics): pilot metrics dashboard in Looker Studio over GA4


guangshinhaha:
## User story

As the ReliefCher product team, I want all seven pilot scorecard metrics in one Looker Studio report filtered by school and date range, so that we can report against the pilot success metrics without asking schools to hand-count anything each day.

## Background

Split from #639. This is the **dashboard** slice. The GA event instrumentation it reads is #640; the surveys behind metrics 6 and 7 are #641.

### Approach decided 2026-08-07

The dashboard is a **Looker Studio report over the GA4 property**, not an in-app superadmin view. This changes the shape of the work from what #639 originally described — there is no daily rollup table, no admin-portal page and no bespoke CSV endpoint to build. Looker provides sharing, filtering and CSV export natively, and access is controlled by Google account sharing rather than app RBAC.

Precedent: Glow Analytics (MOE) runs custom GA events into a Looker Studio dashboard with derived metrics such as "Engaged users (played ≥20% of any learning bite)" and "User Penetration rate (logged-in users / addressable market)" — the same shape as the auto-assign efficacy and absence-source metrics here.

### The scorecard

| # | Metric | Category | Baseline | Target | Source |
|---|---|---|---|---|---|
| 1 | Daily active users | Adoption | n/a | n/a | GA4 native |
| 2 | % of absences reported through ReliefCher vs usual comms | Adoption | 0% | 80% | #640 |
| 3 | Time spent on relief assignment each day | Ops efficiency | 30 to 45 mins | <= 5 mins | GA4 timing events; baseline from #641 |
| 4 | Auto-assign adoption rate | Ops efficiency | 0 | 60% | #640 |
| 5 | Period coverage rate each day | Reliability | 90 to 95% | <= 1% | **Not derivable from client events — see below** |
| 6 | Weekly check-in (1 to 5 ease of planning) | Satisfaction | n/a | >= 4 avg | #641 |
| 7 | NPS | Satisfaction | n/a | >= 30 | #641 |

### Period coverage rate does not fit the event model

Coverage rate is a state-of-the-world figure: how many periods needed relief versus how many were filled on a given day. An **uncovered** period produces no user action, so there is no client-side event to count. It has to come from the database. Two options, to be settled at grooming:

1. Emit a nightly server-side event per school to the GA4 Measurement Protocol carrying that day's coverage rate, keeping the report on a single source. Less work.
[...]

Issue #330, opened 2026-06-15
bug(security): public reporting page leaks staff roster — gate /{slug}/reporting behind school-level auth by default


guangshinhaha:
## Problem
The public absence-reporting page `/{slug}/reporting` exposes school data with **no authentication** by default. The gate is the per-school `School.publicReportRequiresAuth` flag, which **defaults to `false`** ([prisma/schema.prisma:150](prisma/schema.prisma#L150)). When it's off (the default for every new school), anyone who knows or guesses a school's slug can load the page and see:

- The **entire staff roster** — every teacher's **name** (and subjects) rendered into the dropdown ([src/app/[slug]/reporting/page.tsx](src/app/[slug]/reporting/page.tsx), the `prisma.teacher.findMany` branch).
- The school's **period / timetable structure** (period numbers + start/end times).

For an MOE pilot this is a staff-data exposure: a school directory is leaked to the open internet behind a guessable URL, and absence reports can be submitted by anyone.

## Desired behaviour
Gate the reporting page (and anything else school-scoped) behind login **at the school level — nothing should render publicly**.

- An unauthenticated visitor to `/{slug}/reporting` should see **only** a sign-in prompt — never the staff list, periods, or any school-identifying data beyond what's needed to sign in.
- Make authenticated reporting the **default** (effectively `publicReportRequiresAuth = true` semantics for everyone), rather than opt-in per school.
- Decide whether the public-no-auth mode should be **removed entirely** or kept as an explicit, clearly-labelled opt-in that an admin must turn on knowing the exposure. (Recommendation: default-secure; if the open mode stays, it must be a deliberate, warned toggle.)
- Confirm `/` (root) does not leak any school list or school-identifying data either.

## Scope / implementation notes
- Today the auth check lives in the page component ([page.tsx:45](src/app/[slug]/reporting/page.tsx#L45)) and in the API routes ([api/absence-reports/route.ts:105](src/app/api/absence-reports/route.ts#L105), [api/teacher-periods/route.ts:78](src/app/api/teacher-periods/route.ts#L78)). All three read `publicReportRequiresAuth` — flipping the default (or removing the public path) must cover **all three** so the roster isn't reachable via the API even if the page is gated.
- `middleware.ts` currently gates `/{slug}/dashboard`, `/{slug}/teacher`, `/admin`, `/onboarding` but **not** `/{slug}/reporting` — consider gating reporting in middleware too for defence-in-depth.
[...]

PR #646, opened 2026-08-11
feat(analytics): instrument auto-assign efficacy and absence reporting channel as GA events

### LINKED ISSUE #640: feat(analytics): instrument auto-assign efficacy and absence reporting channel as GA events
## User story

As the ReliefCher product team, I want auto-assign runs and the channel an absence arrived through recorded as GA events during normal use, so that pilot scorecard metrics 2 and 4 can be computed without anyone hand-counting.

## Background

Split from #639. This is the **instrumentation** slice. The Looker Studio report that consumes these events is #642; the surveys behind metrics 6 and 7 are #641.

### Approach decided 2026-08-07

Tracking is done with **GA4 custom events and event parameters**, read by a **Looker Studio** report, not a bespoke in-app analytics pipeline. Precedent: the SDT Data Tool (MOE DXD) ships GTM to production (`PRD_GA_TAG_MANAGER_ID` in its `.gitlab-ci.yml`), and the Glow Analytics product (MOE) runs custom GA events into a Looker Studio dashboard with derived metrics such as "Engaged users (played >=20% of any learning bite)". ReliefCher already has a GA4 tag live at `src/app/layout.tsx:36-47` and a `reliefcher` GA4 property collecting automatic events.

### Scope reduction agreed 2026-08-07

An earlier draft of this issue proposed a new `AbsenceSource` enum, a `source` column on `AbsenceReport`, and a channel dropdown in the coordinator's absence form. **That is no longer in scope.**

Metric 2 is "% of absences reported through ReliefCher vs usual comms", which is a binary, and the route already resolves it: `src/app/api/absence-reports/route.ts:68-104` sets `submittedBy` to `TEACHER` or `ADMIN` on every submission. A teacher filing their own absence came through ReliefCher; a coordinator filing on someone's behalf was told some other way and keyed it in. That mapping is sufficient for the metric and costs nothing.

Recording *which* channel (WhatsApp vs phone vs email) would need the enum, the column and the form field. It is useful colour for a school but it is not metric 2, so it is deferred. This removes one migration and one form change from the pilot.

`AutoAssignRun` remains the only schema addition, and is unavoidable: see Technical context.

### The two gaps this issue closes

**Auto-assign is measured in a way that overstates it.** `relief.auto_assign` is only written in the confirm branch of `src/app/api/auto-assign/route.ts`, and `preview` is the default. A coordinator who runs auto-assign, dislikes the result and assigns manually is invisible. Worse, the confirm flow accepts client-edited assignments (#425) while the audit row records only `createdCount`, so "accepted the algorithm's picks" and "used the auto-assign UI then overrode half of them" produce identical rows. As instrumented, metric 4 measures UI usage rather than algorithm efficacy.

**Metric 2 should not require daily manual counting.** Decided in Slack on 2026-08-05 with Caleb ANG (GovTech TP): we can distinguish self-reported absences from those raised by an HOD or coordinator from data we already hold, so we should not ask the committee to count manual reports each day.

### Known defect in the current GA setup

The existing tag at `src/app/layout.tsx:36-47` is **unconditional and hardcoded** (`G-KN5H3Z3JJR`) and is committed to `main`. Local, test, preview deploys and production all report into the same GA property with nothing distinguishing them, so no figure currently in that property is trustworthy as a pilot metric. This must be fixed as part of this issue, and should be fixed first so that later testing does not pollute production data further.

### Constraints noted in Slack, 2026-08-07

- **Do not push to production.** Land on dev only; production is frozen until the week of 2026-08-10 to protect Tuesday's pilot usage.

Related: #608 (school-facing absence analytics) is a different audience and not a dependency.

### Auto-assign efficacy is distinguishable from auto-assign usage

- **Given** a coordinator runs auto-assign, edits some of the proposed pairings, and confirms
- **When** the product team views the pilot metrics for that day
- **Then** the run is reported both as an auto-assign use and with the proportion of proposed pairings kept unedited, and a run that was previewed but never confirmed is counted as an abandoned run rather than being absent from the data

### Absence reporting channel is captured without the coordinator counting anything

- **Given** a coordinator is filing an absence on behalf of a teacher who told them by WhatsApp
- **When** they submit the absence
- **Then** it is recorded as having arrived outside ReliefCher, and the percentage of absences that arrived through ReliefCher is calculated with no manual tally and no additional field for the coordinator to fill in

### Events identify the school but never the person

- **Given** any instrumented action fires an event
- **When** that event is sent to GA
- **Then** it carries a school identifier and the actor's role, and contains no name, email address or other personal identifier

### Analytics from non-production environments never reach pilot metrics

- **Given** the GA tag today fires the same measurement ID from local, test, preview and production
- **When** an event is recorded from a non-production environment
- **Then** it does not appear in the pilot metrics for any school

### Events that cannot afford to be lost survive client-side blocking

- **Given** a user's browser or managed device blocks client-side analytics
- **When** they complete an instrumented action
- **Then** the metrics that the pilot scorecard depends on are still recorded

## Out of scope

- Recording which specific channel an absence arrived through (WhatsApp, phone, email, in person). The pilot metric needs only the binary. See the scope reduction above.
- Late-absence and incremental-run measurement (`E_late`, `R_preservation`, and the `runType` / `previousRunId` / `lateSlotsCount` fields). Deferred 2026-08-11, see grooming decision 4.
- The Looker Studio report, its custom dimension definitions and the period-coverage-rate feed: #642.
- All in-app surveys (baseline, satisfaction check-in, NPS): #641.
- Deploying to production. Dev only. (The production freeze referenced here lifted on 2026-08-11; production rollout is still handled separately from this issue.)
- School-facing or school-leader-facing analytics. School-facing absence trends are #608.
- Connecting an external BI tool directly to the production database.
- WOGAA onboarding. Worth evaluating separately for the traffic and sentiment slice.
- Engaging SDIO for the baseline figure.
- Cross-school benchmarking or league tables. Metrics are reported per school.
- Retention or pruning policy for the underlying event data.

## Design assets

No Figma. This issue owns the `Absence reporting channel` and `Auto-assign preview, edit and confirm events` nodes:

```mermaid
flowchart LR
  A[Coordinator and teacher actions] --> B[Absence reporting channel, from submittedBy]
  A --> C[Auto-assign preview, edit and confirm events]
  A --> D[Auth events]
  A --> E[Inline surveys: baseline, satisfaction, NPS]

  B --> F[(GA4 property)]
  C --> F
  D --> F
  E --> F
  G[(Period coverage rate, server-side)] --> F

  F --> H[Looker Studio pilot metrics report]
  H --> I[CSV export]
```

## Dependencies

Blocked by: none. Blocks: #642, since metrics 2 and 4 have no source until this lands.

---

## Technical context

Codebase areas touched:

- `src/app/api/auto-assign/route.ts`: preview flag at :60, client-assignment short-circuit at :83, existing audit write at :338
- `src/app/[slug]/dashboard/ReliefOverview.tsx`: preview call at :168, confirm call at :187
- `src/app/api/absence-reports/route.ts`: `submittedBy` resolution at :68-104, audit write at :594
- `src/app/layout.tsx:36-47`: the existing hardcoded GA tag

New files:

- `src/lib/analytics.ts`: server-side GA4 Measurement Protocol sender
- `src/lib/analyticsClient.ts`: browser gtag wrapper
- `src/lib/constants/analyticsEvents.ts`: event name and param constants, declared once and imported (shared-lookup-data rule in CLAUDE.md)
- `src/components/Analytics.tsx`: env-gated replacement for the inline tag in layout.tsx

Patterns to follow:

- `src/lib/absenceAudit.ts`: tolerant append-only writer. An analytics failure must never change a request's HTTP outcome. Catch, log, continue.
- `src/lib/auditLog.ts`: `writeAuditLog` signature and metadata shape.

Client vs server split:

- Server-side (Measurement Protocol): `absence_reported`, `auto_assign_preview`, `auto_assign_confirm`. These are scorecard-critical and must survive ad blockers and managed-device policy. A client-reported edit count would also be self-scoring.
- Client-side (gtag): page and session activity for metrics 1 and 3, which are inherently browser-side.

Server events must be dispatched via `after()` from `next/server`, not left as an un-awaited promise. On serverless a floating promise is killed when the function returns its response, which would drop events unpredictably.

### Why `AutoAssignRun` is unavoidable

Preview and confirm are two separate POSTs, and the confirm branch skips `autoAssign()` entirely when the client supplies its own assignments (`hasClientAssignments` at :83). The server therefore holds no record of what it proposed by the time it is asked to save. GA is write-only and cannot be read back mid-request, so the proposal set must be persisted server-side between the two calls. Redis was considered and rejected: a TTL'd key expires, which would lose exactly the abandoned runs the issue wants counted.

### GA property setup (prerequisite, not code)

The GA4 property exists and is collecting automatic events. Still required:

1. A second GA4 property for dev, so the existing one becomes prod-only.
2. A Measurement Protocol API secret (Admin, Data streams) for server-side events.
3. Registered custom dimensions for `school_id`, `submitted_by`, `run_id`, `environment`, plus custom metrics for the auto-assign counts. Unregistered params are invisible in Looker Studio, take 24 to 48 hours to populate, and are not retroactive. Register these before firing events.
4. Google Signals disabled on the prod property, to reduce the thresholding risk tracked on #642.

### Deployment configuration

Deployment is to a self-managed AWS environment owned by the ReliefCher team. There is no Vercel
tier: preview-deploy scoping does not apply, and the environment split is handled entirely by the
three variables below.

Three server-side environment variables. None is a `NEXT_PUBLIC_` variable, so none needs a build
arg or a Dockerfile change.

| Variable | Secret? | Value | Notes |
|---|---|---|---|
| `GA_MEASUREMENT_ID` | No | `G-KN5H3Z3JJR` in prod; the new non-prod property elsewhere | Different value per tier. This is the primary separation |
| `GA_MEASUREMENT_PROTOCOL_SECRET` | **Yes** | GA Admin > Data streams > Measurement Protocol API secrets | Per property, so prod and non-prod have different secrets |
| `APP_ENV` | No | `production` / `test` / `development` | Emitted as the `environment` event param |

All three are optional at runtime. If `GA_MEASUREMENT_ID` or the secret is unset, the sender no-ops
and nothing is sent, which is the intended state for local development. If `APP_ENV` is unset the
param is emitted as `unknown`, so a misconfigured tier shows up as a visible bucket in Looker rather
than silently merging into a real one.

`APP_ENV` is a new variable. `NODE_ENV` cannot do this job: a staging or test build is also
`NODE_ENV=production`, so it cannot distinguish tiers. No existing variable in `.env.example`
carries a tier label.

**Why `APP_ENV` is needed even though the properties are already separate.** Browser events are
separable after the fact by GA4's `hostname` dimension. Measurement Protocol events carry no
hostname. If a non-prod deployment is ever misconfigured with the prod measurement ID, `APP_ENV` is
the only signal that would reveal it.

**Repo state:** `docker-compose.yml` passes these through to the `app:` service. That block is an
explicit allowlist, so a variable present in `.env` but not named there never reaches the container.

**Manual steps owned by the deployer**, in whichever mechanism that AWS environment uses to inject
configuration (`.env` beside `docker-compose.yml`, ECS task definition, or otherwise):

1. Set all three variables per tier, with a different `GA_MEASUREMENT_ID` for prod and non-prod.
2. Keep `GA_MEASUREMENT_PROTOCOL_SECRET` out of plaintext task definitions. If ECS, reference it
   from Secrets Manager or SSM.
3. Leave all three unset locally unless deliberately testing analytics.

Full writeup: `docs/ops/analytics-env-vars.md`.

### Measurement Protocol `client_id`

Every Measurement Protocol event requires a `client_id`. This is a correctness decision, not a
detail: if server events send a freshly generated ID per request, GA4 counts each one as a distinct
user, which inflates user counts and corrupts metric 1 (DAU), the one metric that otherwise comes
for free.

The `_ga` cookie already set by the browser tag holds the real client ID. Server events must read it
from the request and pass it through, so a server event and the browser events from the same session
attribute to one user. When the cookie is absent (managed device blocking the tag, or a request with
no browser context), fall back to a stable value derived from the session's user ID rather than a
random one, so repeated events from the same person do not fan out into phantom users.

### `Analytics.tsx` must be a Server Component

The claim above that no `NEXT_PUBLIC_` variable is needed holds only if `src/components/Analytics.tsx`
reads `GA_MEASUREMENT_ID` on the server and renders the resolved ID into the emitted script tag.
Adding `"use client"` to that file makes the variable read as `undefined` in the browser and the tag
silently stops firing, with no error. Keep it a Server Component.

## Data model

One new model. No changes to `AbsenceReport` or `AbsenceReportAudit`.

```prisma
model AutoAssignRun {
  id                String    @id @default(cuid())  // the runId
  schoolId          String
  date              DateTime
  runByUserId       String
  proposedCount     Int
  totalNeededSlots  Int?      // slots requiring cover for this school and date at preview time
  attemptNumber     Int       @default(1)  // nth preview for (schoolId, date)
  proposals         Json      // [{ timetableEntryId, reliefTeacherId }]
  previewedAt       DateTime  @default(now())
  confirmedAt       DateTime?
  keptUneditedCount Int?
  editedCount       Int?
  addedCount        Int?
  removedCount      Int?
  createdCount      Int?
  correlated        Boolean   @default(true)  // false when confirm arrived with no usable runId

  school School @relation(fields: [schoolId], references: [id], onDelete: Cascade)

  @@index([schoolId, previewedAt])
  @@index([schoolId, date])
  @@index([confirmedAt])
}
```

`totalNeededSlots` and `attemptNumber` were added on 2026-08-11 from the proposal in the third comment. Both are captured at preview time. `attemptNumber` is `count(runs for schoolId + date) + 1`, which is what the `[schoolId, date]` index is for. `totalNeededSlots` is nullable so a run recorded before the count is resolvable still saves.

Diff definitions. A proposal is identified by `timetableEntryId`:

- **kept unedited**: confirmed assignment matches a proposal on both `timetableEntryId` and `reliefTeacherId`
- **edited**: `timetableEntryId` matches, `reliefTeacherId` differs
- **removed**: proposed `timetableEntryId` absent from the confirmed set
- **added**: confirmed `timetableEntryId` that was never proposed
- **abandoned**: `confirmedAt IS NULL`

Three rates come out of this, and they are deliberately different questions:

- **Logic efficacy** = `keptUneditedCount / proposedCount`. When the algorithm proposed something, was it right?
- **Real efficacy** = `keptUneditedCount / totalNeededSlots`. What share of the day was automated with no human touch?
- **Intervention rate** = `(editedCount + removedCount + addedCount) / totalNeededSlots`. How much did the coordinator rework?

Real efficacy and intervention rate do not sum to 100%. The remainder is passive shortage: slots nobody filled because no candidate was available. That is a different problem from the algorithm proposing badly (recruit more relief teachers, versus tune the weights), and `keptUneditedCount / proposedCount` alone cannot tell them apart. A school with a chronic relief shortage and a school where the algorithm works perfectly both score high on logic efficacy.

None of the three are stored. All are computed at read time from the counts.

Adoption is confirmed runs over relief-planning days.

**Open product decision, not an implementation choice:** the 60% scorecard target does not currently say which efficacy it means. Logic efficacy flatters a school with a relief shortage; real efficacy penalises one through no fault of the algorithm. Recommendation is to report real efficacy as the headline (it is what a school experiences) with logic efficacy alongside as the diagnostic that explains a bad number. To be confirmed with Caleb ANG before the pilot produces numbers.

The diff itself belongs in a pure function with no DB or network access, so it can be unit tested in `src/__tests__/` rather than the integration suite.

Metric 2 needs no model change. It is `count(submitted_by = TEACHER) / count(*)` over `absence_reported` events, using the value the route already resolves. `AbsenceReportAudit.submittedBy` keeps its current meaning and existing audit consumers are unaffected.

## API contract

`POST /api/auto-assign?preview=true`

- Response gains `runId: string`.

`POST /api/auto-assign?preview=false`

- Request gains `runId?: string`. Response unchanged.

`POST /api/absence-reports`

- Request and response unchanged. The route emits a GA event using the `submittedBy` value it already computes.

GA4 events emitted:

```
absence_reported     { school_id, submitted_by, environment }
auto_assign_preview  { school_id, run_id, proposed_count, total_needed_slots,
                       attempt_number, environment }
auto_assign_confirm  { school_id, run_id, proposed_count, total_needed_slots,
                       attempt_number, kept_unedited_count, edited_count,
                       added_count, removed_count, created_count, environment }
```

No name, email address, NRIC or teacher identifier appears in any event payload.

Counts only. Do not send precomputed percentages (`logic_efficacy_pct`, `real_efficacy_pct`, `intervention_rate_pct`) as event parameters. They are derivable in Looker from the counts already present in the same event, and each one sent would consume a registered custom dimension for no additional information.

## Error contract

No new error responses. Every failure mode in this issue is deliberately non-fatal:

- `runId` absent, unknown, or belonging to another school on confirm: save the assignments, record the run with `correlated = false`, log a warning, continue. Analytics must never block relief planning.
- GA delivery failure (network error, non-2xx from the Measurement Protocol): log and continue. The HTTP response is unchanged.
- `AutoAssignRun` write failure at preview: return the proposals with `runId` omitted.

## Additional test scenarios

- **Given** a confirm arrives with a `runId` belonging to another school, **when** it is processed, **then** the assignments save, the run is recorded with `correlated = false`, and no cross-school data is read.
- **Given** two coordinators preview concurrently for the same school and date, **when** both confirm, **then** each run diffs against its own proposal set.
- **Given** a confirm sends assignments identical to the proposals, **when** the diff runs, **then** `keptUneditedCount` equals `proposedCount` and `editedCount` is 0.
- **Given** a preview returns zero proposals, **when** the run is recorded, **then** `proposedCount` is 0 and efficacy is reported as no-data rather than 0% or a divide-by-zero.
- **Given** the Measurement Protocol endpoint times out, **when** an absence is submitted, **then** the absence is created and a 201 is returned.
- **Given** the GA measurement ID env var is unset, **when** any page renders, **then** no GA script tag is emitted and no request is made to Google.
- **Given** an absence is submitted with no session (`submittedBy = PUBLIC`), **when** the event fires, **then** it is not counted as either self-service or coordinator-relayed in metric 2.
- **Given** a coordinator previews three times for the same school and date, **when** each run is recorded, **then** `attemptNumber` is 1, 2 and 3 respectively.
- **Given** auto-assign proposes for 10 of 12 slots needing cover and the coordinator keeps all 10, **when** the rates are computed, **then** logic efficacy is 100% and real efficacy is 83%.
- **Given** `totalNeededSlots` is null on a run, **when** the rates are computed, **then** real efficacy and intervention rate report as no-data rather than 0% or a divide-by-zero, and logic efficacy is still reported.

## Hard constraints

- Do not write migration SQL by hand. Run `npx prisma migrate dev --name <name>` (CLAUDE.md).
- Do not deploy to production as part of this issue. Dev only. (The freeze that originally motivated this lifted on 2026-08-11.)
- Do not send any personal identifier to GA. School ID and role only.
- Do not add a `source` field to `AbsenceReport` or change the meaning of `AbsenceReportAudit.submittedBy`.
- Do not add a channel question to the absence form. Explicitly descoped.
- Do not trust a client-reported edit count. The diff is computed server-side from stored proposals.
- Do not add `runType`, `previousRunId` or `lateSlotsCount`. Deferred on 2026-08-11, see grooming decision 4.
- Do not send precomputed percentages to GA. Counts only.
- Do not add `"use client"` to `src/components/Analytics.tsx`. It must read `GA_MEASUREMENT_ID` server-side.
- Do not generate a random `client_id` per Measurement Protocol event. Reuse the `_ga` cookie value, falling back to a stable per-user value.
- Update `.env.example` with `GA_MEASUREMENT_ID`, `GA_MEASUREMENT_PROTOCOL_SECRET` and `APP_ENV`, placeholder values only (CLAUDE.md).
- Do not add an analytics SDK dependency. Use `gtag` plus `fetch` to the Measurement Protocol.
- No `any` casts and no eslint-disable comments (CLAUDE.md).
- Update `prisma/seed.ts` for the new model in the same PR (CLAUDE.md).

## Suggested order

1. Env-gate the GA tag. Small, and until it lands every test run pollutes the production property further.
2. Register the custom dimensions in GA (starts a 24 to 48 hour clock, not retroactive).
3. `src/lib/analytics.ts` plus the `absence_reported` event. Self-contained, no schema change, ships alone.
4. `AutoAssignRun`, the diff function, and the `runId` round-trip through `ReliefOverview.tsx`.

Steps 1 and 2 are worth doing before any of the rest.

## Grooming decisions taken (revisable at implementation)

1. **Metric 2 is derived from the existing `submittedBy` value rather than a new field.** Agreed 2026-08-07. Removes one migration and one form change. The cost is that the channel breakdown (WhatsApp vs phone vs email) is not available; add it later if a school asks for it.
2. **An uncorrelated confirm saves silently** with a logged warning rather than failing. This keeps relief planning safe, but a client bug could quietly degrade metric 4 to nothing, and the log line is the only signal.
3. **Verification is via GA DebugView**, adding `debug_mode: true` during development, rather than waiting on the standard reporting delay.
4. **`totalNeededSlots` and `attemptNumber` accepted; the late-absence fields deferred.** Decided 2026-08-11 in response to the dual-efficacy proposal in the third comment. The two accepted fields separate algorithm quality from relief shortage and expose preview friction, for two nullable columns. `runType`, `previousRunId` and `lateSlotsCount` are deferred because a run cannot currently be classified correctly: the only available rule ("a confirmed run already exists for this school and date") cannot distinguish a genuine late-absence patch from a coordinator redoing a plan they were unhappy with. Both produce identical rows, so `E_late` and `R_preservation` would be computed over a denominator that includes redos. That yields confidently wrong numbers rather than missing ones. The correct rule needs a third case (INITIAL, INCREMENTAL, REDO) keyed on whether any slot traces to an `AbsenceReport.createdAt` later than the previous run's `confirmedAt`, and settling it needs observation of a real late-absence day during the pilot. All three fields are nullable additions, so adding them afterwards is a trivial migration with nothing lost by waiting.

---

*🤖 Generated with aif-split-issue*
*🤖 Groomed with aif-groom-issue*

guangshinhaha:
Closes #640

## Summary

- Environment-gates the GA tag, which was hardcoded and unconditional on `main`, so local/test/production stop reporting into one property.
- Adds a server-side Measurement Protocol sender so the scorecard metrics survive browsers and managed devices that block client-side analytics.
- Emits `absence_reported` from the existing `submittedBy` value, giving metric 2 with no new column and no extra question on the absence form.
- Adds `AutoAssignRun` plus a pure diff so metric 4 can separate algorithm efficacy from UI usage, and so abandoned previews are counted instead of vanishing.

## Why

The pilot scorecard has seven metrics. Two of them assumed the Relief Planning Committee would hand-count numbers daily, and metric 4 was instrumented in a way that overstates it. Details in #640.

## Changes

| File | Why |
|---|---|
| `src/lib/analyticsConfig.ts` | Pure resolution of `GA_MEASUREMENT_ID` / `GA_MEASUREMENT_PROTOCOL_SECRET` / `APP_ENV`, so the gating rule is unit-testable and lives in one place |
| `src/components/Analytics.tsx` | Env-gated replacement for the inline tag. **Server Component by requirement** |
| `src/app/layout.tsx` | Drops the hardcoded `G-KN5H3Z3JJR` tag |
| `src/lib/analytics.ts` | Measurement Protocol sender, dispatched via `after()` |
| `src/lib/analyticsClientId.ts` | `client_id` resolution from the `_ga` cookie, with a stable hash fallback |
| `src/lib/constants/analyticsEvents.ts` | Event and param names declared once |
| `src/app/api/absence-reports/route.ts` | Emits `absence_reported` |
| `src/lib/autoAssignDiff.ts` | Pure proposed-vs-confirmed diff |
| `src/lib/autoAssignRun.ts` | Run lifecycle and the two auto-assign events |
| `src/app/api/auto-assign/route.ts` | Records the preview, returns `runId`, closes the run on confirm |
| `src/app/[slug]/dashboard/ReliefOverview.tsx` | Echoes `runId` back on confirm |
| `prisma/schema.prisma` + migration | One new table, additive only |
| `prisma/seed.ts` | Four runs covering accepted / overridden / abandoned / retried |
| `.env.example`, `docker-compose.yml` | The three new variables |

## Test plan

| Acceptance criterion | Covered by |
|---|---|
| Auto-assign efficacy distinguishable from usage | `tests/api/autoAssignRun.test.ts` (abandoned preview, accepted vs overridden, needed-slots), `src/__tests__/autoAssignDiff.test.ts` (12 cases) |
| Absence channel captured with no manual tally | `src/__tests__/analytics.test.ts` ("distinguishes a self-reported absence from a coordinator-relayed one") |
| Events identify the school but never the person | `src/__tests__/analytics.test.ts` ("carries the school and role but no personal identifier") |
| Non-production analytics never reach pilot metrics | `src/__tests__/analyticsConfig.test.ts` (7 cases incl. unset ID emits nothing, tiers resolve separately) |
| Events survive client-side blocking | `src/__tests__/analyticsClientId.test.ts` (cookie-blocked fallback is stable, no phantom users) |

Commands run:

- `npx tsc --noEmit` clean
- `npm test` 96 files, 1013 passed, 2 skipped
- `npm run build` clean, 0 prerender errors
- `npm audit` 0 vulnerabilities
- `npm run test:api` new file 10/10 passing

### Note on the integration suite

`npm run test:api` is red on `main` and stays equally red here. I ran `absence-reports`, `auto-assign` and `surveys` on `origin/main` and on this branch: **41 failed / 29 passed on both**. The cause is test fixtures hardcoding dates in April 2026 that semantic date validation now rejects, four months on. Unrelated to this PR and deliberately not touched.

## Not in this PR

- The Looker Studio report (#642).
- GA console setup: second property for non-prod, Measurement Protocol secrets, custom dimension/metric registration, Google Signals off. **Registration is not retroactive**, so it must happen before events are relied on.
- Setting the three variables in the AWS environment. See the issue comment for the dev and prod values.

---

> **Before marking ready for review**: run the app locally and walk the golden path (report an absence as a teacher, run auto-assign, edit one proposal, confirm) with `GA_MEASUREMENT_ID` unset, to confirm nothing regresses when analytics is off.

*🤖 Generated with aif-implement-issue*


src/lib/analytics.ts
```diff
@@ -0,0 +1,106 @@
...
+      method: "POST",
+      headers: { "Content-Type": "application/json" },
+      body: JSON.stringify({
+        client_id: clientId,
+        events: [
+          {
+            name: input.name,
+            params: {
+              ...input.params,
+              [ANALYTICS_DIMENSIONS.ENVIRONMENT]: config.appEnv,
```
guangshinhaha, line 110: **[High] Measurement Protocol payload omits `session_id` and `engagement_time_msec`, so these events will not show as user activity in GA reports**

```ts
      body: JSON.stringify({
        client_id: clientId,
        events: [
          {
            name: input.name,
            params: {
              ...input.params,
              [ANALYTICS_DIMENSIONS.ENVIRONMENT]: config.appEnv,  // ← no session_id, no engagement_time_msec
            },
          },
        ],
      }),
```

**Problem:** Google is explicit that "in order for user activity to display in reports like Realtime, `engagement_time_msec` and `session_id` must be supplied as part of the params for an event" ([sending-events](https://developers.google.com/analytics/devguides/collection/protocol/ga4/sending-events), [Chrome GA4 guide](https://developer.chrome.com/docs/extensions/how-to/integrate/google-analytics-4)). Concretely: a coordinator confirms an auto-assign run, the MP call returns 2xx (which this code treats as success, and which it must — the endpoint 2xxes on silently discarded payloads too), and the event lands with no session attached. `absence_reported` / `auto_assign_confirm` then do not appear in Realtime, are not attributed to the browser session the same user generated via the gtag tag, and contribute nothing to engaged-sessions / average-engagement-time. Since registration is not retroactive and #642 builds the scorecard on exactly these events, a pilot could run for weeks before anyone notices the session-scoped views are empty.

The `session_id` is available from the same request that already supplies `_ga`: GA writes it into the `_ga_<STREAM>` cookie (the measurement ID minus the `G-` prefix), third dot-separated component.

**Suggestion:**
```ts
// analyticsClientId.ts — alongside parseGaClientId
/** `_ga_<STREAM>` is `GS1.1.<session_id>.<n>....` — component 3 is the session ID. */
export function parseGaSessionId(cookieValue: string | null | undefined): string | null {
  if (!cookieValue) return null;
  const parts = cookieValue.trim().split(".");
  return parts.length >= 3 && /^\d+$/.test(parts[2]) ? parts[2] : null;
}

// analytics.ts — caller passes gaSessionCookie from
// request.cookies.get(`_ga_${config.measurementId.replace(/^G-/, "")}`)?.value
params: {
  ...input.params,
  [ANALYTICS_DIMENSIONS.ENVIRONMENT]: config.appEnv,
  // Required for the event to register as user activity. Falls back to the
  // client_id-derived value so a cookie-blocked request still lands in *a*
  // session rather than none.
  session_id: resolveSessionId(input) ?? clientId.split(".")[0],
  // 1ms is the conventional floor: enough to mark the event as engaged
  // without inflating average engagement time.
  engagement_time_msec: 1,
},
```

> Tradeoff: if you would rather not thread a second cookie through every caller, sending only `engagement_time_msec: 1` plus a `session_id` derived from `client_id` still gets the events into Realtime — it just groups all of one user’s server events into one long-lived pseudo-session rather than matching the browser’s.

---

src/lib/autoAssignRun.ts
```diff
@@ -0,0 +1,201 @@
...
+
+/**
+ * Mark a run as uncorrelated without failing the request.
+ *
+ * Called when a confirm carried a run ID that could not be resolved. Kept
+ * separate from the emit path so the warning is greppable in logs: a client bug
+ * that stops echoing the run ID would otherwise degrade metric 4 to nothing
+ * with no signal.
+ */
+export async function markRunUncorrelated(runId: string): Promise<void> {
```
guangshinhaha: **[Medium] `markRunUncorrelated` has no caller, so `AutoAssignRun.correlated` is always `true` and the signal it exists to provide never fires**

```ts
/**
 * Mark a run as uncorrelated without failing the request.
 *
 * Called when a confirm carried a run ID that could not be resolved. Kept
 * separate from the emit path so the warning is greppable in logs: a client bug
 * that stops echoing the run ID would otherwise degrade metric 4 to nothing
 * with no signal.
 */
export async function markRunUncorrelated(runId: string): Promise<void> {  // ← never called
  try {
    await prisma.autoAssignRun.updateMany({
      where: { id: runId },   // ← no schoolId scope
      data: { correlated: false },
    });
```

**Problem:** `git grep markRunUncorrelated` returns only this definition — no route, no lib, no test calls it. The `!run` branch in `recordAutoAssignConfirm` (line 130), which is the one place the doc comment says it fires, emits the uncorrelated event and returns without touching the column. So every row written by this PR carries `correlated: true`, including the ones that are demonstrably not correlated, and `prisma/schema.prisma:973` ("False when a confirm arrived with no usable runId") describes a state the code cannot produce. The stated failure mode — a client regression stops echoing `runId`, metric 4 quietly degrades to nothing — is exactly the one that still has no signal, because a run that is never confirmed and a run whose confirm lost its ID are both just `confirmedAt: null`.

There is also a latent multi-tenancy gap if this is wired up later: `updateMany({ where: { id: runId } })` is unscoped, so unlike the `findFirst` above it (which deliberately puts `schoolId` in the lookup) a coordinator could flip the flag on another school's run by posting a guessed ID.

**Suggestion:** call it from the branch that motivates it, and scope it the same way the read is scoped.
```ts
export async function markRunUncorrelated(
  runId: string,
  schoolId: string,
): Promise<void> {
  try {
    await prisma.autoAssignRun.updateMany({
      where: { id: runId, schoolId },
      data: { correlated: false },
    });
  } catch (err) {
    console.error("[autoAssignRun] failed to mark run uncorrelated", err);
  }
}

// recordAutoAssignConfirm, in the `if (!run)` branch:
if (!run) {
  if (input.runId) {
    // An ID was echoed but did not resolve — a stale preview, or a run from
    // another school. Distinct from "no ID was sent at all", which has no row
    // to mark.
    console.warn(`[autoAssignRun] confirm carried unresolvable runId ${input.runId}`);
    await markRunUncorrelated(input.runId, input.schoolId);
  }
  queueServerEvent({ /* … unchanged … */ });
  return;
}
```

> Alternative, if the flag is not worth wiring now: drop `markRunUncorrelated` and the `correlated` column in this PR and add them when something reads them. Shipping an unread column plus an uncalled export is the cost either way, and the schema comment currently promises behaviour that does not exist.

---

src/lib/autoAssignRun.ts
```diff
@@ -0,0 +1,201 @@
...
+ */
+export async function recordAutoAssignPreview(
+  input: RecordPreviewInput,
+): Promise<string | null> {
+  try {
+    const dateUTC = new Date(input.date + "T00:00:00.000Z");
+
+    // Nth preview for this school and date. Counted rather than incremented so
+    // concurrent previews cannot collide on a shared counter.
+    const priorRuns = await prisma.autoAssignRun.count({
```
guangshinhaha, line 56: **[Low] `attemptNumber` is count-then-create, so concurrent previews still collide — the comment claims the opposite**

```ts
    // Nth preview for this school and date. Counted rather than incremented so
    // concurrent previews cannot collide on a shared counter.   // ←
    const priorRuns = await prisma.autoAssignRun.count({
      where: { schoolId: input.schoolId, date: dateUTC },
    });
    …
        attemptNumber: priorRuns + 1,
```

**Problem:** Counting instead of incrementing removes lost-update on a shared counter row, but not the race on the value itself. Two coordinators in the same school both hit "Auto-assign" for 2026-08-12 within the same second: both `count()` calls return 3 before either `create()` lands, both rows are written with `attemptNumber: 4`, and no row ever carries 5. `attemptNumber` is the friction signal in metric 4 — "a coordinator previewing repeatedly before accepting" — so the sequence being non-unique means max(attemptNumber) understates the retry count for exactly the busy mornings the metric is meant to catch. Two coordinators on one date is normal in a school with a shared relief inbox, and the pilot deliberately runs several coordinators per school.

Nothing breaks (there is no unique constraint), so this is metrics-only — but the comment reads as a concurrency guarantee that a future reader would rely on.

**Suggestion:** either correct the comment, or make the number derivable after the fact rather than at write time.
```ts
    // Nth preview for this school and date, computed at write time. Two
    // previews racing can both read the same count and land on the same
    // attemptNumber; the sequence is a usage signal, not a unique key, so
    // an occasional duplicate is accepted rather than serialised.
    const priorRuns = await prisma.autoAssignRun.count({
      where: { schoolId: input.schoolId, date: dateUTC },
    });
```
```ts
    // …or drop the stored column and rank by previewedAt in reporting, where
    // ties resolve deterministically and no write-time read is needed:
    //   ROW_NUMBER() OVER (PARTITION BY "schoolId", "date" ORDER BY "previewedAt")
    // The (schoolId, date) index this PR already adds covers it.
```

> The second option also removes a round-trip from the preview path, which currently awaits `count()` then `create()` sequentially before the response can be built.

---

src/app/api/absence-reports/route.ts
```diff
@@ -603,6 +608,28 @@ export async function POST(request: NextRequest) {
...
+    // comms". `submittedBy` already carries the answer: TEACHER means the
+    // teacher filed it themselves in the app, ADMIN means a coordinator was
+    // told some other way and keyed it in. Emitting it here means the metric
+    // needs no new column and no extra question on the absence form, which is
+    // the whole point: the pilot schools were going to be asked to hand-count
+    // this every day.
+    //
+    // PUBLIC submissions are emitted too and are excluded when the metric is
+    // computed, rather than being silently folded into either side.
+    if (resolvedSchoolId) {
```
guangshinhaha: **[Low] `if (resolvedSchoolId)` is always true here, and no `submitted_by: PUBLIC` event can ever be emitted — contrary to the comment above it**

```ts
    // PUBLIC submissions are emitted too and are excluded when the metric is
    // computed, rather than being silently folded into either side.   // ←
    if (resolvedSchoolId) {   // ← always true at this point
      queueServerEvent({
        name: ANALYTICS_EVENTS.ABSENCE_REPORTED,
        params: {
          [ANALYTICS_DIMENSIONS.SCHOOL_ID]: resolvedSchoolId,
          [ANALYTICS_DIMENSIONS.SUBMITTED_BY]: submittedBy,
        },
```

**Problem:** The #330 auth guard at line 124 returns 401 whenever `resolvedSchoolId` is null, and `submittedBy` is only ever left at its `"PUBLIC"` initial value on that same unauthenticated path. By the time execution reaches line 621, `resolvedSchoolId` is non-null and `submittedBy` is `TEACHER` or `ADMIN` — never `PUBLIC`. So the guard is a dead branch and the comment describes a bucket that cannot exist.

The failure this produces is interpretive rather than runtime: whoever builds the #642 Looker Studio report reads this comment, adds `submitted_by != PUBLIC` to the metric-2 filter, sees zero PUBLIC rows, and reasonably concludes "nobody uses the public form" — when in fact the public form is closed and every rejected attempt 401s before any event is emitted. Rejected submissions are recorded by `recordAbsenceAttempt`, not by GA, and that distinction is worth stating where the metric is defined.

**Suggestion:**
```ts
    // #640 metric 2 — … (unchanged)
    //
    // Only TEACHER and ADMIN reach here: the #330 auth guard above 401s every
    // request that did not resolve to a school, and `PUBLIC` is only ever the
    // pre-auth default. Rejected attempts are captured by recordAbsenceAttempt
    // (AbsenceAudit), not by GA, so the metric denominator is authenticated
    // submissions only.
    queueServerEvent({
      name: ANALYTICS_EVENTS.ABSENCE_REPORTED,
      params: {
        [ANALYTICS_DIMENSIONS.SCHOOL_ID]: resolvedSchoolId,
        [ANALYTICS_DIMENSIONS.SUBMITTED_BY]: submittedBy,
      },
      gaCookie: request.cookies.get("_ga")?.value ?? null,
      fallbackSeed: submittedByUserId,
    });
```

---

guangshinhaha:
## Code Review Summary

| Severity | Count |
|----------|-------|
| Critical | 0 |
| High     | 1 |
| Medium   | 1 |
| Low      | 2 |

## What Looks Good

- **`AutoAssignRun` is the right call over a GA-only solution.** The reasoning in the schema comment holds up: preview and confirm are separate POSTs, the confirm branch skips `autoAssign()` when the client supplies its own assignments, and GA is write-only — so the proposal set genuinely has to be persisted between the two calls. Rejecting Redis because a TTL would expire exactly the abandoned runs the table exists to count is the correct trade to have made.
- **Diff counts are computed server-side from the stored `proposals`, never taken from the client.** For a metric whose subject is "how far did the human override the algorithm", a client-reported figure would be self-scoring. `diffAutoAssignProposals` is pure and keyed on `timetableEntryId`, and the four counts partition the union of proposed and confirmed slots cleanly.
- **`queueServerEvent` uses `after()` rather than a floating promise.** A dropped un-awaited send on a serverless runtime would surface as an inexplicably low count rather than an error — the least debuggable failure mode available. Every path through `sendServerEvent` and `recordAutoAssign*` also swallows and logs, so a measurement failure cannot change the HTTP outcome of a relief-planning request.
- **`recordAutoAssignConfirm` puts `schoolId` inside the `findFirst` lookup rather than checking it afterwards**, so a run ID belonging to another school reads as not-found and falls through to the uncorrelated path. Same pattern the rest of the codebase uses for tenant scoping, applied without being asked.
- **`resolveClientId` drops the event instead of inventing an identity.** A random `client_id` per request would register as a new user and inflate metric 1 — the one scorecard metric that otherwise costs nothing.

---

guangshinhaha:
## Review findings addressed — `25b3b18`

| Finding | Fix |
|---|---|
| **High** — MP payload omits `session_id` / `engagement_time_msec` | Both now sent on every server event. `session_id` comes from the per-stream `_ga_<STREAM>` cookie (`parseGaSessionId`, handles both the `GS1.` and `GS2.` layouts); `readGaCookies` reads the pair so no route has to know how GA names the second cookie. Cookie-less requests fall back to a 30-minute time bucket — GA's own inactivity window — rather than to a constant, which would have given every user one session of unbounded length. `engagement_time_msec` is a flat `1`: enough to register as engaged activity, too small to move average engagement time, and the server has no honest measurement to offer instead. |
| **Medium** — `markRunUncorrelated` uncalled, `correlated` always `true` | Both removed. The column was unwritable by construction — a confirm with no usable runId has no row to mark. The uncorrelated case is now observable where it actually can be: an `auto_assign_confirm` with `created_count` but no `proposed_count`, plus a log line separating "no runId" (client regression) from "unresolvable runId" (stale preview / another school's run). |
| **Low** — `attemptNumber` count-then-create race | Behaviour unchanged, comment corrected. Two coordinators previewing the same date can land on the same number; it is a friction signal, not a unique key, and serialising a measurement write behind a lock is not worth it. Reporting should read `max(attemptNumber)` as a lower bound. |
| **Low** — dead `if (resolvedSchoolId)` + wrong PUBLIC comment | Guard removed, comment rewritten. The #330 auth guard 401s first, so only `TEACHER` and `ADMIN` reach the emit — there is no third bucket to exclude. Rejected submissions live in `AbsenceAudit` via `recordAbsenceAttempt`, not GA, so the metric denominator is authenticated submissions only. |

### ⚠️ Migration note — needs your attention

`correlated` was removed from the `CREATE TABLE` in `20260811032802_add_auto_assign_run` rather than dropped in a follow-up migration. **This is a deliberate exception to the CLAUDE.md "never hand-edit migration SQL" rule**, approved in-session, and narrow: the migration is introduced by this PR and has never been applied to test or prod, so there is no ordering conflict and no drift on real data.

Consequence: **anyone whose local dev DB already applied the earlier version of this migration will hit a checksum mismatch** and needs to reset it. The integration suite is unaffected — `tests/api/setup.ts` uses `prisma db push`, not `migrate deploy`.

### Verification

- `npx tsc --noEmit` — clean (exit 0; `tsconfig` includes `tests/**`, so the integration test file typechecks against the new schema)
- `npm test` — 96 files, **1026 passed** / 2 skipped, up from 1013 (+13 covering session-ID parsing, the time-bucket fallback, the session params on the wire, and `gaSessionCookieName`)
- `npm run build` — clean, `✓ Generating static pages (96/96)`, 0 prerender errors
- `npm run test:api` — **not run.** Prisma 7 blocks the `prisma db push` in `tests/api/setup.ts` when it detects an agent, pending explicit human consent. Worth a local run before merge, though the changes to the covered functions are additive optional fields plus a log line.

### Still outstanding from the original PR description

GA console setup is unchanged and still blocking: `session_id` and `engagement_time_msec` are GA built-ins and need no registration, but the custom dimensions and metrics do — and **registration is not retroactive**.

---

Issue #641, opened 2026-08-07
feat(analytics): inline pilot surveys for time taken, satisfaction check-in and NPS


guangshinhaha:
## User story

As the ReliefCher product team, I want the time-spent, satisfaction and NPS questions asked inline in the app on a weekly or monthly cadence, so that pilot metrics are captured without daily surveys and without blocking anyone's work.

## Background

Split from #639. This is the **surveys** slice, covering scorecard metrics 3 (time spent on relief planning, tracked as an ongoing weekly average), 6 (weekly ease-of-planning check-in, target >= 4 avg) and 7 (NPS, target >= 30). The GA event instrumentation is #640; the Looker Studio report is #642.

**Surveys need to be inline, not daily.** Decided in Slack on 2026-08-05 with Caleb ANG (GovTech TP): daily surveys are annoying and will not get filled in, so cadence is weekly or monthly. All three surveys (time-spent, satisfaction, NPS) should be inline in the app, with an email poll at the end of the pilot as fallback if in-app response rates are poor.

**Metric 3 is tracked as an ongoing self-reported average, not a before/after baseline.** The scorecard originally suggested asking SDIO for a 30 to 45 minute pre-ReliefCher baseline, then (in an earlier version of this issue) a one-time in-app recall question as a substitute. Neither is used: SDIO involvement was ruled out for communication overhead, and a single retrospective recall number is a weak substitute for a trend. Metric 3 instead comes from a recurring weekly "how many minutes did relief planning take you per day, roughly?" question, asked in the same touchpoint as the ease-of-planning check-in — giving a time series across the pilot instead of one static number.

**FormSG is the fallback.** Confirmed in Slack on 2026-08-07 — an external FormSG will be prepared as a last resort if in-app response rates are poor.

Responses feed the same GA4 property that #640 writes to, so the Looker Studio report reads one source. See #640 for the GA + Looker approach and its constraints (no PII in GA, environment separation).

### Weekly check-in (time spent and ease of planning) is answered inline after relief planning

- **Given** a coordinator has finished assigning relief and has not yet answered this week's check-in
- **When** they complete their last assignment for the day
- **Then** they are shown both the time-spent-per-day question and the 1 to 5 ease-of-planning question inline in one prompt, and their answers are stored against their school and the date

### Relief planning is never blocked by an unanswered survey

- **Given** a coordinator is shown any inline survey
- **When** they dismiss it or ignore it
[...]

Issue #488, opened 2026-07-03
refactor: timetable parsers duplicate ~500 lines and autoAssign.ts is an 879-line god module


guangshinhaha:
## Summary
Two long-lived duplication/size hotspots raise the cost of every change in the timetable-import and auto-assign areas. Tracking issue — low priority, not pilot-blocking; captured so the debt is visible before the next feature lands on top of it.

## Location
- **Three timetable parsers** implement near-identical grid→period-entry logic (column detection, cell parsing, subject classification): `src/lib/parseAscPdf.ts`, `src/lib/parseAscXml.ts`, `src/lib/parseGridCsv.ts` (~500 lines combined). A fix to one (e.g. room-parsing thresholds) must be replicated in the others, and they drift.
- **`src/lib/autoAssign.ts` is ~879 lines** — an order of magnitude above the `src/lib` median — bundling absence-slot collection, coverage dedup, the rule cascade, weighted ranking, the 2-opt pass, and result formatting in one module. Stage transitions are hard to test in isolation and new rules require deep edits.

## Impact
Higher change cost and regression risk in two actively-evolving areas (both are churn hotspots in `git log`). Not a correctness bug today — a maintainability tax.

## Fix direction
- Parsers: extract a shared intermediate representation (`type Cell = { content: string[]; rooms: string[] }`) and a `parseGridToCells(...)` step; each format adapts its raw input to a grid and reuses the shared middle layer, with per-format quirks kept in thin adapters. Unit-test the shared layer once.
- `autoAssign.ts`: extract each labelled stage into a named, individually-testable pure function returning an explicit type, wired in sequence by a thin `autoAssign()` coordinator.
Do this incrementally, behind the existing test suite, not as a big-bang rewrite. Suggest splitting into two PRs (parsers; algorithm) if picked up.

---
*Filed from the 2026-07 full-app audit (DX/tests/perf pass). Low leverage / L effort — tracked, not urgent.*

Issue #487, opened 2026-07-03
dx: eslint is configured but enforced nowhere (no hook, no CI)


guangshinhaha:
## Summary
`npm run lint` is defined but invoked in no git hook and no CI workflow, and `eslint.config.mjs` carries only the `eslint-config-next` defaults with no project rules. The configured linter provides no feedback loop.

## Location
- `package.json` — `"lint": "eslint"` exists.
- `.husky/pre-commit` — enforces custom greps (no `any`, no raw SQL, Dockerfile checks) but never calls `eslint`.
- `.husky/pre-push` — runs `npm audit`, `tsc --noEmit`, `npm test`; no lint.
- `.github/workflows/` — no workflow runs lint.
- `eslint.config.mjs` — `next/core-web-vitals` defaults only; no custom rules or formatter (`.prettierrc` absent).

## Impact
Style and quality drift (unused imports, exhaustive-deps violations, accessibility lints from the Next config) accumulate with no signal. Low severity, but a cheap hygiene gap given the config is already present.

## Fix direction
Add lint to the CI test workflow (preferred — keeps hooks fast), or add `lint-staged` to `.husky/pre-commit` for changed files only. Optionally adopt a formatter (Prettier or Biome) — see the repo's `aif-lint-setup` skill, which is designed for exactly this. Keep scope to enabling enforcement; a separate pass can tighten rules.

---
*Filed from the 2026-07 full-app audit (DX/tests/perf pass). Verified against hooks, workflows, and eslint config.*

PR #647, opened 2026-08-11
feat(analytics): pilot metrics dashboard in Looker Studio over GA4

guangshinhaha:
Part of #642. **Deliberately does not close it:** every one of #642's acceptance criteria is about the Looker Studio report, which is built in Google's UI and does not exist yet. This PR builds the data sources that report needs. The issue should stay open until the report is created and its URL is recorded in `docs/analytics/pilot-metrics-report.md`.

> **Rebased onto `main`** on 2026-08-11, after #640 landed as #646 (squash-merged, `9ecdded`). Only this PR's five commits were replayed (`git rebase --onto origin/main b472d49`), so none of #640's content is duplicated here. Rebase was conflict-free and the diff is byte-identical to the pre-rebase one.

## Summary

- Sends **period coverage rate** (scorecard metric 5) to GA4 nightly from the server. It is the only metric with no user action behind it: an uncovered period is the *absence* of an event, so it cannot be counted client-side.
- Sends **survey scores** (metrics 3, 6 and 7) to GA4 when a coordinator answers. These had no source at all: see the finding below.
- Specifies the Looker Studio report itself in `docs/analytics/pilot-metrics-report.md`, since the report has no exportable source and cannot be version controlled.

## Finding: metrics 3, 6 and 7 had no source

#642's flow diagram shows survey responses feeding the GA4 property, and #641's body says the same. What shipped in PR #643 writes to Postgres and stops. Verified on `origin/main`: no GA or Measurement Protocol call anywhere under `src/lib/surveys.ts` or `src/app/api/surveys/**`.

Reopening #641 would not have helped, since the sender it needs does not exist until #640 lands. This PR takes the ~20 line emit and puts it next to the report that consumes it.

## Grooming decisions carried into the code

| Decision | Why |
|---|---|
| Coverage rate uses option 1 (nightly Measurement Protocol event), not a blended second Looker source | Option 2 would need Looker to read the database, which #642's own Out of scope forbids |
| Only `survey_answered` is emitted | Response rates are #641's AC and `/admin/survey-results` already serves them |
| Idempotency guarded by a database row | GA4 has no upsert; a double-fired cron would inflate metric 5 silently |
| No feature flag | The emitter is already inert when GA is unconfigured |
| Counts only, never percentages | Same reasoning that rejected `*_pct` params on #640 |

## Two integration findings that changed the spec

**The cron has no `_ga` cookie and no acting user.** `resolveClientId` returns null for that combination and `sendServerEvent` drops the event, so a constant `fallbackSeed` is load-bearing rather than decorative. That creates one synthetic GA client, which GA4 still counts in `activeUsers`, so **metric 1 must be measured over clients that fired a `page_view`**. The synthetic client never does. This is documented in the report spec and commented at the constant.

**`sendServerEvent` returns `void` and never throws**, so the cron cannot observe a GA rejection. Rather than edit `src/lib/analytics.ts` and collide with #640's branch, `failed` counts schools whose *coverage computation* threw, which is observable and worth retrying. GA-side rejections are logged by the sender, with a documented delete-row-and-re-emit repair path.

## Changes

| File | Why |
|---|---|
| `src/lib/scheduling/coverageRate.ts` (new) | `computeDailyCoverage`, lifted out of `daily-summary` so the summary email and the metrics event cannot report different numbers. Hard constraint: no third copy (#488) |
| `src/app/api/cron/daily-summary/route.ts` | Now calls the helper. 90 lines down to a destructure. Behaviour unchanged |
| `src/app/api/cron/pilot-coverage/route.ts` (new) | The nightly emitter |
| `prisma/schema.prisma` | `PilotMetricEmission`, unique on (school, metric, period) |
| `src/lib/constants/analyticsEvents.ts` | Two event names, four dimensions, four metrics, and the synthetic client seed |
| `src/lib/surveys.ts` | `respondToPrompt` returns the emit payload, or null when already answered |
| `src/app/api/surveys/[id]/route.ts` | Queues `survey_answered`, in its own try/catch so an analytics failure cannot turn a recorded answer into a 500 |
| `docs/analytics/pilot-metrics-report.md` (new) | Metric formulas, dimension registry, empty-state rules, thresholding checklist, cron runbook |
| `tests/api/surveys.test.ts` | **Unrelated repair.** Failing 18/18 since #641 landed: it calls `mockSession` without declaring the `next/headers` mock that `vi.mocked(cookies)` needs. It sits on top of the code changed here and left the change unverifiable |

## Test plan

| Acceptance criterion | Covered by |
|---|---|
| Report shows every metric for a school and date range | `surveys-analytics.test.ts` (metrics 3/6/7 payloads incl. TIME_SPENT, EASE_OF_PLANNING, NPS), `cron-pilot-coverage.test.ts` (metric 5), plus the formula table in the report spec. Metrics 1/2/4 are #640's events |
| Coverage rate appears alongside the behavioural metrics | `cron-pilot-coverage.test.ts` "sends one coverage event per school with the day's slot counts" |
| Distinguishes "no data yet" from a genuine zero | `cron-pilot-coverage.test.ts` "emits zeroes for a school with no absences", plus the denominator table in the report spec |
| Per-school figures survive GA4 thresholding | **Not automatable.** Needs real pilot-volume traffic. Lands as a sign-off checklist in the report spec |
| Per school, no personal data | `cron-pilot-coverage.test.ts` and `surveys-analytics.test.ts` both assert directly on the serialised payloads |

Additional scenarios covered: idempotent re-run, different date after one is recorded, one school failing without stopping the others and being retried alone, unconfigured analytics no-op, cron auth 401, invalid date 400, weekend skip, double submit emitting once, dismissal emitting nothing, analytics failure not failing the request, `_ga` cookie passthrough, multi-tenancy isolation in `computeDailyCoverage`, and the #285 cycle-aware week regression.

**Results**

- `npm test`: **1033 passed, 2 skipped, 0 failed** (96 files), re-run after the rebase.
- `npm run build`: clean. `Compiled successfully`, `Generating static pages (97/97)`, no prerender errors. `/api/cron/pilot-coverage` registered.
- `npm audit`: 0 vulnerabilities. No dependency changes.
- `npm run test:api`: the suite has a **pre-existing red baseline**. Measured back to back against the same `main` this PR now targets:

  | | Failures | Files |
  |---|---|---|
  | `origin/main` (`9ecdded`) | 73 | 20 |
  | This branch | **55** | **19** |

  The failing file sets are **identical except `tests/api/surveys.test.ts`**, which this PR repairs. The 18-test delta is exactly that file. This PR adds no new failures, and all five files it touches or adds pass.

## Required before this can merge

1. ~~Generate the migration.~~ **Done** (`20260811070053_add_pilot_metric_emission`). Generated by `prisma migrate dev` against the local dev database, not hand-authored, so Prisma owns the SQL and the timestamp. No reset was needed: the dev DB had nine pending migrations and zero drift, so they applied cleanly first.
2. **Register the remaining custom dimensions and metrics in GA4 before any event fires.** As of 2026-08-11 the property has #640's three dimensions and eight metrics. **None of this PR's eight are registered yet:** dimensions `coverage_date`, `instrument`, `cadence_period`, `role`; metrics `needed_slots`, `covered_slots`, `absent_teachers`, `score`. Set the unit of measurement to Standard for all four metrics, `score` included: it carries minutes only for `TIME_SPENT`, and is 1 to 5 for `EASE_OF_PLANNING` and 0 to 10 for `NPS`. GA4 does not backfill: an event fired against an unregistered parameter is permanently unusable. The full list is in `docs/analytics/pilot-metrics-report.md`.

> Note on the red Vercel check: it fails with *"Cannot deploy from a private GitHub organization repository on the Hobby plan"*, which is an account plan limit, not a code failure. It is failing identically on every recent `main` commit. The real CI job (`build-and-push`) is green on `main` for the #640 merge, and does not run on pull requests (#482).

## Deliberate deviations from CLAUDE.md

- **The migration was generated by Claude rather than handed to a human**, at the maintainer's explicit request. CLAUDE.md says to hand the command over; the rule's stated concern is hand-authored SQL and hand-picked timestamps, and `prisma migrate dev` is Prisma generating both, so the intent is preserved. Flagged here so the rule can be revised if that reading is wrong.
- **No seed rows for `PilotMetricEmission`.** A seeded emission makes the cron skip that (school, date), so seed data would suppress the behaviour it is meant to illustrate.
- **ESLint was not run.** It is broken in this tree with `TypeError: expand is not a function` from `minimatch`, independent of this branch. Tracked by #487.

---

> **Before marking ready for review**: trigger the cron by hand against a school with absences and confirm the emitted payload, then answer an inline survey and confirm one `survey_answered` event.

*Generated with aif-implement-issue*


src/app/api/cron/pilot-coverage/route.ts
```diff
@@ -0,0 +1,143 @@
...
+    select: { id: true, timetableCycle: true, weekOneAnchorDate: true },
+  });
+
+  let emitted = 0;
+  let skipped = 0;
+  let failed = 0;
+
+  for (const school of schools) {
+    try {
+      const existing = await prisma.pilotMetricEmission.findUnique({
```
guangshinhaha: **[Medium] The idempotency guard is check-then-act, so two overlapping runs both send the event it exists to deduplicate**

```ts
const existing = await prisma.pilotMetricEmission.findUnique({ // ←
  where: { schoolId_metricKey_periodKey: { schoolId: school.id, metricKey: ANALYTICS_EVENTS.DAILY_COVERAGE, periodKey: date } },
  select: { id: true },
});
if (existing) { skipped++; continue; }

const coverage = await computeDailyCoverage(school, date);
await sendServerEvent({ /* ... */ });
await prisma.pilotMetricEmission.create({ /* ... */ });
```

**Problem:** The read and the write are separate statements with two `await`s (and a network call) between them, so the unique constraint never gets a chance to arbitrate. Trigger: the nightly EventBridge/ECS schedule fires with at-least-once delivery, or an operator runs a `?date=` backfill for today while the nightly run is in flight. Both invocations `findUnique` → null, both `sendServerEvent`, and GA4 now holds two `daily_coverage` events for the same (school, date). Because every Looker tile uses a plain `SUM` on the strength of this table (per `docs/analytics/pilot-metrics-report.md`), `needed_slots` and `covered_slots` are both doubled for that school-day — exactly the silent inflation the `PilotMetricEmission` docblock says the constraint prevents. The loser's `create` then throws P2002, lands in the catch, and is reported as `failed: 1`, so the run looks like it under-emitted when it actually over-emitted.

**Suggestion:** Let the unique index do the arbitration — claim the slot before sending, and treat P2002 as "somebody else has it".

```ts
const coverage = await computeDailyCoverage(school, date);

try {
  await prisma.pilotMetricEmission.create({
    data: { schoolId: school.id, metricKey: ANALYTICS_EVENTS.DAILY_COVERAGE, periodKey: date },
  });
} catch (err) {
  // P2002 = another run already claimed this (school, metric, date).
  if (err instanceof Prisma.PrismaClientKnownRequestError && err.code === "P2002") {
    skipped++;
    continue;
  }
  throw err;
}

await sendServerEvent({ /* ... unchanged ... */ });
emitted++;
```

Computing the coverage first preserves the property the current comment cares about: a school whose coverage query throws still ends the run with no row and is retried next time.

---

src/app/api/surveys/[id]/route.ts
```diff
@@ -25,7 +31,44 @@ export async function POST(request: NextRequest, { params }: Params) {
...
+              [ANALYTICS_DIMENSIONS.SCHOOL_ID]: recorded.schoolId,
+              [ANALYTICS_DIMENSIONS.INSTRUMENT]: recorded.instrument,
+              [ANALYTICS_DIMENSIONS.CADENCE_PERIOD]: recorded.cadencePeriod,
+              [ANALYTICS_DIMENSIONS.ROLE]: session.user.role,
+              [ANALYTICS_METRICS.SCORE]: recorded.score,
+            },
+            // Reusing the browser's own client ID keeps this event and the
+            // same visit's page views attributed to one user, rather than
+            // inflating metric 1 with a phantom.
+            gaCookie: request.cookies.get("_ga")?.value ?? null,
```
Comment (guangshinhaha) [inline src/app/api/surveys/[id]/route.ts]: **[Medium] Hand-rolled `_ga` read drops the session cookie, so `survey_answered` lands in a different GA4 session from the coordinator's visit**

```ts
queueServerEvent({
  name: ANALYTICS_EVENTS.SURVEY_ANSWERED,
  params: { /* ... */ },
  // Reusing the browser's own client ID keeps this event and the
  // same visit's page views attributed to one user, rather than
  // inflating metric 1 with a phantom.
  gaCookie: request.cookies.get("_ga")?.value ?? null, // ←
  fallbackSeed: session.user.id,
});
```

**Problem:** `ServerEventInput` carries two cookies, not one — `gaCookie` (client ID) and `gaSessionCookie` (session ID, read from the per-stream `_ga_<STREAM>` cookie whose name only `getAnalyticsConfig()` can derive). This call supplies only the first. `sendServerEvent` then calls `resolveSessionId({ gaSessionCookie: undefined })`, which falls through to `deriveFallbackSessionId()` — a 30-minute wall-clock bucket shared by every server event in that window.

Concretely: a coordinator with the GA tag running answers the weekly ease-of-planning prompt. Their `page_view` events carry the browser's real `session_id`; the `survey_answered` event carries a time-bucket integer instead. GA4 sees one client with two sessions, so the answer is orphaned from the visit that produced it, session counts inflate, and any Looker breakdown that joins `survey_answered` to session-scoped context (landing page, session source, engaged sessions) silently drops it. Worse, two different coordinators answering inside the same 30-minute window are assigned the *same* `session_id` under different client IDs.

This is also a reuse miss: `readGaCookies` exists for exactly this and is what #640's two other callers use — `src/app/api/absence-reports/route.ts:635` (`...readGaCookies(request.cookies)`) and `src/app/api/auto-assign/route.ts:95`.

**Suggestion:**
```ts
import { queueServerEvent, readGaCookies } from "@/lib/analytics";

// ...
queueServerEvent({
  name: ANALYTICS_EVENTS.SURVEY_ANSWERED,
  params: { /* ... unchanged ... */ },
  // Reusing the browser's own client and session IDs keeps this event and
  // the same visit's page views attributed to one user and one session.
  ...readGaCookies(request.cookies),
  fallbackSeed: session.user.id,
});
```

`tests/api/surveys-analytics.test.ts` "reuses the browser's own GA client id" only asserts `events[0].gaCookie`, which is why this passed — worth extending it to assert `gaSessionCookie` too.

---

src/lib/scheduling/coverageRate.ts
```diff
@@ -0,0 +1,173 @@
...
+  const coveredSlots = assignments.length;
+
+  return {
+    activeAbsences,
+    assignments,
+    absentTeacherCount: new Set(activeAbsences.map((a) => a.teacherId)).size,
+    neededSlots,
+    coveredSlots,
+    uncoveredSlots: neededSlots - coveredSlots,
+    coverageRatePct:
```
guangshinhaha, line 234: **[Medium] `coverageRatePct` is documented as "0 to 100" but can exceed 100, and `uncoveredSlots` can go negative**

```ts
const assignments = await prisma.reliefAssignment.findMany({
  where: { schoolId: school.id, date: dateUTC, status: "ACTIVE" }, // ← no tie to activeAbsences
  include: ASSIGNMENT_INCLUDE,
});
// ...
const coveredSlots = assignments.length;

return {
  neededSlots,
  coveredSlots,
  uncoveredSlots: neededSlots - coveredSlots,            // ← can be negative
  coverageRatePct:
    neededSlots > 0 ? Math.round((coveredSlots / neededSlots) * 100) : 100, // ← can exceed 100
};
```

**Problem:** `neededSlots` is filtered three ways — `coverageType: { not: "NO_RELIEF" }`, then `latestAbsencePerTeacher` (which drops superseded reports), then `requiredPeriodsForAbsence` (RELIEF_REQUIRED periods only). `coveredSlots` is filtered none of those ways: it is every ACTIVE `ReliefAssignment` on the date for the school.

Reachable state: a teacher's absence is assigned relief for periods 1–3, then the coordinator revises the report down to NO_RELIEF (or files a superseding report covering fewer periods) without cancelling the assignments. `neededSlots` drops to 0–1 while `coveredSlots` stays at 3 → `uncoveredSlots = -2` and `coverageRatePct = 300`.

Before this PR that only mis-rendered one SLT email. This PR promotes the same arithmetic into scorecard metric 5, whose Looker formula is `sum(covered_slots) / sum(needed_slots)` against a **>= 99%** target — so a single stale assignment can push a school over 100% and mask genuinely uncovered periods in the aggregate. The `DailyCoverage` docblock added here asserts a range the function does not enforce, so the next caller will trust it.

**Suggestion:** Count only assignments that belong to an absence the denominator counted.

```ts
const countedAbsenceIds = new Set(activeAbsences.map((a) => a.id));
const coveredSlots = assignments.filter((a) =>
  countedAbsenceIds.has(a.absenceReportId),
).length;
```

If changing the email's long-standing number is out of scope for this PR, the cheaper option is to clamp at the boundary and correct the docblock:

```ts
const coveredSlots = Math.min(assignments.length, neededSlots);
```

> Tradeoff: the filter is the correct fix and keeps `assignments` (which the email renders) untouched; the clamp is a one-liner but hides the underlying data inconsistency rather than reporting it.

---

src/app/api/cron/pilot-coverage/route.ts
```diff
@@ -0,0 +1,143 @@
...
+        fallbackSeed: ANALYTICS_SYSTEM_CLIENT_SEED,
+      });
+
+      // Written after the send. `sendServerEvent` swallows delivery failures by
+      // design (an analytics failure must never change a request's outcome), so
+      // this cannot distinguish a delivered event from a rejected one. What it
+      // does guarantee is that a school whose coverage query threw has no row
+      // and is retried on the next run. A GA-side rejection is logged by the
+      // sender and repaired by deleting the row and re-triggering.
+      await prisma.pilotMetricEmission.create({
```
guangshinhaha: **[Medium] A GA outage burns every school's idempotency row for that night, and the reason given for not fixing it no longer holds**

```ts
await sendServerEvent({ /* ... */ });   // never throws — swallows fetch errors AND non-2xx

// Written after the send. `sendServerEvent` swallows delivery failures by
// design ... A GA-side rejection is logged by the sender and repaired by
// deleting the row and re-triggering.
await prisma.pilotMetricEmission.create({ /* ... */ });  // ←
emitted++;                                               // ←
```

**Problem:** `sendServerEvent` (`src/lib/analytics.ts:76–133`) wraps everything in a `try`, logs, and returns `void` on both a thrown fetch (timeout / DNS / connection reset) and a non-2xx from the Measurement Protocol. So this route writes the "already emitted" row and increments `emitted` for events GA never accepted.

Concrete failure: `GA_MEASUREMENT_PROTOCOL_SECRET` is rotated and the new value has not reached the cron task yet. Every send gets a 401, every send logs and returns, every school gets a `PilotMetricEmission` row, and the route responds `{ success: true, emitted: 12, failed: 0 }`. The next night skips all twelve as already-emitted. Metric 5 has a permanent one-day hole for every school, and the only signals are `console.error` lines in the task log — nothing in the response, nothing that a retry would repair. The `failed` counter, as the PR body notes, only ever counts coverage-computation errors, so it stays 0 through the entire outage.

The stated reason for not fixing this — "rather than edit `src/lib/analytics.ts` and collide with #640's branch" — is stale: #640 landed as #646 (`9ecdded`) and this branch is already rebased onto it, so there is nothing left to collide with.

**Suggestion:** Have the sender report delivery, and only claim the slot on success.

```ts
// src/lib/analytics.ts — still never throws, just reports.
export async function sendServerEvent(input: ServerEventInput): Promise<boolean> {
  try {
    // ... unchanged ...
    if (!response.ok) {
      console.error(`[analytics] ${input.name} rejected with status ${response.status}`);
      return false;
    }
    return true;
  } catch (err) {
    console.error(`[analytics] failed to send ${input.name}`, err);
    return false;
  }
}
```

```ts
// src/app/api/cron/pilot-coverage/route.ts
const delivered = await sendServerEvent({ /* ... */ });
if (!delivered) {
  failed++;                       // no row → retried on the next run
  continue;
}
await prisma.pilotMetricEmission.create({ /* ... */ });
emitted++;
```

`queueServerEvent` ignores the return value, so no other caller changes. If the sender must stay `void` for this PR, at minimum make `failed` distinguish delivery from computation so the outage is visible in the response rather than only in logs.

---

src/lib/scheduling/coverageRate.ts
```diff
@@ -0,0 +1,173 @@
...
+/**
+ * Coverage for one school on one school day.
+ *
+ * Does no weekend check and no feature-flag check: both are the caller's
+ * business, and they differ between the two callers.
+ */
+export async function computeDailyCoverage(
+  school: CoverageSchool,
+  date: string,
+  prisma: PrismaClient = defaultPrisma,
```
guangshinhaha, line 131: **[Low] The `prisma` parameter is only honoured for two of the four queries this function issues**

```ts
export async function computeDailyCoverage(
  school: CoverageSchool,
  date: string,
  prisma: PrismaClient = defaultPrisma,   // ←
): Promise<DailyCoverage> {
  // ...
  const absenceReports = await prisma.absenceReport.findMany({ ... });   // uses the parameter
  const assignments   = await prisma.reliefAssignment.findMany({ ... }); // uses the parameter
  const entries       = await timetableEntriesForDate({ ... });          // uses the module singleton
```

**Problem:** `timetableEntriesForDate` imports `prisma` from `@/lib/prisma` at module scope (`src/lib/scheduling/timetableResolver.ts:1`) and issues two of its own queries — `timetableOverride.findFirst` and `timetableEntry.findMany` — against that singleton regardless of what is passed here. The injection point therefore covers half the read set.

Failure it invites: the first caller that wraps this in `prisma.$transaction((tx) => computeDailyCoverage(school, date, tx))` gets absences and assignments from inside the transaction and timetable entries from outside it, so the coverage figure is computed across two snapshots — non-deterministic and near-impossible to reproduce. `tests/api/coverageRate.test.ts` passes `testPrisma` and passes only because `DATABASE_URL` already points the singleton at the same port-5433 database, so the suite gives no warning that the parameter is partial.

**Suggestion:** Either thread the client all the way through, or drop the parameter and let callers rely on the singleton (which is what both current callers do).

```ts
// Option A — thread it through (timetableEntriesForDate gains an optional client):
const entries = await timetableEntriesForDate({
  schoolId: school.id,
  date,
  dayOfWeek,
  weekType,
  teacherIds: activeAbsences.map((a) => a.teacherId),
  prisma,
});

// Option B — if A is out of scope, document the limit so nobody passes a tx:
/**
 * @param prisma Overrides the client for the absence and assignment reads only.
 *   `timetableEntriesForDate` still uses the module singleton, so this must NOT
 *   be a `$transaction` client.
 */
```

> Tradeoff: A is the real fix but touches a resolver used across the scheduling code; B costs nothing and stops the trap being sprung.

---

src/app/api/cron/daily-summary/route.ts
```diff
@@ -53,82 +50,19 @@ export async function GET(request: NextRequest) {
...
+    // #642 — the slot arithmetic lives in `computeDailyCoverage` so this
+    // email and the nightly pilot-metrics event report the same figure.
+    const {
+      activeAbsences: activeSick,
+      assignments,
+      neededSlots: totalSlots,
+      coveredSlots: covered,
+      uncoveredSlots: uncovered,
+      coverageRatePct: coverageRate,
+    } = await computeDailyCoverage(school, dateStr);
```
guangshinhaha, line 62: **[Low] The refactor moves the `continue` after the expensive work, so quiet schools now pay for three queries they used to skip**

```ts
for (const school of schools) {
  const {
    activeAbsences: activeSick,
    assignments,
    // ...
  } = await computeDailyCoverage(school, dateStr);  // ← 4 queries, unconditionally

  if (activeSick.length === 0) continue;            // ← guard now runs after, not before
```

**Problem:** Before this PR the loop queried `absenceReport` only, hit `if (activeSick.length === 0) continue;`, and skipped the `reliefAssignment.findMany` and `timetableEntriesForDate` calls entirely. `computeDailyCoverage` now issues all four reads (absences, assignments, `timetableOverride.findFirst`, `timetableEntry.findMany`) before the caller gets a chance to bail.

For the common case — a school with `summaryEmails` configured and nobody absent that day — that is three extra round trips per school per run, two of which return nothing useful (`teacherIds: []` guarantees an empty entry set). Cheap at pilot scale, but it is a short-circuit that was deliberately there and is silently gone, and it grows linearly with the school count.

**Suggestion:** Give `computeDailyCoverage` an early exit so both callers benefit — the pilot-coverage cron wants zeroes for a quiet school anyway, and this returns exactly those without touching the timetable.

```ts
const activeAbsences = Array.from(latestAbsencePerTeacher(absenceReports).values());

if (activeAbsences.length === 0) {
  // Nothing needs cover: the assignment and timetable reads cannot change any
  // of these figures, so skip them.
  return {
    activeAbsences: [],
    assignments: [],
    absentTeacherCount: 0,
    neededSlots: 0,
    coveredSlots: 0,
    uncoveredSlots: 0,
    coverageRatePct: 100,
  };
}
```

Note this only holds together once the `coveredSlots` filtering in the other comment on this file is applied — as written, a school with zero counted absences can still have ACTIVE assignments on the date, and the early return would report `coveredSlots: 0` where the current code reports a non-zero count.

---

src/app/api/cron/pilot-coverage/route.ts
```diff
@@ -0,0 +1,143 @@
...
+  ANALYTICS_DIMENSIONS,
+  ANALYTICS_EVENTS,
+  ANALYTICS_METRICS,
+  ANALYTICS_SYSTEM_CLIENT_SEED,
+} from "@/lib/constants/analyticsEvents";
+import { computeDailyCoverage } from "@/lib/scheduling/coverageRate";
+import { isWeekend, todayInSGT } from "@/lib/scheduling/date";
+import { NextRequest, NextResponse } from "next/server";
+
+const DATE_PATTERN = /^\d{4}-\d{2}-\d{2}$/;
```
guangshinhaha: **[Low] The date guard validates shape but not calendar validity, so a typo'd backfill returns `success: true` while failing for every school**

```ts
const DATE_PATTERN = /^\d{4}-\d{2}-\d{2}$/;   // ←
// ...
if (dateParam !== null && !DATE_PATTERN.test(dateParam)) {
  return NextResponse.json({ error: "Invalid date. Expected YYYY-MM-DD." }, { status: 400 });
}
const date = dateParam ?? todayInSGT();

if (isWeekend(date)) { ... }
```

**Problem:** The pattern accepts impossible dates — `2026-02-30`, `2026-13-01`, `2026-00-00` all match. Trace `?date=2026-02-30`:

1. Regex passes, no 400.
2. `isWeekend` → `dayOfWeekForSchoolDate` → `new Date("2026-02-30T00:00:00").getDay()` → `NaN`. `NaN === 0 || NaN === 6` is false, so it is not treated as a weekend and the run proceeds.
3. `computeDailyCoverage` → `parseSchoolDateUTC` → Invalid Date → Prisma rejects `startDate: { lte: Invalid Date }`.
4. Every school throws into the per-school catch, so the response is `{ success: true, date: "2026-02-30", schoolsProcessed: 12, emitted: 0, skipped: 0, failed: 12 }` with HTTP 200.

An operator backfilling a missed night gets a 200 and a `success: true` for a request that did nothing, plus twelve stack traces in the task log — when the input was rejectable up front with a clear 400.

**Suggestion:** Round-trip the parsed date to confirm the calendar accepted it.

```ts
function isValidSchoolDate(value: string): boolean {
  if (!DATE_PATTERN.test(value)) return false;
  const parsed = new Date(`${value}T00:00:00.000Z`);
  return (
    !Number.isNaN(parsed.getTime()) &&
    parsed.toISOString().slice(0, 10) === value
  );
}

if (dateParam !== null && !isValidSchoolDate(dateParam)) {
  return NextResponse.json(
    { error: "Invalid date. Expected YYYY-MM-DD." },
    { status: 400 },
  );
}
```

The existing `tests/api/cron-pilot-coverage.test.ts` "rejects a date that is not YYYY-MM-DD" case covers `07-04-2026` only; `?date=2026-02-30` would make a good second assertion.

---

guangshinhaha:
## Code Review Summary

| Severity | Count |
|----------|-------|
| Critical | 0 |
| High     | 0 |
| Medium   | 4 |
| Low      | 3 |

Findings cluster on the nightly emitter's at-most-once guarantee (`src/app/api/cron/pilot-coverage/route.ts`), the GA session cookie on the survey emit, and a range contract in `computeDailyCoverage` that the function does not enforce.

## What Looks Good

- **Lifting `computeDailyCoverage` out of `daily-summary` rather than writing a second copy.** The summary email and the Looker tile now provably read the same arithmetic, and the module docblock names #488 as the reason — this is the right response to "I need this number in a second place", and the 90-line handler becoming a destructure is a real readability win on top.
- **`respondToPrompt` returning the emit payload instead of emitting from inside `src/lib/surveys.ts`.** Keeps that module database-only (its unit tests need no analytics mock), lets the route supply the role from the session it already holds, and makes the already-ANSWERED case a `null` rather than a silent success — which is what actually stops a two-tab double submit counting twice.
- **Emitting `daily_coverage` with zeroes for schools that had no absences.** This is what makes "nothing needed cover today" a reportable zero rather than a gap, and pairing it with an event-count denominator in the report spec is the correct way to satisfy the "no data yet vs genuine zero" criterion in Looker.
- **`docs/analytics/pilot-metrics-report.md` as a rebuild spec, not a description.** The dimension/metric registration table, the "GA4 does not backfill" warning, and the synthetic-client note explaining exactly why metric 1 must filter on `page_view` are the things that would otherwise be lost with the report's owner.

---

Issue #375, opened 2026-06-22
Security: TEACHER role can access school-admin dashboard + config APIs (missing role guard)


guangshinhaha:
## Summary (security — vertical privilege escalation)

A **TEACHER** can access the entire **school-admin dashboard** (`/{slug}/dashboard/*`, including **Settings**) and can call several admin-only config APIs. The admin surfaces authorize on *school membership*, not on *admin role* — `requireSchoolAdmin` exists but is not used by the dashboard pages or by several mutation APIs.

**Severity:** High. Vertical privilege escalation within a tenant (confidentiality + integrity). Cross-tenant isolation is **not** affected (the slug/membership check still holds).

## How it was found

A SUPERADMIN impersonating `maia-teacher-1b@gmail.com` (TEACHER) landed on the admin **Relief Overview** with the full Settings nav (School Profile, Periods & Schedule, Staff Positions, Auto-Assign Rules, Role Priority).

This is **not** an impersonation artifact. Impersonation correctly downgrades the effective role to TEACHER (`getSession` swaps `session.user` to the impersonated user — `src/lib/auth.ts:80-91`). The dashboard simply never checks role, so a **real teacher logging in directly hits the same surface**.

## Root cause

No layer enforces `role === SCHOOL_ADMIN` for `/{slug}/dashboard/*`:

1. **Middleware** (`src/middleware.ts:71-76`) gates `/{slug}/dashboard/*` on **session cookie presence only** — comment says role check is "done server-side", but it isn't for this tree.
2. **Layouts** — `src/app/[slug]/layout.tsx` doesn't even require a session; `src/app/[slug]/dashboard/layout.tsx` is purely presentational (no auth).
3. **Pages** — every dashboard/settings page uses `requireSchoolForSlug` → `requireSchool` → `requireAuth`, which checks *session + has-school + slug-match*, **never role** (`src/lib/auth.ts:108-122`).
4. **Login redirect** actively sends teachers into it: `verify-otp` redirects TEACHER → `/${slug}/dashboard` (the admin dashboard); there is no teacher redirect to `/{slug}/teacher` or `/{slug}/reporting` (`src/app/api/auth/verify-otp/route.ts:172-178`).

`requireSchool` reads like "is this user allowed in this school" but only means *membership*. It's being used where `requireSchoolAdmin` (which checks `role === SCHOOL_ADMIN`, `src/lib/auth.ts:132-138`) is intended.

## Blast radius

[...]

PR #677, opened 2026-08-17
feat(dashboard): give KPs their own relief surface and fix the teacher notification bell

guangshinhaha:
## Summary

Two problems, both from watching a teacher/KP use the pilot.

**1. The teacher notification bell was unusable.** The dropdown panel is `w-80` (320px) anchored `right-0`, and the bell had been mounted inside `TeacherSidebar` — which is `w-56` (224px). The panel overflowed off the left edge of the viewport, so half of it (including "Mark all read" and the close button) was unreachable. The teacher surface was also the only shell in the app with no desktop top bar.

**2. A KP who also teaches had to bounce between two apps.** Every KP (`SCHOOL_ADMIN` / `RELIEF_COORDINATOR`) is teaching staff by default, but their own relief duties and absence reporting lived only at `/[slug]/teacher` — a separate route with its own shell and nav. Reporting their own MC meant leaving the dashboard, and then **searching for their own name** in a dropdown of the entire staff roster, while a plain teacher gets theirs filled in automatically.

### What changed

- **Teacher desktop top bar.** `DesktopTopBar` moved from `dashboard/` to `_components/` and is now shared by both shells rather than copied. The bell and Log out live there on both surfaces; `TeacherSidebar`'s own logout block is gone. The panel now has a full-width parent to open under.
- **"My Relief" dashboard destination.** A new nav item at the bottom of the KP sidebar, below a divider, expanding to **My Duties** / **Report Absence** — the same expandable `?tab=` pattern the sidebar already uses for Relief Teachers. Duties get full prev/next day navigation. Rendered only when the signed-in user has a linked `Teacher` row **at this school**; pure coordinators see nothing.
- **The dashboard's "Report Absence" button** now lands on that tab instead of `/[slug]/reporting`.
- **A teaching KP is pre-selected as themselves** on the absence form — via a new `defaultTeacherId` prop, deliberately distinct from the existing `lockedTeacherId`: *pre-selected and changeable*, not read-only, because filing on behalf of a colleague is still a frequent KP task.
- **A KP keeps dashboard chrome while reporting.** `/[slug]/reporting` sits outside the `/dashboard` route segment and inherited no admin layout, so it wrapped everyone in `TeacherShell`. Right for teachers, inverted for KPs. The dashboard chrome was extracted into `DashboardShell` and the shell is now chosen by access kind.

## Why

`Teacher.userId` (unique FK) already linked a KP to their teacher row, and `GET /api/teacher-dashboard` already gated on *that link* rather than on `role === TEACHER` — so a teaching KP was already served by it. Nothing was missing from the data model; what was missing was any dashboard surface that asked. This branch changes **no schema and no API**.

Design doc: `docs/superpowers/specs/2026-08-17-kp-teacher-ia-unification-design.md`
Plan: `docs/superpowers/plans/2026-08-17-kp-teacher-ia-unification.md`

### A note on the first attempt

The panel was originally stacked on top of the dashboard home, above `ReliefOverview`. That demoted the dashboard's actual job-to-be-done and looked cluttered, and it was rejected on review of the running app. The sidebar-destination shape in this PR is the replacement. `MyTeachingPanel` is deleted, not left behind.

## Test plan

- `npm test` — **1069 passed**, 2 skipped, 99 files. Includes 9 unit cases for `resolveReportingAccess` (covering `self` populated / null / cross-school) and 3 for the new `myTeachingAccess` cross-tenant gate predicate.
- `npm run test:api -- teacher-dashboard` — **12 passed**. First-ever coverage for `GET /api/teacher-dashboard`: 401 unauthenticated, 403 for a user with no teacher row, 200 with correct shape for a `SCHOOL_ADMIN` **with** one, empty-day, and multi-tenancy isolation (adversarial — a same-named teacher in the other school).
- `npm run build` — exit 0, `✓ Generating static pages (99/99)`, no `⨯`.
- `npx tsc --noEmit` — exit 0.
- `npm audit` — 0 vulnerabilities.
- Verified against a locally running instance seeded from `prisma/seed.ts`, for three personas: a KP with a teacher row, a coordinator without one, and a plain teacher. Confirmed the nav item renders/suppresses correctly, the form pre-selects, the KP keeps dashboard chrome on `/reporting`, and a signed-out visitor still leaks no roster.

### Not covered by automated tests

Playwright coverage was attempted and dropped. `e2e/fixtures/auth.ts` sets a `reliefcher_demo` cookie against the static `/demo/*` route tree, which is DB-independent and does not import the changed components — the cases are unwritable without net-new session-minting infrastructure inside the auth boundary. That is its own design task, not something to bolt on here. Filing separately.

`docs/superpowers/plans/2026-08-17-kp-teacher-ia-manual-verification.md` in this PR is the human checklist standing in for it, with seeded personas named. **Step 6 (signed-out / wrong-school must still render no roster, no teacher names, no period structure) is the one not to skip** — it guards the issue #330 fix.

## Screenshots

_To add: teacher sidebar + top bar before/after; KP sidebar with My Relief expanded; the Report Absence tab._

## Follow-ups (not in this PR)

- e2e harness cannot mint a real `reliefcher_session` — blocks all e2e on authenticated `/[slug]/*` routes.
- `e2e/report.spec.ts` navigates to `/demo/report`, which 404s (real route is `/demo/reporting`). Pre-existing.
- `lockedTeacherId` is not membership-checked against the rendered roster the way `defaultTeacherId` now is — same class of issue, pre-existing, lower risk since that path locks rather than pre-selects.


src/app/[slug]/dashboard/ReliefOverview.tsx
```diff
@@ -351,7 +351,7 @@ export default function ReliefOverview({
           )}
         </div>
         <Link
-          href={`/${slug}/reporting`}
+          href={`/${slug}/dashboard/my-relief?tab=report`}
```
Comment (guangshinhaha) [inline src/app/[slug]/dashboard/ReliefOverview.tsx]: **[High] The dashboard "Report Absence" CTA now dead-ends for any KP without a linked Teacher row**

```tsx
        <Link
          href={`/${slug}/dashboard/my-relief?tab=report`}   // ←
          className="inline-flex items-center gap-2 rounded-lg bg-primary px-4 py-2 ..."
        >
          <Form className="h-4 w-4" />
          Report Absence
        </Link>
```

**Problem:** This href is unconditional, but `my-relief/page.tsx:59-61` bounces anyone who fails `canShowMyTeaching` straight back to `/${slug}/dashboard`:

```ts
const linkedTeacher = await getLinkedTeacherForUser(session.user.id);
if (!linkedTeacher || !canShowMyTeaching(linkedTeacher, schoolId)) {
  redirect(`/${slug}/dashboard`);
}
```

So a **pure `RELIEF_COORDINATOR`** (no Teacher row — the case this branch is careful to handle everywhere else) or a **`SUPERADMIN` viewing a school** clicks "Report Absence" on the dashboard and lands back on the dashboard. Before this PR the button went to `/${slug}/reporting`, which serves exactly that user the unlocked roster dropdown.

That is not a cosmetic dead link: `grep "Report Absence"` shows the only other dashboard-side entry points are the "My Relief → Report Absence" sidebar sub-link and the `my-relief` tab strip — both gated on the same `showMyRelief`. For a pure coordinator this removes **every** UI path to the absence form, and CLAUDE.md's RBAC intent explicitly lists "may file absences on behalf of staff" as a `RELIEF_COORDINATOR` capability.

**Suggestion:** pick the href from the same signal the nav uses. `dashboard/page.tsx` already awaits `requireSchoolForSlugOrRedirect`, and `getLinkedTeacherForUser` is request-cached, so this costs nothing:

```tsx
// src/app/[slug]/dashboard/page.tsx
const { session, schoolId } = await requireSchoolForSlugOrRedirect(slug);
const linkedTeacher = await getLinkedTeacherForUser(session.user.id);
const canReportSelf = canShowMyTeaching(linkedTeacher, schoolId);
...
<ReliefOverview ... canReportSelf={canReportSelf} />
```

```tsx
// ReliefOverview.tsx
<Link href={canReportSelf ? `/${slug}/dashboard/my-relief?tab=report` : `/${slug}/reporting`}>
```

(Alternative, if you'd rather not thread a prop: let `my-relief?tab=report` render the report tab for a KP with no Teacher row — filing on behalf of a colleague needs no `self` — and only gate the `duties` tab. That keeps one destination but widens the page's contract.)

---

src/app/[slug]/reporting/page.tsx
```diff
@@ -157,40 +178,62 @@ export default async function ReportPage({
...
+    // Unreachable in practice: resolveReportingAccess only returns the
+    // "admin" kind when `session.user` exists. Guarding explicitly (instead
+    // of a non-null assertion) keeps this branch honest for the type
+    // checker without a cast. `notFound()` rather than `return null`: a bare
+    // `null` render is a blank white page served as HTTP 200, the worst
+    // failure mode if a future edit ever makes this reachable.
+    if (!session) notFound();
+    const flags = await getFeatureFlags();
+    return (
+      <DashboardShell role={session.user.role} flags={flags}>
```
Comment (guangshinhaha) [inline src/app/[slug]/reporting/page.tsx]: **[Medium] `showMyRelief` is omitted here, so a teaching KP's sidebar loses "My Relief" on this one page**

```tsx
    if (!session) notFound();
    const flags = await getFeatureFlags();
    return (
      <DashboardShell role={session.user.role} flags={flags}>   // ← no showMyRelief
```

**Problem:** `DashboardShell.tsx:31` declares `showMyRelief = false` as the default, and passes it straight to `DashboardSidebar` and `BottomTabBar`. Every page under `/[slug]/dashboard/*` gets the real value from `dashboard/layout.tsx:50`; this branch is the one place that doesn't.

Concretely: a teaching KP who deep-links or bookmarks `/[slug]/reporting` gets dashboard chrome whose sidebar and bottom tab bar are silently missing the "My Relief" item that is present on every other dashboard page for that same user. Given the whole point of the branch is nav continuity for KPs, a nav that changes shape depending on which of two absence-reporting routes you're on is the specific thing being fixed.

**Suggestion:** `self` is already computed at `:125` and is exactly "linked Teacher row at this school" — no extra query needed:

```tsx
      <DashboardShell
        role={session.user.role}
        flags={flags}
        showMyRelief={self !== null}
      >
```

---

src/app/[slug]/dashboard/my-relief/page.tsx
```diff
@@ -0,0 +1,168 @@
...
+  const tab = resolveTab(rawTab);
+
+  let reportTab: React.ReactNode = null;
+  if (tab === "report") {
+    // Copied from src/app/[slug]/reporting/page.tsx — same two queries and
+    // the same "is this KP's own teacher id actually in the dropdown"
+    // membership guard, so the pre-select never points at a hidden row (a
+    // KP's own teacher row can carry PERMANENT_RELIEF, which is filtered
+    // out of the dropdown just below).
+    const teachersForDropdown = await prisma.teacher.findMany({
```
Comment (guangshinhaha) [inline src/app/[slug]/dashboard/my-relief/page.tsx]: **[Medium] The absence-form data queries (and the `PERMANENT_RELIEF` business rule) are now copy-pasted into a second page**

```ts
    // Copied from src/app/[slug]/reporting/page.tsx — same two queries and   // ←
    // the same "is this KP's own teacher id actually in the dropdown" ...
    const teachersForDropdown = await prisma.teacher.findMany({
      where: {
        schoolId,
        OR: [
          { schoolRoleId: null },
          { schoolRole: { category: { not: "PERMANENT_RELIEF" } } },
        ],
      },
      ...
```

**Problem:** `reporting/page.tsx:133-167` and this block are now byte-for-byte the same roster query, the same `periods` query, and the same membership guard. The comment says so out loud. CLAUDE.md → Code Organisation: *"Shared lookup data (categories, enums, static lists) must be declared once in `src/lib/` and imported — never redeclared as local variables per page."*

The `PERMANENT_RELIEF` exclusion is not formatting — it is the business rule "relief teachers can't submit sick reports because relief teachers don't get covered". The failure mode is concrete: when a school adds a second non-reporting `SchoolRoleCategory`, or the rule changes to include `overrideId`-scoped periods, whichever copy the author doesn't grep for keeps serving the old roster, and the two absence forms disagree about who can be reported absent — on the same school, in the same session.

**Suggestion:** one exported loader, two callers.

```ts
// src/lib/reporting/formData.ts
export async function getAbsenceFormData(schoolId: string) {
  const [teachersForDropdown, periods] = await Promise.all([
    prisma.teacher.findMany({
      where: {
        schoolId,
        OR: [
          { schoolRoleId: null },
          { schoolRole: { category: { not: "PERMANENT_RELIEF" } } },
        ],
      },
      orderBy: { name: "asc" },
      select: {
        id: true,
        name: true,
        subjectLinks: { select: { subject: { select: { name: true } } } },
      },
    }),
    prisma.period.findMany({
      where: { schoolId, overrideId: null },
      orderBy: { number: "asc" },
      select: { id: true, number: true, startTime: true, endTime: true },
    }),
  ]);
  return { teachersForDropdown, periods };
}

/** A pre-select is only honest if the row is actually in the rendered list. */
export function resolveDefaultTeacherId(
  teachers: { id: string }[],
  candidateId: string | null,
): string | null {
  return candidateId && teachers.some((t) => t.id === candidateId) ? candidateId : null;
}
```

Both call sites then shrink to two lines each, and the `Promise.all` also fixes the sequential-await pair both pages currently pay for.

---

src/app/[slug]/dashboard/layout.tsx
```diff
@@ -37,28 +28,30 @@ export default async function DashboardLayout({
...
+  // The layout only has the session, not the resolved school for the URL
+  // slug — that resolution happens per-page via
+  // `requireSchoolForSlugOrRedirect` (needed for the #417 SUPERADMIN case,
+  // where the session itself carries no usable school). So this compares
+  // the linked Teacher row's schoolId against the session's OWN schoolId,
+  // not the slug's school: "does this user have a Teacher row whose
+  // schoolId matches their session's schoolId?" That's the same comparison
+  // dashboard/page.tsx used to make before this rework, just relocated.
+  const linkedTeacher = await getLinkedTeacherForUser(session.user.id);
+  const showMyRelief = canShowMyTeaching(linkedTeacher, session.user.schoolId ?? "");
```
Comment (guangshinhaha) [inline src/app/[slug]/dashboard/layout.tsx]: **[Low] The nav gate and the page gate use two different definitions of "this school"**

```tsx
  const linkedTeacher = await getLinkedTeacherForUser(session.user.id);
  const showMyRelief = canShowMyTeaching(linkedTeacher, session.user.schoolId ?? "");   // ←
```

vs. `my-relief/page.tsx:58-61`, which compares against the **slug's** resolved school:

```ts
const { session, schoolId } = await requireSchoolForSlugOrRedirect(slug);
const linkedTeacher = await getLinkedTeacherForUser(session.user.id);
if (!linkedTeacher || !canShowMyTeaching(linkedTeacher, schoolId)) redirect(...);
```

**Problem:** For a normal KP these are forced equal by `[slug]/layout.tsx`'s `TENANT_LOCKED_SUBTREE` check, so the comment's reasoning holds. The exception is exactly the case that check exempts: a `SUPERADMIN` (#417) who has a non-null `session.user.schoolId` **and** a Teacher row at that school, viewing a *different* school's dashboard. `session.user.schoolId === linkedTeacher.schoolId` → the sidebar renders "My Relief"; clicking it hits `my-relief/page.tsx`, where `schoolId` is the *other* school's, fails the gate, and redirects back. A visible nav item that can only bounce you.

Low, not higher, because it leaks nothing — the page-level gate is the correct one and it holds. But two spellings of one predicate is how the correct one eventually gets "simplified" to match the wrong one.

**Suggestion:** layouts receive `params`, so the layout can use the same notion of "this school" the page does:

```tsx
export default async function DashboardLayout({
  children,
  params,
}: {
  children: React.ReactNode;
  params: Promise<{ slug: string }>;
}) {
  const { slug } = await params;
  ...
  const linkedTeacher = await getLinkedTeacherForUser(session.user.id);
  const showMyRelief =
    session.user.school?.slug === slug &&
    canShowMyTeaching(linkedTeacher, session.user.schoolId ?? "");
```

---

src/app/[slug]/teacher/AssignmentCard.tsx
```diff
@@ -0,0 +1,229 @@
...
+  acknowledgedAt: string | null;
+  declinedAt: string | null;
+  declineReason: string | null;
+};
+
+/**
+ * Shape returned by GET /api/teacher-dashboard.
+ *
+ * Shared by MyReliefView (teacher's own full view) and the KP dashboard's
+ * MyTeachingPanel — both call the same endpoint, so this lives next to
```
Comment (guangshinhaha) [inline src/app/[slug]/teacher/AssignmentCard.tsx]: **[Low] Both doc comments in this new file point at `MyTeachingPanel`, which this PR deletes**

```ts
/**
 * Shape returned by GET /api/teacher-dashboard.
 *
 * Shared by MyReliefView (teacher's own full view) and the KP dashboard's
 * MyTeachingPanel — both call the same endpoint, so this lives next to   // ←
 * `Assignment` instead of being redeclared per consumer.
 */
```

**Problem:** `MyTeachingPanel.tsx` is gone (correctly — the PR description says it was rejected on review and deleted, and it is genuinely absent from `src/app/[slug]/dashboard/`). But this file's `Assignment` doc at `:11` and `DashboardData` doc at `:32` both still name it as the second consumer. The actual second consumer is `dashboard/my-relief/MyDutiesView.tsx`. Since the stated justification for extracting this file at all is "the other consumer needs it", a reader grepping for the named consumer finds nothing and can reasonably conclude the extraction is now dead weight.

**Suggestion:**

```ts
 * Shared by MyReliefView (the teacher portal's full view) and MyDutiesView
 * (the KP dashboard's My Relief → My Duties tab) — both call the same
 * endpoint, so this lives next to `Assignment` instead of being redeclared
 * per consumer.
```

Same substitution at `:11`.

---

src/app/[slug]/dashboard/DashboardSidebar.tsx
```diff
@@ -74,6 +81,9 @@ export default function DashboardSidebar({ role }: { role: UserRole }) {
   const isOnReliefTeachers = pathname.startsWith(basePath + "/relief-teachers");
   const rawReliefTab = searchParams.get("tab");
   const activeReliefTab = rawReliefTab === "my-list" || rawReliefTab === "rems" ? rawReliefTab : "my-list";
+  const isOnMyRelief = pathname.startsWith(basePath + "/my-relief");
+  const rawMyReliefTab = searchParams.get("tab");
```
Comment (guangshinhaha) [inline src/app/[slug]/dashboard/DashboardSidebar.tsx]: **[Low] `rawMyReliefTab` re-reads a query param already in scope, and this is the third verbatim copy of the expand/hydrate/persist triplet**

```tsx
  const rawReliefTab = searchParams.get("tab");
  const activeReliefTab = rawReliefTab === "my-list" || rawReliefTab === "rems" ? rawReliefTab : "my-list";
  const isOnMyRelief = pathname.startsWith(basePath + "/my-relief");
  const rawMyReliefTab = searchParams.get("tab");   // ← identical to rawReliefTab
  const activeMyReliefTab = rawMyReliefTab === "report" ? "report" : "duties";
```

**Problem:** Two things, both small:

1. `rawMyReliefTab` and `rawReliefTab` are the same call on the same object. There is one `?tab=` in the URL; naming it twice implies there are two, and a future reader has to diff the two lines to confirm they aren't.
2. Lines `129-144` are now the third near-identical copy of the same six-line pattern (`useState` pair → read localStorage on mount → force-expand when on the section → persist when hydrated), after `reliefExpanded` at `:112-127` and alongside `settingsExpanded`. Each copy carries the same subtle hydration contract the block comment at `:88-98` explains once; the third copy is where that contract starts getting mis-transcribed.

Not blocking — the copy is faithful and works.

**Suggestion:**

```tsx
const rawTab = searchParams.get("tab");
const activeReliefTab = rawTab === "my-list" || rawTab === "rems" ? rawTab : "my-list";
const activeMyReliefTab = rawTab === "report" ? "report" : "duties";
```

and, if you want the duplication gone too:

```tsx
function usePersistedExpansion(storageKey: string, isActive: boolean) {
  const [expanded, setExpanded] = useState(false);
  const [hydrated, setHydrated] = useState(false);
  useEffect(() => {
    setExpanded(localStorage.getItem(storageKey) === "true");
    setHydrated(true);
  }, [storageKey]);
  useEffect(() => {
    if (isActive) setExpanded(true);
  }, [isActive]);
  useEffect(() => {
    if (hydrated) localStorage.setItem(storageKey, String(expanded));
  }, [expanded, hydrated, storageKey]);
  return [expanded, setExpanded] as const;
}
```

---

src/app/[slug]/teacher/TeacherShell.tsx
```diff
@@ -22,6 +23,7 @@ export default function TeacherShell({ children }: { children: React.ReactNode }
       </div>
       <main className="flex flex-1 flex-col overflow-hidden">
         <TeacherMobileHeader />
+        <DesktopTopBar />
```
Comment (guangshinhaha) [inline src/app/[slug]/teacher/TeacherShell.tsx:26]: **[Medium] The demo teacher shell is a hand-copy of this file and wasn't updated — `/demo/teacher` now has no sign-in control and no bell on desktop**

```tsx
      <main className="flex flex-1 flex-col overflow-hidden">
        <TeacherMobileHeader />
        <DesktopTopBar />   // ← added here, but not in the demo copy
```

**Problem:** `src/app/demo/teacher/layout.tsx` duplicates this shell inline (it imports `TeacherSidebar`, `TeacherMobileHeader`, `TeacherBottomTabBar`, `LockBodyScroll` directly rather than using `TeacherShell`), and this PR does not touch it.

This PR removed two things from `TeacherSidebar`: the `NotificationBell` in the header block, and the entire footer `<div>` containing the `slug === "demo" ? "Sign in" : "Log out"` control. Both were relocated into `DesktopTopBar`, which the demo layout does not render. `TeacherMobileHeader` — the only other host for either control — is `md:hidden`.

Net effect on desktop `/demo/teacher`: **no notification bell and no "Sign in" link at all.** On the public demo tree, "Sign in" is the conversion CTA; it is now reachable only by shrinking the viewport below `md`.

That this is an oversight rather than a decision is visible in the diff: the sibling `src/app/demo/dashboard/layout.tsx` — the same kind of hand-copied shell — *was* updated in this PR (new `DesktopTopBar` import path plus `showMyRelief={false}` on both nav components).

**Suggestion:**

```tsx
// src/app/demo/teacher/layout.tsx
+import DesktopTopBar from "@/app/[slug]/_components/DesktopTopBar";
 ...
       <main className="flex flex-1 flex-col overflow-hidden">
         <TeacherMobileHeader />
+        <DesktopTopBar />
```

Worth a follow-up: both demo layouts exist only because they hard-code `role`/`showMyRelief` and skip auth. Now that `DashboardShell` exists, `demo/dashboard/layout.tsx` is a one-line wrapper away from being deleted, and the same could be done for the teacher side — which is what would have prevented this drift.

---

guangshinhaha:
## Code Review Summary

| Severity | Count |
|----------|-------|
| Critical | 0 |
| High     | 1 |
| Medium   | 3 |
| Low      | 3 |

The one that blocks: the dashboard's "Report Absence" CTA now points unconditionally at a destination that redirects pure coordinators (and superadmins) away, removing their last UI path to the absence form. Everything else is nav/duplication drift.

Both named security invariants were checked and hold:
- **#330** — `wrong-school` and `signed-out` both `return` at `reporting/page.tsx:49` / `:75`, *before* the `teachersForDropdown` and `periods` queries at `:133`. No roster, names, subjects or period structure is reachable without a same-school session. Unchanged by this PR.
- **#375/#382** — `DashboardShell` renders on `/[slug]/reporting` only under `access.kind === "admin"`, which `resolveReportingAccess:63-64` defines as `user.schoolId === schoolId && can(role, VIEW_DASHBOARD)`. That is *strictly narrower* than `dashboard/layout.tsx`'s own gate (which additionally admits cross-school SUPERADMIN), so extracting the chrome reaches nobody the layout wouldn't. The shell is presentation-only; auth stayed in the layout.
- The new cross-tenant gate is correct where it counts: `my-relief/page.tsx:59` compares the linked Teacher row against the **slug's** resolved school from `requireSchoolForSlugOrRedirect`, so a stale cross-school link cannot satisfy it. (The nav-visibility copy in the layout uses a looser comparison — filed as a Low.)

## What Looks Good

- **`lockedTeacherId` vs `defaultTeacherId` is genuinely airtight.** The read-only branch keys on `lockedTeacherId` alone (`:1063`, `:244`), never on `teacherId`, so a pre-selected KP provably lands in the changeable Combobox. The `??` chain is applied consistently at all four sites — both initial `useState`s and both "Submit Another" resets — and the reset at `:925-932` correctly still drops `periodsByDate` manually, since the teacher-change effect now won't fire for a *defaulted* id either, not just a locked one. That was the easy thing to miss.
- **The membership guard on the pre-select** (`reporting/page.tsx:164-167`). Noticing that a KP's own row can be `PERMANENT_RELIEF` and therefore filtered out of the very dropdown it would be selected in — producing a blank field that silently satisfies `required` — is the kind of thing that normally ships and gets found in prod.
- **`MyTeachingPanel` really is deleted**, not orphaned: absent from `src/app/[slug]/dashboard/`, zero code references, only doc/comment mentions remain. The `DesktopTopBar` move is complete too — no importer still points at the old `dashboard/` path.
- **`canShowMyTeaching` extracted as a named predicate rather than an inline `?.schoolId ===`**, with the doc comment explaining exactly which future "simplification" it exists to survive, plus the three unit cases that pin it. Same for lifting `dateNav` out of `MyReliefView` instead of copying three date helpers into the new view.
- **First-ever coverage for `GET /api/teacher-dashboard`**, including the adversarial same-named-teacher-in-school-B isolation case — written before the feature that depends on it, to falsify the premise rather than confirm it.

---

PR #720, opened 2026-08-27
feat(absences): let partial-day absences use a time range, not clicks

### LINKED ISSUE #712: fix(absences): let a partial-day absence mark free periods between lessons as unavailable
## User story

As a teacher reporting a partial-day absence, I want to also mark a period where I have no lesson but am still away from school (e.g. a free period sandwiched between two absence-covered lessons) as unavailable, so that I'm not incorrectly shown as available to cover relief during a period I'm not actually at school for.

## Background

Reported scenario: a teacher is away from P1 through P3, but only teaches at P1 and P3 (P2 is a free period on their normal timetable). Verified precisely against the code:

- **The partial-day period picker only offers lesson periods.** `AbsenceReportForm.tsx`'s "Which periods need relief?" picker (~line 1435-1470) iterates `teacherPeriods` - the teacher's own timetabled lessons for that date - not the school's full period list. Whenever a teacher has at least one lesson that day (the normal case), a free period like P2 is never offered as a selectable option at all. There's no way to indicate "I'm also away at P2" through this picker, not just a downstream logic gap.
- **Even if it were selectable, the server would record it wrong.** `writeAbsencePeriodRequirements`'s `PARTIAL_DAY` branch (`src/lib/absencePeriodRequirements.ts`, ~line 121-147) unconditionally writes every submitted periodId with `coverageType: RELIEF_REQUIRED` - meaning "a class needs covering here." A free period has no class, so writing it as `RELIEF_REQUIRED` would be actively wrong, not just unhelpful.
- **The data model already anticipated exactly this need.** `PeriodCoverageType.UNAVAILABLE` is documented directly in `schema.prisma`: *"reserved for teachers marking a free period as unavailable for relief."* It's already correctly written today, but only by a separate, parallel flow: the "Time Off" case (`isTimeOff` branch in the same function, triggered by `reason === "TIME_OFF"` combined with `NO_RELIEF` coverage). That flow lets a teacher select *any* period, not just lesson periods, and correctly writes `UNAVAILABLE` for each one selected. The read side already handles it correctly too: `unavailablePeriodsForAbsence` (`src/lib/scheduling/absenceCoverage.ts`) treats a row with *any* of `RELIEF_REQUIRED`, `NO_RELIEF`, or `UNAVAILABLE` as "the teacher isn't at school during that period," per its own doc comment - confirmed no changes are needed on that side.

So the actual gap is narrow and specific: **Time Off** (nothing needs covering at all) already supports marking free periods unavailable. **Partial-day** (some lesson periods genuinely need relief) has no equivalent - there's no way to combine "P1 and P3 need relief coverage" with "P2 is also a period I won't be at school for" in the same report. This issue bridges that gap for the mixed case, reusing the exact mechanism (`UNAVAILABLE`) the Time Off flow already proves works correctly end to end.

### Teacher can mark a free period as unavailable within a partial-day absence

- **Given** a teacher is filing a partial-day absence and has a free period (no lesson) that day
- **When** they view the period picker
- **Then** they can select that free period, distinct from the lesson periods that need relief, and submit the absence with it included

### A marked free period is recorded as unavailable, not as needing relief

- **Given** a teacher has selected a free period alongside lesson periods in a partial-day absence
- **When** the absence is submitted
- **Then** the free period is recorded with unavailable coverage, while the selected lesson periods are still recorded as needing relief

### The teacher is excluded from relief consideration during that free period

- **Given** a teacher has marked a free period as unavailable within a partial-day absence
- **When** a different teacher's absence needs relief coverage for that same period
- **Then** the first teacher is excluded as a relief candidate for it, the same way they're already excluded during their absence-covered lesson periods

### Not marking a free period leaves today's behavior unchanged

- **Given** a teacher files a partial-day absence and does not select any free period
- **When** the absence is submitted
- **Then** only the explicitly selected lesson periods are recorded, and the teacher remains available for relief during any unselected free period, exactly as today

## Out of scope

- The FULL_DAY absence flow - already covers the whole day via existing timetable/no-relief-flag resolution, unaffected by this gap.
- The existing "Time Off" flow - already correct; used here only as the reference pattern for reusing `UNAVAILABLE` coverage, not modified by this issue.
- Automatically inferring which free periods "should" be marked unavailable from a start/end period range (e.g. "P1 to P3" as a single range input) - this issue adds the ability to explicitly select a free period one at a time, matching the existing picker's interaction model; range-based inference could be a future UX improvement but isn't required to close this gap.
- Any change to `unavailablePeriodsForAbsence`, `getAbsentTeacherPeriods`, or any other read-side consumer - already verified correct for `UNAVAILABLE` rows, no changes needed.

## Design assets

N/A (no Figma). The picker gains a second, visually distinct section beneath the existing one:

```mermaid
flowchart TD
    A[Teacher files a partial-day absence] --> B["Which periods need relief?" - lesson periods only, unchanged]
    A --> C["Also unavailable? - free periods with no lesson, new"]
    B --> D[Selected lesson periods -> RELIEF_REQUIRED]
    C --> E[Selected free periods -> UNAVAILABLE]
    D --> F[Teacher excluded from relief consideration for all selected periods]
    E --> F
```

---

## Technical context

**Client (`src/app/[slug]/reporting/AbsenceReportForm.tsx`, ~line 1424-1493):** add a second selectable section under the existing "Which periods need relief?" picker, listing the teacher's free periods for `currDate` (the school's `periods` list, filtered to exclude anything already in `teacherPeriods`), labeled to make the distinction clear (e.g. "Also unavailable - no lesson, but you won't be at school"). Needs its own selection state, kept separate from the existing `selectedPeriods` (which represents "needs relief") - e.g. a parallel `selectedUnavailablePeriods: Record<string, string[]>`, reusing the existing `selectedPeriodsInDateWindow` clamping helper the same way `selectedPeriods` already does for date-window changes.

**Submission payload (~line 799-829):** the request body needs to carry both sets distinguishably - e.g. a new `unavailablePeriods: Record<string, string[]>` field alongside the existing `selectedPeriods`, not merged into one undifferentiated list, since the server needs to know which periodIds get which coverage type.

**Server (`src/app/api/absence-reports/route.ts` and `writeAbsencePeriodRequirements` in `src/lib/absencePeriodRequirements.ts`):** thread the new field through `writeAbsencePeriodInput` as e.g. `unavailablePeriods?: Record<string, string[]> | null`. In the `PARTIAL_DAY` branch (~line 121-147): keep writing `RELIEF_REQUIRED` for `datePeriods` entries exactly as today, and add a parallel loop writing `UNAVAILABLE` for `unavailablePeriods` entries, validating each against `validPeriodIds` the same way the existing loop does. Reject (400, via `AbsencePeriodValidationError`) if the same periodId appears in both `datePeriods` and `unavailablePeriods` for the same date - an ambiguous request, not something to silently resolve one way or the other.

## Data model

No schema changes. `PeriodCoverageType.UNAVAILABLE` already exists and is already used correctly by the Time Off flow.

## API contract

`POST /api/absence-reports` gains an additive `unavailablePeriods: Record<string, string[]>` field (same shape as the existing `selectedPeriods`), read only when `coverageType: "PARTIAL_DAY"`. Omitted or empty is fully backward compatible - no change from today's behavior.

## Error contract

Reuses `AbsencePeriodValidationError`'s existing 400 mapping (invalid/cross-tenant periodId) for the new field. Adds one new validation case, also a 400: the same periodId present in both `datePeriods` and `unavailablePeriods` for the same date.

## Additional test scenarios

- A periodId submitted in both `datePeriods` and `unavailablePeriods` for the same date is rejected with 400, not silently resolved to one coverage type or the other.
- A free period marked unavailable on one date within a multi-day partial-day absence, but not selected on a different date in the same window, is recorded correctly per date - mirrors the existing per-date handling `datePeriods` already has.
- The mutual-swap-locked period behavior (`isSwapLocked` in the client, forcing a swap-covered period into the relief-needed set) is unaffected by the new free-period section - unrelated concerns.
- Submitting `unavailablePeriods` with a `coverageType` other than `PARTIAL_DAY` (e.g. `FULL_DAY`) is ignored rather than erroring, consistent with how the existing PARTIAL_DAY-only fields already behave when coverage type doesn't match.

## Hard constraints

- Reuse `PeriodCoverageType.UNAVAILABLE` - do not introduce a new enum value; the schema already anticipates this exact use and the read side already handles it.
- Do not modify the `isTimeOff` branch or the `FULL_DAY` branch in `writeAbsencePeriodRequirements` - both are correct and unrelated to this gap.
- A periodId must not be accepted as both "needs relief" and "unavailable" for the same date - catch this with a clear validation error before it reaches the database, not left to surface as an opaque unique-constraint violation on `AbsencePeriodRequirement`'s existing `@@unique([absenceReportId, periodId, applicableDate])`.

---

superb-sushi:
## Summary

- Replaces the "Which periods need relief?" per-period button grid with a start/end **time range** input on the partial-day absence flow, mirroring the existing Time Off UI.
- The range auto-resolves into lesson periods (needing relief) and free periods (marked unavailable) - closing the reported bug where a free period sandwiched between absence-covered lessons was invisible to relief-candidate exclusion.

 Fixes #712 

## Why

Original report: a teacher away P1-P3 who only teaches P1 and P3 had no way to also flag P2 (their free period) as unavailable - the old picker only ever offered lesson periods, and even if it had offered P2, the server would have written it as `RELIEF_REQUIRED` (wrong - there's no class there).

Initial plan was a narrower fix (a second "also unavailable" picker, click each free period manually). During implementation we discussed the UX further and settled on a materially better design: instead of asking the teacher to notice their own free period exists and remember to tick it, they enter a time range - exactly like they already do on the Time Off form - and the system derives both sets automatically. This removes the exact failure mode that caused the original bug (a step easy to forget) rather than just adding a place to fix it manually.

## Changes

- `src/lib/partialDayRange.ts` (new): pure `resolvePartialDayRange({startTime, endTime, periods, teacherPeriods})` - the same overlap logic `TimeOffRequestForm` already uses (`p.startTime < endTime && p.endTime > startTime`), splitting the result into `reliefPeriodIds` (the teacher's own lessons in range) and `unavailablePeriodIds` (everything else in range). Deliberately decoupled from the form's own types so it's cheap to unit test.
- `src/lib/absencePeriodRequirements.ts`: `writeAbsencePeriodInput` gains `unavailablePeriods?: Record<string, string[]> | null`. The PARTIAL_DAY branch now writes a parallel `UNAVAILABLE` row for each, and rejects (400, via the existing `AbsencePeriodValidationError`) any periodId present in both `datePeriods` and `unavailablePeriods` for the same date - an ambiguous request caught before it reaches the DB rather than surfacing as an opaque unique-constraint violation.
- `src/app/api/absence-reports/route.ts`: new `multiDayUnavailablePeriods` field threaded through to the write helper. PARTIAL_DAY validation relaxed from "must have `multiDayPeriods`" to "must have `multiDayPeriods` OR `multiDayUnavailablePeriods`" - a range covering only a free period (nothing to cover, but something to mark unavailable) is now legitimately empty on the relief side.
- `src/app/[slug]/reporting/AbsenceReportForm.tsx`:
  - New per-date state: `partialStartTimes`/`partialEndTimes` (the time inputs), `selectedUnavailablePeriods` (mirrors the existing `selectedPeriods` shape).
  - A new effect recomputes both `selectedPeriods[currDate]` and `selectedUnavailablePeriods[currDate]` from `resolvePartialDayRange` whenever the range, that date's timetable, or the mutual-swap-locked period changes. A swap-locked period is unioned in separately so it stays selected even if it falls outside the chosen range - it's guaranteed covered regardless.
  - Removed the now-dead `togglePeriod` function and the button-grid UI; replaced with the time-range inputs plus a small transparency summary ("Needs relief: P1 Math, P3 English" / "Also unavailable: P2 (no lesson there, so nothing to cover)").
  - Submit-disable and the "Relief Instructions" notes section both already read from `selectedPeriods`, so they keep working unchanged once it's populated by the effect instead of clicks - only the disable condition needed updating, to also allow submission when `selectedUnavailablePeriodCount > 0` even if no relief periods were resolved.
  - State reset (on teacher change, on successful submit) extended to cover the three new state slots.

## Known tradeoff (flagging, not silently dropped)

The old picker let a teacher deselect one lesson within their absence for a self-study exception (e.g. "away P1-P3, but P2's class can work independently"). That manual override is gone - relief periods are now fully derived from the time range, with no per-period opt-out. This wasn't requested as part of the redesign, so it wasn't rebuilt, but worth knowing before merging.

## Test plan

- [x] `npx tsc --noEmit` - clean.
- [x] `npm run build` - clean, `✓ Generating static pages (108/108)`, no `⨯`.
- [x] `npx vitest run` - 1132 passed, 2 pre-existing skips, 102 files (14 new tests: `partialDayRange.test.ts` covers the pure resolver's split/overlap/edge cases; `absencePeriodRequirements.test.ts` covers the new `UNAVAILABLE` write path, the same-date conflict rejection, cross-tenant rejection, and the "unavailable periods alone" case).
- [x] `npm audit` (via pre-push hook) - 0 vulnerabilities.
- [ ] `npm run test:api` - **not run in this session.** No integration test changes were needed for `tests/api/absence-reports.test.ts` itself (reviewed its existing PARTIAL_DAY/overlap tests - none reference the new field, all still supply `multiDayPeriods` the same way as before, so they're unaffected), but this hasn't been confirmed by actually running the suite against Postgres. **Please run `npm run test:api` locally before merging.**
- [ ] Manual browser check recommended: the swap-locked-period-stays-included-outside-the-range interaction, and the "range covers only a free period" case, are both easier to eyeball in the actual form than to fully trust from code review alone.

---

*🤖 Generated with Claude Code*


superb-sushi:
## Update: self-study override restored

Follow-up commit on this branch, addressing feedback on the first version.

**What changed:** the initial time-range redesign auto-resolved every lesson period in the chosen range as needing relief, with no way to exclude one. That was a real regression from the old picker, which let a teacher untick a period their class could self-study through (e.g. "away P1-P3, but P2's class can work independently"). This commit restores that.

- `partialRangeResolution` (a `useMemo` wrapping `resolvePartialDayRange`) is now the single computation of what the range resolves to - recomputed only when the range, date, or that date's timetable actually changes, not on every render.
- The pre-fill effect reads from that memo instead of recomputing inline, so it only re-asserts the full relief/unavailable sets when something real changed.
- `togglePeriod` is back (previously removed), scoped to the lesson periods the current range resolved - rendered as the same clickable button grid the old picker used. A manual untick now survives between renders, since the effect above won't stomp it unless the range itself moves.
- The mutual-swap-covered period is still locked and un-toggleable, exactly as before.
- Free periods are **not** toggleable - there's no lesson to self-study through, so the "also unavailable" list stays fully automatic.
- Split the "nothing selected" validation message in two: a range that resolves to nothing at all ("No periods fall within this time range...") now reads differently from a range with periods where the teacher unticked every one ("You've unticked every period in this range...").

**Test plan:** `tsc --noEmit` and `npm run build` clean; unit suite unaffected (1132/1132 - this is a component-only UI change with no new test coverage, consistent with this repo having no component-render test harness, per the original PR's test plan note).

